PR #17, opened 2025-07-28
[Doc] Installation of ucm

flesher0813:
### What this PR does / why we need it?
Add installation docs

### Does this PR introduce any user-facing change?
no

### How was this patch tested?
no need test

### These installations shows like below
image
image

### DIFF
--- a/docs/source/getting-started/installation.md
+++ b/docs/source/getting-started/installation.md
@@ -4,7 +4,7 @@ This document describes how to install unified-cache-management.
 ## Requirements
 - OS: Linux
 - Python: >= 3.9, < 3.12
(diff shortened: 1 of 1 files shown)


docs/source/getting-started/installation.md
```diff
@@ -1 +1,56 @@
 # Installation
+This document describes how to install unified-cache-management.
+
+## Requirements
+- OS: Linux
+- Python: >= 3.9, < 3.12
+- GPU: compute capability 7.0 or higher (e.g., V100, T4, RTX20xx, A100, L4, H100, etc.)
```
ygwpz: if we want to start vllm v1 engine, compute capability should > 8.0

PR #79, opened 2025-08-11
[Feature] add ucm_sparse v1.0: unified sparse attention algorithm framework 

hek14:
# Prupose
This PR implements an unified sparse attention framework for vllm.

It injects several necessary hooks into the vllm codebase, which provides the possibilities of:

- layer-wise retrieval of "important blocks"
- manipulations of kv cache before and after attention

# Modifications 
- `unifiedcache/ucm_sparse/base.py`: provides coupled hooks like `request_begin`/`request_finished`, `execute_begin`/`execute_finished`, `attention_begin`/`attention_finished`
- `unifiedcache/ucm_sparse/factory.py`: ucm_sparse instance initializer
- `unifiedcache/ucm_sparse/ucm_sparse_state.py`: state management with single agent pattern


unifiedcache/integration/vllm/ucm_sparse/base.py
propanone1006, line 1: It is recommended to use pre-commit for formatting before submission, avoiding not going through the workflow

PR #103, opened 2025-08-21
[Perf]Modify start_load_kv

qyh111:
# Prupose

# Modifications 

# Test

How was this patch tested?


unifiedcache/integration/vllm/uc_connector.py
```diff
@@ -127,6 +127,7 @@ def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
         # dump tasks record request -> block -> list[task]
         self.dump_tasks: dict[str, dict[str, List[Task]]] = {}
         self.load_tasks: dict[str, tuple[Task, Task]] = {}
+        self.layerwise_load_tasks: dict[str, dict[str, tuple[Task, Task]]] = {}
```
ygwpz, line 130: we should put the same lifecycle param togather, and gc in request_finish()

PR #109, opened 2025-08-22
[Fix] Fix duplicate create/commit errors upon preemption

flesher0813:
# Prupose

Fix duplicate create/commit errors when happen preempt
# Modifications 

Users will find it's possible to save block when happen preempting, but nothing would be changed on using ucm
# Test

How was this patch tested?

CI completed
image

Tested with dram connector
image
Tested with nfs connector
image
image

### DIFF
--- a/unifiedcache/integration/vllm/uc_connector.py
+++ b/unifiedcache/integration/vllm/uc_connector.py
@@ -556,6 +556,12 @@ class UnifiedCacheConnectorV1(KVConnectorBase_V1):
             the number of tokens that can be loaded from the
             external KV cache beyond what is already computed.
         """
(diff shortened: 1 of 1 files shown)


unifiedcache/integration/vllm/uc_connector.py
```diff
@@ -556,6 +556,9 @@ def get_num_new_matched_tokens(
             the number of tokens that can be loaded from the
             external KV cache beyond what is already computed.
         """
+        if hasattr(request, "succeed_dumped_blocks") and request.succeed_dumped_blocks:
```
ygwpz, line 562: add comment

PR #166, opened 2025-09-06
[Feat] Support load async

flesher0813:
# Purpose

Support load async when using disagg prefill feature

# Modifications 

Support load async
# Test

How was this patch tested?

Tested with offline test by setting kv_role == "kv_consumer"
image

### DIFF
--- a/ucm/integration/vllm/uc_connector.py
+++ b/ucm/integration/vllm/uc_connector.py
@@ -552,12 +552,12 @@ class UnifiedCacheConnectorV1(KVConnectorBase_V1):
         for req_id, tasks in self._need_load_reqs.items():
             unfinished_tasks = []
             for task in tasks:
-                ret, finish = self.connector.check(task)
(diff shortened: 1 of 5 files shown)


ucm/integration/vllm/uc_connector.py
```diff
@@ -535,6 +546,30 @@ def wait_for_tasks():
         self.dump_tasks.clear()
         return success_dumped_blocks if success_dumped_blocks else None
 
+    def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
+        """Get the finished recving and sending requests."""
+        done_recving: set[str] = set()
+        for req_id, tasks in self._need_load_reqs.items():
+            unfinished_tasks = []
+            for task in tasks:
+                ret, finish = self.connector.check(task)
```
ygwpz: return value not matched

PR #187, opened 2025-09-16
GSA Impl

HelenJia98:
# Purpose

# Modifications 

# Test

How was this patch tested?


setup.py
```diff
@@ -25,93 +25,126 @@
...
-        os.makedirs(os.path.dirname(dst_path), exist_ok=True)
-        shutil.copy(src_path, dst_path)
-        print(f"[INFO] Copied {src_path} → {dst_path}")
-        if isinstance(self.distribution.get_command_obj("develop"), develop):
-            shutil.copy(src_path, dev_path)
-            print(f"[INFO] Copied in editable mode {src_path} → {dev_path}")
+        so_files = []
+        search_patterns = [ext.name]
+        
+        if ext.name == "fsstore":
```
qyh111: it should be "nfsstore" not "fsstore"

PR #188, opened 2025-09-18
[perf]prepare offset in advance

qyh111:
# Purpose
Computing the data offset is a time-consuming operation. We should prepare it in advance instead of recalculating it each time.

# Modifications 
 prepare offset in advance

# Test
deepseek-v2-lite 20 tokens
image
image
image

QwQ-32B 100 tokens
image
image
image

How was this patch tested?


ucm/integration/vllm/uc_connector.py
```diff
@@ -109,6 +109,7 @@ def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
         self._need_load_reqs: dict[str, Union[list[int], list[Task]]] = {}
         self._load_failed_reqs: set[str] = set()
         self._load_req_to_blocks: dict[str, set[int]] = {}
+        self._initialize_dataoffset(vllm_config)
```
ygwpz: scheduler does not need calculate offset

qyh111: fixed

PR #215, opened 2025-09-26
[Docs]Update the performance results of prefix-cache

zhou-haitao:
# Purpose

Update the performance results of prefix-cache

# Modifications 

Update the performance results of prefix-cache

# Test

image


docs/source/user-guide/prefix-cache/nfs_store.md
ygwpz: now automaically caculate

PR #253, opened 2025-09-30
[Doc] Add README_zh.md of Chinese version

ChenyuZhu1:


README_zh.md
```diff
@@ -0,0 +1,60 @@
+<p align="center">
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="docs/source/logos/UCM.png">
```
mag1c-h: logo path error in dark mode, which should be "docs/source/logos/UCM-dark.png"

ChenyuZhu1: Modified in recent commit

PR #278, opened 2025-10-11
[Fix] Each request in the decode instance encounters a load failure

sumingZero:
Before the bug fix：
Snipaste_2025-10-10_19-46-34

After the bug fix：
image

### DIFF
--- a/ucm/integration/vllm/uc_connector.py
+++ b/ucm/integration/vllm/uc_connector.py
@@ -492,14 +492,23 @@ class UnifiedCacheConnectorV1(KVConnectorBase_V1):
                 continue
             unfinished_tasks = []
             for task in tasks:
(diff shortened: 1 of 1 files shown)


ucm/integration/vllm/uc_connector.py
```diff
@@ -492,12 +492,22 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
-                    continue
+                elif ret == 0 and finish == True:
+                    wret = self.connector.wait(task)
+                    if wret == 0:
+                        continue
+                    else:
+                        logger.error(
+                            f"Task {task} failed,  wait return {wret} for request {req_id}"
+                        )
+                else:
```
ygwpz: take care cyclomatic complexity.

PR #285, opened 2025-10-15
[WIP] Implementation of DRAM connector

ChenyuZhu1:
# Purpose

Implementation of DRAM connector

# Modifications 

A new infrastructure, the MemoryPool class is implemented and referenced in the dramstore implementation class, to store the KVCache data. The memory pool is initialized at the beginning with capacity and blockSize, and NewBlock(), LookupBlock(), GetAddress(), CommitBlock() interfaces are implemented here. It acts like an LRU-Cache. The storage pool is divided into slotNum=capacity/blockSize slots with dummy block names in the format `__slot_{index}`. During the initialization stage, all slots are pushed back into the LRU queue. Each time an allocation request (blockId) is requested, the last element of the LRU queue is popped, with the corresponding address assigned to the new blockId, which is soonly added into `addressMap_`.
Each time a block is committed, whether true or false, the block will be pushed into the LRU queue. If success==true, then it will be pushed to the front of the LRU queue (so that it is unlikely to be popped out soon); and if success==false, then it will be pushed to the tail of the LRU queue (so that it will be soonly popped out next time an allocation request comes). If success==true, the blockId will be added into the `availableBlocks_ `set.
Also, in the dramstore/domain directory, DramTsfTaskManager is implemented with a transfer task queue similar to the counterpart in the nfsstore.

# Test

under folder infra/mem_pool_test.cc and passed in CI

### DIFF
--- a/ucm/store/dramstore/cpy/dramstore.py.cc
+++ b/ucm/store/dramstore/cpy/dramstore.py.cc
@@ -100,7 +100,7 @@ PYBIND11_MODULE(ucmdramstore, module)
     module.attr("build_type") = UCM_BUILD_TYPE;
     auto store = py::class_<UC::DRAMStorePy>(module, "DRAMStore");
     auto config = py::class_<UC::DRAMStorePy::Config>(store, "Config");
-    config.def(py::init<const size_t, const size_t>(), py::arg("ioSize"), py::arg("capacity"));
+    config.def(py::init<const size_t, const size_t, const size_t, const size_t>(), py::arg("ioSize"), py::arg("capacity"), py::arg("streamNumber"), py::arg("timeoutMs"));
     config.def_readwrite("ioSize", &UC::DRAMStorePy::Config::ioSize);
     config.def_readwrite("capacity", &UC::DRAMStorePy::Config::capacity);
     config.def_readwrite("deviceId", &UC::DRAMStorePy::Config::deviceId);


ucm/store/dramstore/cc/api/dramstore.cc
```diff
@@ -24,27 +24,86 @@
...
 
 class DRAMStoreImpl : public DRAMStore {
 public:
-    int32_t Setup(const size_t ioSize, const size_t capacity, const int32_t deviceId) { return -1; }
-    int32_t Alloc(const std::string& block) override { return -1; }
-    bool Lookup(const std::string& block) override { return false; }
-    void Commit(const std::string& block, const bool success) override {}
+    int32_t Setup(const Config& config) {
+        // config里之前没有blockSize，稳妥起见还是先打桩
+        int32_t blockSize = 128;
```
harrisonyhq, line 37: can get blocksize here from vllm.config?

ChenyuZhu1: Yes you are right, but currently **blockSize** is not added into `config`. If possible, it will be added into the **Config** class soon.

qyh111, line 37: Do we need blocksize? config.def_readwrite("ioSize", &UC::DRAMStorePy::Config::ioSize) we have iosize here

ucm/store/dramstore/cc/domain/tsf_task/dram_tsf_task_queue.cc
```diff
@@ -0,0 +1,144 @@
...
+
+Status DramTsfTaskQueue::Setup(const int32_t deviceId, DramTsfTaskSet* failureSet, MemoryPool* memPool)
+{
+    this->_failureSet = failureSet;
+    this->_memPool = memPool;
+    if (deviceId >= 0) {
+        this->_device = DeviceFactory::Make(deviceId, 0, 0); // 这里不需要buffer，暂时都先传0吧
+        if (!this->_device) { return Status::OutOfMemory(); }
+    }
+    if (!this->_streamOper.Setup([this](DramTsfTask& task) { this->StreamOper(task); })) {
```
qyh111: I think this->_device->Setup() shoulld be called here

ucm/store/infra/memory/memory_pool.h
```diff
@@ -0,0 +1,105 @@
...
+    MemoryPool& operator=(const MemoryPool&) = delete;
+
+    void Reset() {
+        head_ = 0;
+    }
+
+    uint32_t GetNextAvailableOffset() const { return head_; }
+
+    Status NewBlock(std::string blockId) {
+        if (head_ >= capacity_) {
```
qyh111: Do we need to consider concurrency issues here?

ucm/store/dramstore/cc/api/dramstore.cc
```diff
@@ -24,27 +24,86 @@
...
+        this->memPool_ = std::make_unique<MemoryPool>(config.capacity, blockSize).release();
+        // 初始化memPool的办法是否正确？如果失败的话怎么办？
+        int32_t streamNumber = 60; // 这个参数是否需要，以及怎么传，还要讨论
+        int32_t timeoutMs = 10000; // 这个参数是否需要，以及怎么传，还要讨论
```
qyh111: We need to add these two variables to the config.

ucm/store/dramstore/cc/domain/tsf_task/dram_tsf_task_queue.cc
ChenyuZhu1: Yes

ucm/store/dramstore/cc/api/dramstore.cc
ChenyuZhu1: OK.

ucm/store/infra/memory/memory_pool.h
ChenyuZhu1: memory pool is rewritten today

PR #288, opened 2025-10-16
[Misc] add store intf with tensor addr ptr

mag1c-h:
# Purpose

The C++ base class in Store already takes a raw pointer, now add the same pointer-based API to the Python base class to keep it extension-friendly.

# Modifications 

Add new pointer-based API to the Store  Python base class.

# Test

CI passed with new added/existing test.

### DIFF
--- a/ucm/store/ucmstore.py
+++ b/ucm/store/ucmstore.py
@@ -113,7 +113,7 @@ class UcmKVStoreBase(ABC):
         self, block_ids: List[str], offset: List[int], src_tensor: List[torch.Tensor]
     ) -> Task:
         """
-        dump kv cache to device.
+        dump kv cache from device.
 
         Args:
             block_ids (List[str]): vLLM block hash.
@@ -154,7 +154,7 @@ class UcmKVStoreBase(ABC):
         size: List[int],
     ) -> Task:
         """
-        dump kv cache data to device.
+        dump kv cache data from device.
 
         Args:
             block_ids (List[str]): vLLM block hash.


pyxyzc:
@mag1c-h The semantics of the offset and size params kind of overlap here. Host–Storage transfers are already defined to operate on a page/block basis. Plz consider whether it really makes sense to expose this flexibility to the upper inference engines.

mag1c-h:
> @mag1c-h The semantics of the offset and size params kind of overlap here. Host–Storage transfers are already defined to operate on a page/block basis. Plz consider whether it really makes sense to expose this flexibility to the upper inference engines.

@pyxyzc  For the following reasons, we need to expose both offset and size parameters in the interface:
1. Block semantics differ across storage layers:
	- On Host/Device, a "block" refers to a layer-block.
	- On Storage, a "block" contains all layers, each mapped to a different offset.
2. GQA partitioning introduces per-tp offsets:
	When tp_range > 1 in GQA models, each tensor parallel rank starts from a distinct offset.
3. Future-proofing for variable-size metadata:
	We may store profiling/characterization data where a single batch could contain entries of differing sizes.

ucm/store/ucmstore.py
```diff
@@ -124,6 +124,48 @@ def dump(
...
+    @abstractmethod
+    def dump_data(
+        self,
+        block_ids: List[str],
+        offset: List[int],
+        src_addr: List[int],
+        size: List[int],
+    ) -> Task:
+        """
+        dump kv cache data to device.
```
ygwpz: maybe dump from device?

mag1c-h: fixed.

ygwpz (approved):

PR #292, opened 2025-10-20
[Feat] Adapt to vllm 0.11.0, remove finish dumping

harrisonyhq:
# Purpose

Adapt uc_connector to vllm 0.11.0, also adapt ut and workflow.

# Modifications 

1. Update the unittest.
2. Upgrade succeed_dumped_blocks to class variable.
3. In save_kv_layer, skip mla dump only when using ucmnfsstore.
4. Add Update_connector_output.

# Test

Offline/Online inference, workflow

### DIFF
--- a/ucm/integration/vllm/uc_connector.py
+++ b/ucm/integration/vllm/uc_connector.py
@@ -449,10 +449,17 @@ class UnifiedCacheConnectorV1(KVConnectorBase_V1):
 
         metadata = self._get_connector_metadata()
         assert isinstance(metadata, UCConnectorV1Metadata)
+        if self.is_mla and self.rank != 0 and isinstance(self.connector, UcmNfsStore):
+            for request in metadata.requests:
+                storage_block_ids = [block[0] for block in request.dump_blocks]
+                self.succeed_dumped_blocks.update(storage_block_ids)
+            return
+
         if self.use_layerwise:
             wait_for_tasks()
             # clear dump_tasks for all request
             self.dump_tasks.clear()
+            return
 
         for request in metadata.requests:
             if not request.dump_blocks:


ucm/integration/vllm/uc_connector.py
```diff
@@ -375,7 +380,7 @@ def save_kv_layer(
             attn_metadata (AttentionMetadata): the attention metadata.
             **kwargs: additional arguments for the save operation.
         """
-        if self.is_mla and self.rank != 0:
+        if self.is_mla and self.rank != 0 and isinstance(self.connector, UcmNfsStore):
```
flesher0813, line 383: What about the scenario tp > 1 with nfs connector and deepseek model? Seems the worker which rank > 0 wouldn't return succeed_dumped_blocks

harrisonyhq: fixed in [[Fix] Update succeed dumped blocks in mla scene when rank > 0](https://github.com/ModelEngine-Group/unified-cache-management/pull/292/commits/f826ac08b41613e2c3ebb63fcd9239ba4cc85dd4)

ucm/integration/vllm/uc_connector.py
```diff
@@ -434,23 +439,27 @@ def wait_for_save(self) -> Optional[dict[str, list[str]]]:
...
             for request_id, block_dump_tasks in self.dump_tasks.items():
                 for block_id, dump_tasks in block_dump_tasks.items():
                     if any(self.connector.wait(task) != 0 for task in dump_tasks):
                         continue
-                    success_dumped_blocks.setdefault(request_id, []).append(block_id)
+                    self.succeed_dumped_blocks.add(block_id)
 
         metadata = self._get_connector_metadata()
         assert isinstance(metadata, UCConnectorV1Metadata)
+        if self.is_mla and self.rank != 0 and isinstance(self.connector, UcmNfsStore):
```
flesher0813, line 452: This condition check needs to be implemented in the non-layerwise scenario as well.

harrisonyhq: The check to the mla mode and connector is ahead of layerwise, so this should work in both layerwise/non layerwise mode.

PR #299, opened 2025-10-21
[Feature]: pluggable SpaceLayout to avoid in-place rename for better performance on PCFS

mag1c-h:
# Purpose

In the default design, a Block file has two states:  
- **activated** – while data is being written  
- **archived** – after the write is finished  

The state is encoded in the file name and switched by an in-directory rename.  
On some storage systems (e.g. PCFS) this rename is expensive.  
This PR introduces an optional layout that keeps the two states in separate directories so the switch becomes a fast cross-directory move.

# Modifications 

- Turn `SpaceLayout` into an interface  
- Keep the existing behaviour in `SpaceShardLayout` (state encoded in file name, activated & archived live in the same directory)  
- Add `SpaceShardTempLayout`; when `tempDumpDirEnable=true` activated files are written to a temporary directory and moved to the final directory on completion

# Test

- CI passed with new added/existing test.
- E2E tests in NFSStore passed.

### DIFF
--- a/ucm/store/nfsstore/cc/domain/space/space_manager.cc
+++ b/ucm/store/nfsstore/cc/domain/space/space_manager.cc
@@ -95,20 +95,20 @@ Status SpaceManager::CommitBlock(const std::string& blockId, bool success) const
 {
     const auto activatedParent = this->layout_->DataFileParent(blockId, true);
     const auto activatedFile = this->layout_->DataFilePath(blockId, true);
+    const auto archivedParent = this->layout_->DataFileParent(blockId, false);
     auto status = Status::OK();
     do {
         if (!success) { break; }
-        const auto archivedParent = this->layout_->DataFileParent(blockId, false);
-        const auto archivedFile = this->layout_->DataFilePath(blockId, false);
         if (archivedParent != activatedParent) {
             status = File::MkDir(archivedParent);
             if (status == Status::DuplicateKey()) { status = Status::OK(); }
             if (status.Failure()) { break; }
         }
+        const auto archivedFile = this->layout_->DataFilePath(blockId, false);
         status = File::Rename(activatedFile, archivedFile);
     } while (0);
     File::Remove(activatedFile);
-    File::RmDir(activatedParent);
+    if (!success || archivedParent != activatedParent) { File::RmDir(activatedParent); }
     if (status.Failure()) {
         UC_ERROR("Failed({}) to {} block({}).", status, success ? "commit" : "cancel", blockId);
     }


ucm/store/nfsstore/cc/domain/space/space_manager.cc
```diff
@@ -77,29 +93,31 @@ Status SpaceManager::NewBlock(const std::string& blockId) const
...
+        const auto archivedParent = this->layout_->DataFileParent(blockId, false);
+        const auto archivedFile = this->layout_->DataFilePath(blockId, false);
+        if (archivedParent != activatedParent) {
+            status = File::MkDir(archivedParent);
+            if (status == Status::DuplicateKey()) { status = Status::OK(); }
+            if (status.Failure()) { break; }
+        }
+        status = File::Rename(activatedFile, archivedFile);
+    } while (0);
+    File::Remove(activatedFile);
```
ygwpz, line 110: if remove failed, what will happen?

mag1c-h: No obvious failure scenario exists; if any, leftover files will be cleaned up after expiration.

ucm/store/nfsstore/cc/domain/space/space_manager.cc
```diff
@@ -77,29 +93,31 @@ Status SpaceManager::NewBlock(const std::string& blockId) const
...
+    File::RmDir(activatedParent);
```
flesher0813: It will remove activatedParent when success? Feels like the successfully dumped block dirs would be removed when using space_shared_layout

ucm/store/nfsstore/cc/domain/space/space_shard_temp_layout.cc
```diff
@@ -0,0 +1,56 @@
...
+}
+
+std::vector<std::string> SpaceShardTempLayout::RelativeRoots() const
+{
+    auto roots = SpaceShardLayout::RelativeRoots();
+    roots.push_back(this->TempDataFileRoot());
+    return roots;
+}
+
+std::string SpaceShardTempLayout::TempDataFileRoot() const { return "temp"; }
```
qyh111, line 54: Would it be better to make ”temp“ configurable

ucm/store/nfsstore/cc/domain/space/space_manager.cc
```diff
@@ -77,29 +93,31 @@ Status SpaceManager::NewBlock(const std::string& blockId) const
...
+    do {
+        if (!success) { break; }
+        const auto archivedParent = this->layout_->DataFileParent(blockId, false);
+        const auto archivedFile = this->layout_->DataFilePath(blockId, false);
+        if (archivedParent != activatedParent) {
+            status = File::MkDir(archivedParent);
+            if (status == Status::DuplicateKey()) { status = Status::OK(); }
+            if (status.Failure()) { break; }
+        }
+        status = File::Rename(activatedFile, archivedFile);
```
zhou-haitao, line 108: Is the Rename method a copy operation? Will it cause performance issues?

ucm/store/nfsstore/cc/domain/space/space_manager.cc
mag1c-h: Yes, u r right. In this scenario, however, `activatedParent` will never be empty, so calling `RmDir` has no real effect.
Nevertheless, we shouldn't do this; I'll fix it right away.

ucm/store/nfsstore/cc/domain/space/space_manager.cc
mag1c-h: No actual data copy occurs; `rename` only updates metadata.

ucm/store/nfsstore/cc/domain/space/space_shard_temp_layout.cc
mag1c-h: No need—it could easily confuse callers.

ucm/store/nfsstore/cc/domain/space/space_manager.cc
mag1c-h: fix at [28245a0](https://github.com/ModelEngine-Group/unified-cache-management/pull/299/commits/28245a0fb20f2cf8f5e6f7afed7e8fd54725e23f)

ucm/store/test/case/nfsstore/space_manager_test.cc
ChenyuZhu1, line 1: In the newly-added unit test, do we need to examine the existence and the names of the temporary files?

ucm/store/nfsstore/cc/domain/space/space_manager.cc
```diff
@@ -77,29 +93,31 @@ Status SpaceManager::NewBlock(const std::string& blockId) const
...
+    auto status = Status::OK();
+    do {
+        if (!success) { break; }
+        if (archivedParent != activatedParent) {
+            status = File::MkDir(archivedParent);
+            if (status == Status::DuplicateKey()) { status = Status::OK(); }
+            if (status.Failure()) { break; }
+        }
+        const auto archivedFile = this->layout_->DataFilePath(blockId, false);
+        status = File::Rename(activatedFile, archivedFile);
```
UESTC-AHao, line 108: Could scenarios like duplicate commits or the file being unexpectedly deleted happen before the Rename?Is it necessary to do a check before Rename?

mag1c-h: No redundant checks are needed, for scenarios outside the design expectations, simply return the status code as usual.

ucm/store/test/case/nfsstore/space_manager_test.cc
mag1c-h: Sure, more detailed unit tests can be added, but they won't affect the current functionality.

PR #300, opened 2025-10-21
Rewrite dramstore using the reconstructed task architecture

ChenyuZhu1:
The memPool is based on PR https://github.com/ModelEngine-Group/unified-cache-management/pull/285, while the reconstructed dramstore is based on PR https://github.com/ModelEngine-Group/unified-cache-management/pull/296

Related unittests: ucm/store/test/case/infra/mem_pool_test.cc

Results of the performance test (dramstore_embed_and_fetch.py):
image

### DIFF
--- a/ucm/store/test/e2e/dramstore_embed_and_fetch.py
+++ b/ucm/store/test/e2e/dramstore_embed_and_fetch.py
@@ -131,7 +131,7 @@ def main():
     _, new_tensors = make_buffers(
         block_number, device_id, batch_size, block_dim, block_len, block_layer
     )
-    for batch in range(batch_number):
+    for batch in range(total_batches):
         start = batch_size * batch
         end = start + batch_size
         fetch(store, hashes[start:end], new_tensors)


ucm/store/infra/memory/memory_pool.h
```diff
@@ -0,0 +1,157 @@
...
+#include <unordered_map>
+#include <set>
+#include "status/status.h"
+
+namespace UC {
+
+class MemoryPool {
+public:
+    MemoryPool(size_t capacity, size_t blockSize)
+        : pool_(new char[capacity]),
```
qyh111, line 40: better use pinned memory here

ucm/store/dramstore/dramstore_connector.py
```diff
@@ -59,111 +61,50 @@ class UcmDramStore(UcmKVStoreBase):
...
-        """
-        get the underlying implementation of Store
-
-        Returns:
-            cc pointer to Store
-        """
-        return 0
+        capacity = int(config.get("capacity", 1073741824))  # Default 1GB
+        block_size = int(config.get("kv_block_size", 262144))  # Default 256KB
+        device_id = int(config.get("device_id", -1))
```
qyh111, line 68: kv_block_size is a concept within NfsStore, representing the total size of all layers of kv-cache contained in a single block. It seems that it is iosize not kv_block_size

ucm/store/dramstore/cc/domain/trans/dram_trans_queue.cc
```diff
@@ -0,0 +1,102 @@
...
+    }
+    auto status = Status::OK();
+    if (shard.type == Task::Type::DUMP) {
+        // TODO: D2H
+        status = this->D2H(shard, device);
+    } else {
+        // TODO: H2D
+        status = this->H2D(shard, device);
+    }
+    this->Done(shard, device, status.Success());
```
qyh111, line 72: Since we use Async copy here ,it seems we shouldn't synchronize every shard

ucm/store/infra/memory/memory_pool.h
```diff
@@ -0,0 +1,157 @@
...
+    char* GetFirstAddr() {
+        return pool_;
+    }
+
+private:
+    char* pool_ = nullptr;
+    size_t capacity_;
+    size_t blockSize_;
+
+    std::unordered_map<std::string, char*> addressMap_;
```
qyh111, line 106: didn't find any lock here

PR #303, opened 2025-10-21
[Feat] Adapt Trace Replay to vLLM >= 0.10.2

sumingZero:
# Purpose

Adjust imports due to benchmark module restructuring in vLLM 0.10.2+.

# Test
Tested on vLLM versions 0.9.2 and 0.11.0

image


ygwpz (approved):

benchmarks/gitkeep.txt
flesher0813, line 1: Do we need changing this file name?

ygwpz (approved):

PR #305, opened 2025-10-21
[Feat] add batch interface for device ops and implement `ScatterGather` with CUDA

mag1c-h:
# Purpose

Add batch interface for device ops and implement `ScatterGather` with CUDA.

```c
virtual Status H2DBatchSync(std::byte* dArr[], const std::byte* hArr[], const size_t number, const size_t count) = 0;
virtual Status D2HBatchSync(std::byte* hArr[], const std::byte* dArr[], const size_t number, const size_t count) = 0;
```

# Modifications 

No _any_ user-facing change.

# Test

CI passed with new added/existing test.

### DIFF
--- a/ucm/store/device/cuda/cuda_device.cu
+++ b/ucm/store/device/cuda/cuda_device.cu
@@ -25,6 +25,61 @@
 #include "ibuffered_device.h"
 #include "logger/logger.h"
 
+#define CUDA_TRANS_UNIT_SIZE (sizeof(uint64_t) * 2)
+#define CUDA_TRANS_BLOCK_NUMBER (32)
+#define CUDA_TRANS_BLOCK_SIZE (256)
+#define CUDA_TRANS_THREAD_NUMBER (CUDA_TRANS_BLOCK_NUMBER * CUDA_TRANS_BLOCK_SIZE)
+
+inline __device__ void H2DUnit(uint8_t* __restrict__ dst, const volatile uint8_t* __restrict__ src)
+{
+    uint64_t a, b;
+    asm volatile("ld.global.cs.v2.u64 {%0, %1}, [%2];" : "=l"(a), "=l"(b) : "l"(src));
+    asm volatile("st.global.cg.v2.u64 [%0], {%1, %2};" ::"l"(dst), "l"(a), "l"(b));
+}
+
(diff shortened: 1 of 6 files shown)


ucm/store/device/cuda/cuda_device.cu
```diff
@@ -103,14 +172,42 @@ class CudaDevice : public IBufferedDevice {
...
+        auto src = MakeDeviceArray((const void**)dArr, number);
+        if (!src) { return Status::OutOfMemory(); }
+        auto dst = MakeDeviceArray((const void**)hArr, number);
+        if (!dst) {
+            ReleaseDeviceArray(src);
+            return Status::OutOfMemory();
+        }
+        D2HBatch((volatile uintptr_t*)dst, (const uintptr_t*)src, number, count, this->stream_);
+        auto status = this->Synchronized();
+        ReleaseDeviceArray(src);
```
ChenyuZhu1, line 208: Do we need to consider the case where the release operation fails?

ucm/store/device/cuda/cuda_device.cu
```diff
@@ -103,14 +172,42 @@ class CudaDevice : public IBufferedDevice {
...
-        auto status =
-            CUDA_API(cudaStreamAddCallback, (cudaStream_t)this->stream_, Trampoline, (void*)c, 0);
+        auto status = CUDA_API(cudaStreamAddCallback, this->stream_, Trampoline, (void*)c, 0);
         if (status.Failure()) { delete c; }
         return status;
     }
-    Status Synchronized() override
+    Status Synchronized() override { return CUDA_API(cudaStreamSynchronize, this->stream_); }
+    Status H2DBatchSync(std::byte* dArr[], const std::byte* hArr[], const size_t number,
+                        const size_t count) override
```
qyh111, line 181: I’m a bit confused about the count parameter here. From what I understand, number represents the number of buffers. Does count correspond to the size of each buffer in bytes?

ucm/store/device/cuda/cuda_device.cu
mag1c-h: There is no scenario here where the release operation could fail.

ucm/store/device/cuda/cuda_device.cu
mag1c-h: Yes, `number` is the number of array elements, and `count` is how many characters to copy from each element.

PR #309, opened 2025-10-23
[Fix]Add import checking to trace_replay and fix the issue of unclose…

hero0307:
…d network resources

# Purpose

Add import checking functionality to trace_replay and fix the issue of unclosed network resources

# Modifications 

Add import checking functionality to trace_replay and fix the issue of unclosed network resources

# Test

You can run trace_replay.py. 
If you haven't installed openpyxl, you will be prompted to install it at the start of the run, instead of being reminded to install it at the end as before.
Meanwhile, the "unclosed client session" warning will no longer appear.

### DIFF
--- a/benchmarks/trace_replay.py
+++ b/benchmarks/trace_replay.py
@@ -540,13 +540,15 @@ async def replay_trace_by_time(
         flat_requests.extend(reqs)
 
     group_results = await asyncio.gather(*tasks)
+
+    if pbar is not None:
+        pbar.close()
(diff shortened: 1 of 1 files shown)


hero0307:
> Maybe using `await session.close()` at the end of func `replay_trace_by_time` can solve the "unclosed client session" warning?

If an exception occurs during execution when using this method, the session will not be closed. Using async with will automatically handle resource cleanup.

PR #312, opened 2025-10-24
[feat] hotness management for gc

Lijiachen1018:
# Purpose
Hotness management 
- add the ID of visited block to a set
- update the Modify time of block file in the set

# Modifications 
- `HotnessSet` that collect ID of blocks and update access time of file
- `HotnessTimer` that periodically calls `HotnessSet::UpdateHotness()`
- `HotnessManager` that manage the timer and set

# Test
test of access time updated when the blocks are fetched 
img_v3_02rg_8650be14-7b5c-4d7e-bc7c-f8a88773b71g

### DIFF
--- a/ucm/store/nfsstore/cc/api/nfsstore.cc
+++ b/ucm/store/nfsstore/cc/api/nfsstore.cc
@@ -26,6 +26,7 @@
 #include "logger/logger.h"
 #include "space/space_manager.h"
 #include "trans/trans_manager.h"
+#include "hotness/hotness_manager.h"
 
 namespace UC {
 
@@ -49,6 +50,13 @@ public:
                 return status.Underlying();
             }
         }
+        if (config.hotnessEnable) {
+            status = this->hotnessMgr_.Setup(config.hotnessInterval, this->spaceMgr_.GetSpaceLayout());
+            if (status.Failure()) {
+                UC_ERROR("Failed({}) to setup HotnessManager.", status);
+                return status.Underlying();
+            }
+        }
         this->ShowConfig(config);
         return Status::OK().Underlying();
     }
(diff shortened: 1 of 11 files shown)


ucm/store/nfsstore/cc/api/nfsstore.cc
```diff
@@ -49,14 +50,26 @@ class NFSStoreImpl : public NFSStore {
...
     int32_t Alloc(const std::string& block) override
     {
         return this->spaceMgr_.NewBlock(block).Underlying();
     }
-    bool Lookup(const std::string& block) override { return this->spaceMgr_.LookupBlock(block); }
+    bool Lookup(const std::string& block) override
+    {
+        auto found = this->spaceMgr_.LookupBlock(block);
+        this->hotnessMgr_.Visit(block);
+        return found;
```
mag1c-h, line 71: Only update the hotness when the block is actually found.

ucm/store/nfsstore/cc/api/nfsstore.cc
```diff
@@ -105,11 +118,14 @@ class NFSStoreImpl : public NFSStore {
         UC_INFO("Set UC::BufferNumber to {}.", config.transferBufferNumber);
         UC_INFO("Set UC::TimeoutMs to {}.", config.transferTimeoutMs);
         UC_INFO("Set UC::TempDumpDirEnable to {}.", config.tempDumpDirEnable);
+        UC_INFO("Set UC::HotnessInterval to {}.", config.hotnessInterval);
+        UC_INFO("Set UC::HotnessEnabled to {}.", config.hotnessEnabled);
```
mag1c-h: Use `Enable`, not `Enabled`, to keep naming consistent with other config keys.

ucm/store/nfsstore/cc/domain/hotness/hotness_manager.h
```diff
@@ -0,0 +1,76 @@
...
+        this->hotnessTimer_.SetInterval(interval);
+        this->layout_ = spaceLayout;
+        this->setupSuccess_ = true;
+        return Status::OK();
+    }
+    
+    void Visit(const std::string& blockId)
+    {
+        if (!this->setupSuccess_) {
+            UC_ERROR("HotnessManager is not setup.");
```
mag1c-h: Don’t log an error here, could simply mean the feature is disabled.

PR #323, opened 2025-10-29
[Feat]Add nfsstore bandwidth testing script

zhou-haitao:
# Purpose

Add nfsstore bandwidth testing script.

# Modifications 

Add nfsstore bandwidth testing script.

# Test

Add nfsstore bandwidth testing script.
image

### DIFF
--- a/ucm/store/test/e2e/nfsstore_embed_fetch.py
+++ b/ucm/store/test/e2e/nfsstore_embed_fetch.py
@@ -1,0 +1,282 @@
+# -*- coding: utf-8 -*-
+#
+# MIT License
+#
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+import csv
+import os
+import secrets
+import time
+from typing import Dict, List, Tuple
+
+import torch
+
+from ucm.store.nfsstore.nfsstore_connector import UcmNfsStore
+from ucm.store.ucmstore import UcmKVStoreBase
+
+
+def setup(
+    storage_backends, block_size, device_id, io_size, transferStreamNumber
+) -> UcmKVStoreBase:
+    config = {
+        "storage_backends": storage_backends,
+        "kv_block_size": block_size,
+        "role": "worker",
+        "device": device_id,
+        "io_size": io_size,
+        "transferStreamNumber": transferStreamNumber,
+    }
+    return UcmNfsStore(config)
+
+
+def make_buffers(
+    block_number, device_id, batch_size, head_dim, block_len, block_layer, num_head, kv
+):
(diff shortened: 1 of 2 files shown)


ucm/store/test/e2e/nfsstore_embed_fetch.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+def setup_store(storage_backends, block_size, device_id, io_size) -> UcmKVStoreBase:
+    config = {
+        "storage_backends": storage_backends,
+        "kv_block_size": block_size,
+        "role": "worker",
+        "device": device_id,
+        "io_size": io_size,
+        "stream_number": 128,
```
ygwpz: can we use different stream number like 32 64 128

zhou-haitao: Certainly, I'll modify this parameter into a list format right away.

ucm/store/test/e2e/nfsstore_embed_fetch.py
```diff
@@ -0,0 +1,311 @@
...
+                    io_num = int(num_tokens / block_len * block_layer)
+
+                    w_bw_list, r_bw_list = [], []
+                    w_time_list, r_time_list = [], []
+                    w_size_sum, r_size_sum = 0.0, 0.0
+
+                    for r in range(repeat):
+                        print(f"\n--- Round {r+1} ---")
+                        store = setup_store(
+                            storage_backends, block_size, device_id, io_size
```
mag1c-h: Calling setup multiple times inside a loop is undefined; the design guarantees it runs only once at process startup.

ygwpz: You can split the file into two Python files, one that accepts the specified connector type and config, and run a set of results, and the other that calls the file to generate a large table.

ucm/store/test/e2e/nfsstore_embed_fetch.py
```diff
@@ -0,0 +1,311 @@
...
+
+def make_buffers(
+    block_number, device_id, batch_size, head_dim, block_len, block_layer, num_head
+):
+    hashes = [secrets.token_hex(16) for _ in range(block_number)]
+    kv_caches = {}
+    for i in range(block_layer):
+        kv_caches[i] = torch.rand(
+            [1, block_number, block_len, num_head, head_dim],
+            dtype=torch.bfloat16,
```
qyh111, line 59: Why the first dim is 1? I believe the first dim is 2 when the model is Qwen Or it just doesn’t matter

zhou-haitao: This is the batch dimension placeholder. Setting it to 1 indicates processing only one batch of cached data at a time, which simplifies the testing process without affecting accuracy.

ucm/store/test/e2e/nfsstore_embed_fetch.py
zhou-haitao: The reason setup gets called multiple times is that the argument list is different on every launch. I think splitting the code into two Python files is worth trying, I'll go ahead and refactor it that way next.

ucm/store/test/e2e/nfsstore_embed_fetch.py
```diff
@@ -0,0 +1,311 @@
...
+        kv = 1
+        model_name = "deepseek-v3"
+        num_head_list = [1]
+    else:
+        block_lens = [128, 256]
+        block_layer = 64
+        head_size = 128
+        kv = 2
+        model_name = "QwQ-32B"
+        num_head_list = [1, 2, 4]
```
UESTC-AHao: Don't need TP=1? In a word, num_head = 8?

ucm/store/test/e2e/nfsstore_embed_fetch.py
ygwpz: > The reason setup gets called multiple times is that the argument list is different on every launch. I think splitting the code into two Python files is worth trying, I'll go ahead and refactor it that way next.

you should consider this problem:
image

ucm/store/test/e2e/nfsstore_embed_fetch.py
zhou-haitao: Added.

ucm/store/test/e2e/nfsstore_embed_fetch.py
zhou-haitao: After separating the current method and calls, the `setup` function can be called outside the 'repeat' loop to ensure that initialization is performed only once during the 'repeat' iterations. Other cases involving parameter changes cannot share a common setup method.

ygwpz: what if i just want to run part of cases?

zhou-haitao: It has been divided into two files, and you can modify the parameters to be used in the nfsstore_embed_fetch_run.py file.

ucm/store/test/e2e/nfsstore_embed_fetch.py
qyh111: When QwQ32B, batch size = 16 and block_size = 128, num_head = 1,it should 16 * 128 * 1 * 128  * 2 * 2 * 64 = 0.0625，but current script tells me 0.0312。I think we should consider k and v

zhou-haitao: You are right, I need to consider k and v in non-MLA scenarios.

PR #327, opened 2025-10-31
[bug fix] fix recycleNum when less than 1

Lijiachen1018:
# Purpose
bug fix in gc

# Modifications 
- fix `recycleNum` when `capacity * recycle_ratio < 1`
- pass recycle related parameters to connector
- fix property file name

# Test

test with some extra log print
image

### DIFF
--- a/ucm/store/nfsstore/nfsstore_connector.py
+++ b/ucm/store/nfsstore/nfsstore_connector.py
@@ -51,6 +51,12 @@ class UcmNfsStore(UcmKVStoreBase):
         if transfer_enable:
             param.transferDeviceId = config["device"]
             param.transferIoSize = config["io_size"]
+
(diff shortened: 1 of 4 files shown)


ucm/store/nfsstore/nfsstore_connector.py
```diff
@@ -51,6 +51,12 @@ def __init__(self, config: Dict):
         if transfer_enable:
             param.transferDeviceId = config["device"]
             param.transferIoSize = config["io_size"]
+
+        recycleEnable = True if config.get("recycleEnable", 0) == 1 else False
+        if recycleEnable:
+            param.storageCapacity = config["storageCapacity"]
+            param.kvcacheBlockSize = config["kvcacheBlockSize"]
```
mag1c-h: These parameters remain configurable even when GC is disabled.

PR #351, opened 2025-11-11
[Fix] Dump/load all tensors when use_layerwise=False

flesher0813:
# Purpose

Dump/load all tensors when use_layerwise=False, adapting to pcstore connector

# Modifications 

Users won't fell changes in using.
# Test

How was this patch tested?

Tested by tracereplay and offline script

### DIFF
--- a/ucm/integration/vllm/uc_connector.py
+++ b/ucm/integration/vllm/uc_connector.py
@@ -275,7 +275,7 @@ class UnifiedCacheConnectorV1(KVConnectorBase_V1):
             is_load_async = request.load_async
             total_offsets = []
             total_tensors = []
-            storage_block_ids = storage_block_ids * 2 ** (0 if self.is_mla else 1)
+            storage_block_ids = storage_block_ids * (1 if self.is_mla else 2)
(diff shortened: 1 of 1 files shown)


ucm/integration/vllm/uc_connector.py
```diff
@@ -261,62 +262,43 @@ def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
...
             storage_block_ids = [block[0] for block in request.load_blocks]
             vllm_block_ids = [block[1] for block in request.load_blocks]
-            blocks_len = len(storage_block_ids)
             self._load_req_to_blocks.setdefault(request.request_id, set()).update(
                 vllm_block_ids
             )
+            is_load_async = request.load_async
+            total_offsets = []
+            total_tensors = []
+            storage_block_ids = storage_block_ids * 2 ** (0 if self.is_mla else 1)
```
ygwpz: storage_block_ids = storage_block_ids *  (1 if self.is_mla else 2)

flesher0813: fixed

PR #359, opened 2025-11-16
switch tensor to ptr & add block_threshold

qyh111:
# Purpose

# Modifications 

# Test

How was this patch tested?


ucm/integration/vllm/uc_connector.py
```diff
@@ -206,24 +219,19 @@ def DataOffset(self, kv_layer, rank, layer_id, is_v):
 
     def get_tensor_and_offset_layerwise(
         self, vllm_block_ids: List[int], kv_layer: torch.Tensor, layer_name: str
-    ) -> tuple[List[torch.Tensor], List[int]]:
+    ) -> tuple[List[int], List[int]]:
```
harrisonyhq: If you change k/v tensors list from list[tensor] to list[ptr], how can broadcast get tensor from ptr?

PR #377, opened 2025-11-20
[Feat] Support get launch config from yaml

harrisonyhq:
# Purpose

Originlly, when you want to launch inference by ucm, you need to configure lots of params in command line, which is not friendly for usage. Now we support launch from config yaml, meanwhile, still support the original launch style.

# Modifications 

Users can launch from yaml config file. for more details please refer to documents.

# Test

No need test.


ucm/utils.py
```diff
@@ -0,0 +1,90 @@
...
+from typing import Any, Dict
+
+import yaml
+
+from ucm.logger import init_logger
+
+logger = init_logger(__name__)
+
+
+class UCMConfig:
```
ygwpz: don't use UCM in internal namespace

harrisonyhq: Already fixed

PR #389, opened 2025-11-21
[Fix]Set the multiprocessing start method of the test tool to 'spawn'

zhou-haitao:
# Purpose

Set the multiprocessing start method of the test tool to 'spawn'

# Modifications 

Set the multiprocessing start method of the test tool to 'spawn'

# Test

Set the multiprocessing start method of the test tool to 'spawn'

### DIFF
--- a/ucm/store/test/e2e/nfsstore_embed_fetch.py
+++ b/ucm/store/test/e2e/nfsstore_embed_fetch.py
@@ -271,9 +271,10 @@ def run(
             if r == 0:
                 store_all_hashes(hashes[:batch_size])
 
-            w_bw_list.append(w_bw)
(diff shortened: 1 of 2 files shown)


ucm/store/test/e2e/nfsstore_embed_fetch.py
```diff
@@ -316,10 +317,12 @@ def run(
 
             if operation_mode == "read_only":
                 del kvcaches
-                torch.cuda.empty_cache()
+                if torch.cuda.is_available():
```
ygwpz, line 324: how about ascend?

zhou-haitao: Already modified.

PR #413, opened 2025-11-26
[Feat] Support record operations

flesher0813:
# Purpose

Support users save operation logs to replay trace. When enable using operation db, all load/save operations would be saved in target log_dir.

# Modifications 

Support replay traces, and users can modify the yaml file to meet their own needs.

# Test

How was this patch tested?

image
image

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -1,11 +1,17 @@
 import hashlib
 import itertools
+import json
 import os
 import pickle
 from dataclasses import dataclass, field
+import queue
+import threading
+import time
 from typing import TYPE_CHECKING, Callable, List, Optional
 
+from sympy import Dict
 import torch
+from transformers import Any
 from vllm.config import VllmConfig
 from vllm.distributed.kv_transfer.kv_connector.v1.base import (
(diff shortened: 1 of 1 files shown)


ucm/shared/metrics/async_op_db.py
```diff
@@ -0,0 +1,297 @@
...
+
+from ucm.logger import init_logger
+
+logger = init_logger(__name__)
+
+
+class AsyncUCMOperDB:
+    """Asynchronous UCM operation log writer (supports log compression + auto-cleanup)"""
+
+    def _load_config(self, config_path: str) -> Dict[str, Any]:
```
ygwpz: can we use unified config.py? because root dir also need config.yaml

ucm/integration/vllm/ucm_connector.py
```diff
@@ -160,6 +160,8 @@ def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
             config["kv_block_size"] / 1024 / 1024,
             config["io_size"] / 1024,
         )
+        if role == KVConnectorRole.WORKER and get_tp_group().is_first_rank:
```
ygwpz: we need a param to control whether to enable this function.

PR #416, opened 2025-11-27
[docs]add metrics doc

hero0307:
# Purpose

UCM has added a new metrics feature, including Prometheus Metrics and Operation Logging. Documentation for the metrics feature has been supplemented to introduce the functionalities and usage of Prometheus Metrics and Operation Logging.

# Modifications 
Add unified-cache-management/docs/user-guide/metrics

# Test

image
image
image
image


ygwpz:
don't use so many "ucm" prefix, whether to use "ucm"  depending on if we expose api to external

flesher0813 (approved):

PR #418, opened 2025-11-27
[Feature]Add an access bandwidth test script for ucm_connector

zhou-haitao:
# Purpose

Add an access bandwidth test script for ucm_connector

# Modifications 

Add an access bandwidth test script for ucm_connector

# Test

image

### DIFF
--- a/test/test_ucm_connector_save_load.py
+++ b/test/test_ucm_connector_save_load.py
@@ -390,12 +390,14 @@ def run_test(
         (w_size, w_time, w_bw), (r_size, r_time, r_bw) = run_once(
             connector, kv_caches, round_hashes, batch_size, mla
         )
-        w_sizes.append(w_size)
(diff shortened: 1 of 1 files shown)


test/test_ucm_connector_save_load.py
```diff
@@ -0,0 +1,585 @@
...
+from vllm.distributed.kv_transfer.kv_connector.v1.base import KVConnectorRole
+
+from ucm.integration.vllm.ucm_connector import (
+    RequestDispatchMeta,
+    UCMConnector,
+    UCMConnectorMetadata,
+)
+
+
+def log(msg: str):
```
ygwpz: why not use ucm logger?

zhou-haitao: It has been changed to ucm logger.

PR #420, opened 2025-11-27
[bugfix]Support tensor parallelism across servers

qyh111:
# Purpose

# Modifications 

# Test
image
image

How was this patch tested?


ucm/integration/vllm/ucm_connector.py
```diff
@@ -101,16 +102,16 @@ def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
         else:
             raise RuntimeError("Unsupported device platform for UCMDirectConnector.")
 
-        if self.rank >= 0:
-            self.device = torch_dev.device(f"{dev_name}:{self.rank}")
+        if self.local_rank >= 0:
+            self.device = torch_dev.device(f"{dev_name}:{self.local_rank}")
```
ygwpz, line 106: global rank can get from vllmConfig

PR #424, opened 2025-11-28
[Feature] Add performance and evaluation testing tools using the pytest framework

zzycode1005:
# Purpose
Added a new testing tools supporting both performance and evaluation.

# Modifications
1、Supports TTFT metrics for all requests and for steady-state requests only.
2、Support accuracy and F1-score calculation in document-QA scenarios.
3、Support saving per-case performance and evaluation results via Excel.

# Test
performance test success
eval test success


test/suites/E2E/test_uc_performance.py
```diff
@@ -156,3 +165,102 @@ def test_performance(
...
+            parallel_num=1,
+            benchmark_mode="default-perf",
+        ),
+        id="multiturn-dialogue-complete-recalculate-default-perf",
+    )
+]
+
+
+@pytest.mark.feature("perf_test")
+@pytest.mark.parametrize("perf_config", multiturn_dialogue_perf_cases)
```
Wwwzff, line 234: maybe add a stage mark here

PR #467, opened 2025-12-04
[Feature] Cache Blend 

wuhuxiao:
# Purpose
Context Independent Cache (Chunk Cache) is all you need

# Modifications 

# Test
ENABLE_SPARSE=True
DATA_DIR=/home/data/kv_cache
MODEL_PATH=/home/models/mistralai/Mistral-7B-Instruct-v0.2
BLEND_DATASET_PATH=/home/datasets/LongBench/data/2wikimqa.jsonl
image

How was this patch tested?

### DIFF
--- a/examples/offline_inference_blend.py
+++ b/examples/offline_inference_blend.py
@@ -169,10 +169,7 @@ def pad_rag_chunks(token_ids, block_size, pad_id, end_id):
     return padded
 
 
-systemPrompt = """
-    You are a helpful assistant.
-    Please read the following Passages and answer the Question below.
-"""
+systemPrompt = "Answer the question based on the given passages. Only give me the answer and do not output any other words.\n\nThe following are given passages.\n"
 
 
 def main():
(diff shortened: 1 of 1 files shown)


ucm/integration/vllm/blend_connector.py
```diff
@@ -0,0 +1,508 @@
...
+class BlendRequestDispatchMeta(RequestDispatchMeta):
+    chunks_meta: List[ChunkMetaData]
+
+
+@dataclass
+class UCMBlendConnectorMetadata(UCMConnectorMetadata):
+    request_meta: dict[str, BlendRequestDispatchMeta] = field(default_factory=dict)
+
+
+class UCMBlendConnector(UCMDirectConnector):
```
ygwpz, line 124: it seem's no need to inherit from UCMDirectConnector but from KVConnectorBase_V1

wuhuxiao: the kv blocks load and dump are now implemented in UCMDirectConnector 's store. If there are contiguous kvcache load and dump api in kv store, we will inherit from KVConnectorBase_V1 and build specified kvcache dump, lookup, load functions

PR #485, opened 2025-12-08
[bugfix]add synchronize on ascend platform

qyh111:
# Purpose
add synchronize on ascend platform otherwise there will be accuracy problem.
But it has been fixed in the latest main branch of vllm_ascend, so synchronize will no longer be needed in future versions.

# Modifications 

# Test

How was this patch tested?

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -559,7 +559,9 @@ class UCMDirectConnector(KVConnectorBase_V1):
         # TODO support PP
         if (self.is_mla or self.is_dsa) and self.global_rank != 0:
             return
-        if self.metrics_config:
+        if self.metrics_config or current_platform.device_type == "npu":
+            # When use vllm_ascend, we should add synchronize here, otherwise accuracy problem will raise
+            # This has already been fixed in the latest main branch of vllm_ascend, so synchronize will no longer be needed in future versions.
             self.synchronize()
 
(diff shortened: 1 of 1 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -29,6 +28,7 @@
     from vllm.attention.backends.abstract import AttentionMetadata
     from vllm.forward_context import ForwardContext
     from vllm.v1.core.kv_cache_manager import KVCacheBlocks
+    from vllm.v1.request import Request
```
ygwpz, line 31: unused import for "Request"?

ucm/integration/vllm/ucm_connector.py
```diff
@@ -570,6 +570,9 @@ def wait_for_save(self) -> None:
         num_saved_block = 0
         num_saved_request = 0
         save_start_time = time.perf_counter() * 1000
+        # This has already been fixed in the latest main branch of vllm_ascend, so synchronize will no longer be needed in future versions.
+        if current_platform.device_type == "npu":
+            torch.npu.current_stream().synchronize()
```
ygwpz: use self.synchronize() and put it together with line 561.

ucm/integration/vllm/ucm_connector.py
qyh111: It will cause circular reference when use git apply. we can still use request obj in the runtime, it will not raise problems

PR #486, opened 2025-12-08
[Feat] Support metrics across language boundary

flesher0813:
# Purpose

This pull request (PR) supports collecting different statistics aggregated by the monitor component from both the Python side (e.g., ucm_connector) and the C++ side (e.g., various stores), and provides a unified interface for users to retrieve such statistics.
Therefore, this PR modifies the current metrics structure in three aspects:
- Implements the Python-side statistics and decouples the implementation from the framework structure.
- Exposes the interfaces required for registering, creating, updating, retrieving, and clearing statistics via global functions, and exposes these interfaces to Python through pybind11 to ensure that both languages can access the same instance within a single process.
- Provides the examples and documentation.
# Modifications 

Users can add or remove the statics to be collected according to the documentation for custom metrics.
# Test

How was this patch tested?

Tested with
```
ucm/shared/test/case/metrics/monitor_test.cc
ucm/shared/test/example/metrics/monitor_stats_example.py
```
image
image

doc
image

### DIFF
--- a/docs/source/developer-guide/add_metrics.md
+++ b/docs/source/developer-guide/add_metrics.md
@@ -3,76 +3,113 @@ UCM supports custom metrics with bidirectional updates from both Python and C++ 
 
 ## Architecture Overview
 The metrics consists of these components below:
-- **monitor** : Central stats registry that manages all metric lifecycle operations (registration, creation, updates, queries)
+- **metrics** : Defined in cpp. Central stats registry that manages all metric lifecycle operations (creation, updates, queries)
 - **observability.py** : Prometheus integration layer that handles metric exposition
 - **metrics_config.yaml** : Declarative configuration that defines which custom metrics to register and their properties
 
 ## Getting Started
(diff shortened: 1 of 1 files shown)


ucm/shared/test/example/metrics/monitor_stats_example.py
```diff
@@ -28,23 +28,17 @@
 import sys
 
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+from ucm.integration.vllm.conn_stats import ConnStats
```
ygwpz: The dependency principle is violated. Advised to place the example in the root directory.

ucm/observability.py
ygwpz: unused import

ygwpz: naming style must be consistent.

ygwpz: config_path seems to be "config" itself, type should be "Dict"

PR #489, opened 2025-12-09
[Feature] Added environment pre-check

Menglths:
# Purpose
To avoid manual errors, an automated pre-check for the initial setup is needed. This helps detect issues early, improve inference stability, and prevent faults from configuration errors.

# Modifications 
Added configurations for SSH password-free login, health status checks of hardware devices, network connectivity between nodes, TLS verification between nodes, model weight file integrity and consistency, and storage mount point rate checks.

- Added test/common/envprecheck/run_env_precheck.py

- Added test/suites/E2E/test_environment_precheck.py

- Modified test/config.yaml

- Modified test/suites/E2E/test_uc_performance.py

# Test
Verified each pre-check function runs correctly. Assertions for overall status are preserved and working.
image

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,6 +2,7 @@ include LICENSE
 include pyproject.toml
 include CMakeLists.txt
 include requirements.txt
+include setup.py
 
 recursive-include examples *
 recursive-include benchmarks *
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,8 +5,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "uc-manager"
 authors = [{name = "UCM Team"}]
-license = "MIT"
-license-files = ["LICENSE"]
+license = { file="LICENSE" }
 readme = "README.md"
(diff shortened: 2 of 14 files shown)


test/suites/E2E/test_environment_precheck.py
```diff
@@ -0,0 +1,238 @@
...
+    run_device_status_check,
+    run_hccn_tls_check,
+    run_set_ssh_login,
+)
+
+
+# ========= Environment Setup =========
+@pytest.mark.stage(0)
+@pytest.mark.feature("test_ssh_login")
+@export_vars
```
Wwwzff: this decorator is only for internal usage and we may not need to update external database while doing precheck things

test/suites/E2E/test_environment_precheck.py
```diff
@@ -0,0 +1,238 @@
+import os
```
Wwwzff: A guide doc is also needed to let users run this test fixture easily

PR #498, opened 2025-12-10
[Feature] Added environment pre-check

Menglths:
# Purpose
To avoid manual errors, an automated pre-check for the initial setup is needed. This helps detect issues early, improve inference stability, and prevent faults from configuration errors.

# Modifications 
Added configurations for SSH password-free login, health status checks of hardware devices, network connectivity between nodes, TLS verification between nodes, model weight file integrity and consistency, and storage mount point rate checks.

- Added test/common/envprecheck/run_env_precheck.py

- Added test/suites/E2E/test_environment_precheck.py

- Modified test/config.yaml

- Modified test/suites/E2E/test_uc_performance.py

# Test

How was this patch tested?
Verified each pre-check function runs correctly. Assertions for overall status are preserved and working.
image


Wwwzff (approved):
LGTM

PR #506, opened 2025-12-11
[Feature] add accuracy test

zhou-haitao:
# Purpose

add accuracy test

# Modifications 

add accuracy test

# Test

add accuracy test
image


Wwwzff:
There's a similar PR merged days ago and it seems better to do accurary tests by that test fixture https://github.com/ModelEngine-Group/unified-cache-management/pull/462

PR #550, opened 2025-12-18
[Perf] parallel block-existence checks + timeout exception

mag1c-h:
## Summary

1. **Parallelize block-existence checks** to cut latency on bulk operations.  
2. **Add timeout guards** that raise a Python exception when the deadline is exceeded.

## Details

- Issue: sequential existence checks become a bottleneck when the caller asks for hundreds/thousands of blocks.  
- Fix: dispatch all `Lookup` calls to a thread-pool and wait for the batch with a configurable timeout.  
- Behavior change: if any single check or the whole batch exceeds `timeout_ms`, a `pybind11::timeout_error` is propagated to the Python side.

### DIFF
--- a/ucm/store/pcstore/cpy/pcstore.py.cc
+++ b/ucm/store/pcstore/cpy/pcstore.py.cc
@@ -47,11 +47,11 @@ public:
     {
         auto ret = Setup(config);
         if (config.transferEnable || ret != Status::OK().Underlying()) { return ret; }
-        auto success =
-            lookupService_.SetNWorker(4)
-                .SetWorkerFn([this](auto& ctx, auto) { OnLookup(ctx); })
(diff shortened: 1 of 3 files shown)


ucm/store/pcstore/cpy/pcstore.py.cc
```diff
@@ -24,25 +24,61 @@
...
 public:
     void* CCStoreImpl() { return this; }
+    int32_t SetupPy(const Config& config)
+    {
+        auto ret = Setup(config);
+        if (config.transferEnable || ret != Status::OK().Underlying()) { return ret; }
+        auto success =
+            lookupService_.SetNWorker(4)
+                .SetWorkerFn([this](auto& pair, auto) { OnLookup(pair.first, pair.second); })
+                .SetWorkerTimeoutFn([this](auto&, auto) { OnLookupTimeouted(); }, 10000)
```
Lijiachen1018: put the timeout to configuration

PR #556, opened 2025-12-19
[Feat] Basic scripts for deployment best practices

sumingZero:
# Purpose

Provide basic scripts and corresponding documentation for best practices.

# Usage example
## 1. Start the ray server of master node.
image

## 2. Start the ray server of the first worker node.
image

## 3. Start the vllm server in master node.
image

### DIFF
--- a/examples/deployments/scripts/vllm/run_vllm.sh
+++ b/examples/deployments/scripts/vllm/run_vllm.sh
@@ -1,0 +1,112 @@
+#!/bin/bash
+echo $CONFIG_FILE
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$SCRIPT_DIR/common.sh"
+
+start_server() {
+    [[ -z "$model" ]] && { echo "ERROR: model not set in config.properties" >&2; exit 1; }
+
+    if [[ "$ucm_enable" == "true" ]]; then
+        [[ -z "$ucm_config_yaml_path" ]] && {
+            echo "ERROR: ucm_config_yaml_path not set but ucm_enable=true" >&2
+            exit 1
+        }
+        LOG_FILE="vllm_ucm.log"
+    else
+        LOG_FILE="vllm.log"
+    fi
+
+    echo ""
+    echo "===== vllm server configuration ====="
+    echo "model                    = $model"
+    echo "served_model_name        = ${served_model_name:-<default>}"
+    echo "tp_size                  = $tp_size"
+    echo "dp_size                  = $dp_size"
+    echo "pp_size                  = $pp_size"
+    echo "enable_expert_parallel   = $enable_expert_parallel"
+    echo "max_model_len            = $max_model_len"
+    echo "max_num_batched_tokens   = $max_num_batch_tokens"
+    echo "max_num_seqs             = $max_num_seqs"
+    echo "block_size               = $block_size"
+    echo "gpu_memory_utilization   = $gpu_memory_utilization"
+    echo "quantization             = $quantization"
+    echo "server_host              = $server_host"
+    echo "server_port              = $server_port"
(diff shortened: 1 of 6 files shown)


docs/source/user-guide/deployment/deployment.md
```diff
@@ -0,0 +1,104 @@
...
+
+#### Multi-Machine Deployment (Ascend)
+
+Ascend multi-node deployments differ based on whether **Data Parallelism (DP)** is enabled.
+
+##### Case 1: DP = 1 (No Data Parallelism)
+
+This case follows the same procedure as CUDA multi-machine deployment and requires the following files:
+- `vllm/config.properties`
+- `vllm/run_vllm.sh和vllm`
```
flesher0813: Different languages.

examples/deployments/scripts/vllm/run_vllm.sh
```diff
@@ -0,0 +1,145 @@
...
+
+    if [[ -n $max_num_seqs ]]; then CMD+=("--max-num-seqs" "$max_num_seqs"); fi
+
+    if [[ "$enable_expert_parallel" == "true" ]]; then CMD+=("--enable-expert-parallel"); fi
+
+    if [[ "$enable_prefix_caching" == "false" ]]; then CMD+=("--no-enable-prefix-caching"); fi
+
+    if [[ "$async_scheduling" == "true" ]]; then CMD+=("--async-scheduling"); fi
+
+    [[ -n "$served_model_name" ]] && CMD+=("--served-model-name" "$served_model_name")
```
flesher0813: Different methods to add new args for CMD， can we use same method to add this arg? And also too much if/then, can we collect all args like below to make it easy to add more new args?
`
ARG_TEMPLATES=(
  "--block-size block_size"
  "--max-num-batched-tokens max_num_batched_tokens"
  "--max-num-seqs max_num_seqs"
  "--seed seed"
)
`

examples/deployments/scripts/vllm/start_ray.sh
```diff
@@ -0,0 +1,131 @@
...
+    echo "local_ip                 = $TARGET_IP"
+    echo "network_interface        = $IFACE"
+    echo "num_gpus/npus (per node)      = $NUM_GPUS"
+    echo "CUDA_VISIBLE_DEVICES     = $CUDA_VISIBLE_DEVICES"
+    echo "ASCEND_RT_VISIBLE_DEVICES= $ASCEND_RT_VISIBLE_DEVICES"
+    echo "======================================"
+    echo ""
+}
+
+load_config
```
flesher0813, line 48: Seems like every bash script need function _load_config_, can we make it a common function? And also for other function like _ensure_ifconfig_installed_

docs/source/user-guide/deployment/deployment.md
sumingZero: This error has been fixed.

examples/deployments/scripts/vllm/start_ray.sh
sumingZero: Common code has been extracted into common.sh.

examples/deployments/scripts/vllm/run_vllm.sh
sumingZero: vLLM has too many parameter types, and even using a parameter template doesn't make it significantly more extensible.

flesher0813: Or some common parameters can be collected to a func?

ygwpz (approved):

examples/deployments/scripts/vllm/run_vllm.sh
sumingZero: This version is being shared now to gather everyone's feedback on the user experience. If the feedback isn't positive, we'll refactor it later using YAML + Python.

ygwpz (approved):

PR #569, opened 2025-12-23
[fix]Add exception handling

qyh111:
# Purpose
Add handling for exceptions thrown by the store interface

# Modifications 
Add handling for exceptions thrown by the store interface
remove unused function

# Test

How was this patch tested?

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -212,7 +212,6 @@ class UCMDirectConnector(KVConnectorBase_V1):
             config["block_size"] = chunk_block_size
             if self.is_dsa or self.is_mla:
                 config["share_buffer_enable"] = True
-                config["local_rank_size"] = max(self.tp_size, 8)
             else:
                 config["share_buffer_enable"] = False
             store = UcmConnectorFactoryV1.create_connector(name, config)


ygwpz (approved):

ucm/integration/vllm/ucm_connector.py
```diff
@@ -302,8 +302,11 @@ def get_num_new_matched_tokens(
         external_block_ids = ucm_block_ids[hbm_hit_block_num:]
         if not external_block_ids:
             return 0, False
-
-        lookup_results = self.store.lookup(external_block_ids)
+        try:
+            lookup_results = self.store.lookup(external_block_ids)
+        except RuntimeError as e:
+            lookup_results = []
+            logger.error(f"request {request.request_id} look up error. {e}")
```
ygwpz, line 308: directly return？

ucm/integration/vllm/ucm_connector.py
```diff
@@ -212,7 +212,7 @@ def _create_store(
             config["block_size"] = chunk_block_size
             if self.is_dsa or self.is_mla:
                 config["share_buffer_enable"] = True
-                config["local_rank_size"] = self.tp_size
+                config["local_rank_size"] = max(self.tp_size, 8)
```
mag1c-h: Why do we need to limit `local_rank_size` to 8? Shouldn't the number of rank cards on a single host node be greater than 8?

ucm/integration/vllm/ucm_connector.py
qyh111: when lookup error, lookup_results becomes [], then connector will dump those blocks

ucm/integration/vllm/ucm_connector.py
qyh111: suppose there're two nodes, use tp=16，then local_rank_size would be 16 rather than 8.

PR #572, opened 2025-12-24
[fix]fix monitor issue

qyh111:
# Purpose
When push task ,worker thread and monitor thread should both wake up, but Push use notify_one()， so monitor wake up but worker didn't  so lookup will stunk.

# Modifications 

monitor  doesn't need wake up when new task come in , it just sleep 1 second

# Test
How was this patch tested?
image

image


mag1c-h (approved):
LGTM

PR #582, opened 2025-12-25
[bugfix]include sparse/kvcomp/configs

qyh111:
# Purpose
include sparse/kvcomp/configs

# Modifications 

# Test
image

How was this patch tested?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -147,5 +147,6 @@ setup(
     ext_modules=[CMakeExtension(name="ucm", source_dir=ROOT_DIR)],
     cmdclass={"build_ext": CMakeBuild},
     zip_safe=False,
-    include_package_data=False,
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -147,5 +147,6 @@ def build_cmake(self, ext: CMakeExtension):
     ext_modules=[CMakeExtension(name="ucm", source_dir=ROOT_DIR)],
     cmdclass={"build_ext": CMakeBuild},
     zip_safe=False,
-    include_package_data=False,
+    include_package_data=True,
+    package_data={"ucm": ["sparse/kvcomp/configs/**/*.json"]},
```
mag1c-h, line 151: When `ENABLE_SPARSE` not to set or set to false, do we need these files?

PR #587, opened 2025-12-25
[optimize]Enhancements to LLMPerf Test Suite and Pytest Integration

Potterluo:
**Description:**

This PR introduces several improvements to the LLMPerf test suite based on real-world usage feedback and aims to simplify maintenance and configuration:

1. **Optimized LLMPerf Test Cases**  
   Updated and refined existing test cases according to actual usage patterns and performance testing scenarios to better reflect real-world conditions.

2. **Simplified Parameter Capture**  
   Refactored the parameter capture logic to reduce code duplication and overall complexity, improving readability and maintainability.

3. **Added `extra_info` Configuration Option**  
   Introduced a new `extra_info` field in the configuration to allow users to annotate or tag service launch parameters (e.g., model name, hardware specs, or custom flags), facilitating better traceability and debugging.

4. **Updated Documentation for Pytest Integration**  
   Revised and expanded the documentation to properly reflect how to run LLMPerf tests under the pytest framework, including examples and best practices.

These changes collectively enhance test robustness, reduce boilerplate, and improve the developer experience when integrating LLMPerf into automated testing pipelines.


test/suites/E2E/test_uc_performance.py
```diff
@@ -11,160 +12,83 @@
 )
 from common.uc_eval.utils.data_class import ModelConfig, PerfConfig
 
+perf_scenarios = [
+    # (mean_in, mean_out, max_req, concurrent, sampling, hit_rate)
+    (4000, 1024, 1, 1, "{}", 80),
+    (4000, 1024, 8, 8, "{}", 80),
+]
+scenario_ids = [f"in_{s[0]}-out_{s[1]}-con_{s[3]}" for s in perf_scenarios]
```
pyxyzc, line 20: It looks like params max_req and concurrent are always considered to be consistent. Plz consider whether this assumption makes sense.

test/suites/E2E/test_uc_performance.py
```diff
@@ -11,160 +12,83 @@
...
+
+    # 构造扁平化的结果字典，方便后续分析和看板展示
+    metrics = {
+        # 输入指标
+        "input_tokens": in_tokens,
+        "output_tokens": out_tokens,
+        "concurrent": concurrent,
+        "sum_requests": max_req,
+        "hit_rate": hit_rate,
+        "ttft_mean": results.get("ttft_s", {}).get("mean"),
```
pyxyzc, line 50: Sometimes we want to see the TTFT and TPOT for all combinations, not just the metrics. Plz consider adding a switch for this :)

test/suites/E2E/test_uc_performance.py
Potterluo: max_req is the total number of requests, and concurrent is the number of concurrent requests per round. Currently, we are concerned with scenarios involving multiple concurrent requests per round, so these two values are consistent. We will optimize these input parameters. It seems more reasonable to replace max_req with the number of request rounds.

Wwwzff: > It looks like params max_req and concurrent are always considered to be consistent. Plz consider whether this assumption makes sense.

there might be cases that have a 16 max cocurrency and 32 requests in total. In that case it's 32 max_req and 16 for concurrent

test/suites/E2E/test_uc_performance.py
Potterluo: I think you might have misunderstood. We will store the test results for all combinations of mean_in, mean_out, max_req, concurrent, sampling, and hit_rate in the database. "mean" refers to the test results within a single combination, which comes from LLMPerf. In fact, all data will be saved.

test/suites/E2E/test_uc_performance.py
Wwwzff, line 20:
```
scenario_ids = [f"in_{s[0]}-out_{s[1]}-con_{s[3]}" for s in perf_scenarios]
```
maybe we can hide this param from user and generate a unique id for each case automatically

PR #636, opened 2026-01-08
[feature]Add LLM connection base components and OpenAI connector

Potterluo:
This PR introduces the foundational building blocks for integrating large language models (LLMs) into the system, with a focus on modularity, type safety, and support for both streaming and non-streaming interactions.

Highlights:
Core Data Models: Added LLMRequest, LLMResponse, and LLMStreamChunk as immutable dataclasses to standardize input/output formats across LLM integrations.
Protocol Abstraction: Defined LLMConnection as a Python protocol to enforce a consistent interface (sync/async chat methods) for any LLM backend.
OpenAI Integration: Implemented OpenAIConn, a fully featured connector that supports synchronous/asynchronous invocation and both streaming and non-streaming modes via the OpenAI-compatible API.
Token Management: Integrated HuggingFaceTokenizer to enable accurate token counting and utility functions like random text generation based on tokenizer vocabulary.
Streaming Support: Built an SSE (Server-Sent Events) parser that converts OpenAI’s streaming response format into the standardized LLMStreamChunk structure.
HTTP Layer: Leveraged httpx (sync and async clients) for robust and efficient HTTP communication with LLM endpoints.
Security & Utilities: Added helper functions for secure token encoding/decoding and request ID validation to ensure integrity and traceability.
This lays the groundwork for pluggable LLM backends—future connectors (e.g., Anthropic, Ollama, or custom endpoints) can implement the LLMConnection protocol and reuse the shared infrastructure.
token_utils test
image
The calculated token is consistent with the engine's return
image
image
Streaming token calculation method
image


test/common/llm_connection/token_counter.py
```diff
@@ -0,0 +1,81 @@
...
+        except Exception as exc:
+            raise TokenizerError("Token count failed") from exc
+
+    def get_some_tokens(self, num_tokens: int) -> str:
+        if num_tokens <= 0:
+            return ""
+        safe_ids = self._get_or_build_safe_ids()
+        selected = random.choices(safe_ids, k=num_tokens)
+        text = self._tok.decode(selected, skip_special_tokens=False)
+        truncated = self._tok.encode(text, add_special_tokens=False)[:num_tokens]
```
Wwwzff: special tokens are needed in encoding or token length may not be that accurate

test/common/llm_connection/openai_connector.py
```diff
@@ -0,0 +1,203 @@
...
+        if self.api_key:
+            headers["Authorization"] = f"Bearer {self.api_key}"
+        return headers
+
+    # ---------- Synchronous methods ----------
+    def chat(self, req: LLMRequest, **kwargs) -> LLMResponse:
+        """Performs a synchronous non-streaming chat completion."""
+        body = self._make_body(req)
+        body["stream"] = False
+        with httpx.Client(base_url=self.base_url, timeout=self.timeout) as client:
```
Wwwzff: may be better if we catch the TimeoutException here and make a notice

PR #647, opened 2026-01-14
[Opt] Gsa npu performance optimize

leideng:
Improve performance for GSAOnDevice in NPU


ucm/sparse/gsa_on_device/gsa_on_device.py
```diff
@@ -546,8 +591,15 @@ def attention_finished(
                     )
                     attn_metadata.decode.num_splits = self.origin_num_splits
         else:  # 判断req decode阶段
-            if self.decode_mask.any():
-                attn_metadata.block_table = self.ori_block_table_decode
+            if (
+                self.decode_mask.any()
+                and int(layer_name.split(".")[2])
```
Infinite666: Why add a check to see if it's the last layer?

leideng: Nice observation. The reason is that we only need to recover `attn_metadata.block_table`  and `attn_metadata.seq_lens` right before next decoding step. During the current deconding step, all layers does not need the original block table. Thus, we only need to recover  it in the last layer's `attention_finished()` function.

Infinite666 (approved):

PR #654, opened 2026-01-16
[Feat] Support C++/Python to use same metrics singleton within a process

flesher0813:
# Purpose

- This PR aims to resolve the problem where Python-side and C++-side components within the same process fail to share the same metrics collection component. 
- Additionally, it refactors the original structure of the Metrics component to simplify the implementation logic and improve maintainability.

# Modifications 

Users now can add new metrics more easily with a provided doc to show how to add new metrics.

# Test

How was this patch tested?

Tested with H20-QwQ-32B-TP2 All Save and All Load
image

The add new metrics doc:
image

Test with cmake: only cost 1us on UpdateStats, and for GetStatsAndClear it would take more time if there is too much stats, we use another thread to handle it so it would not infect main process
image

Test with performance tests: cost stays below 1% in most cases; larger block sizes lead to smaller impact, which is even lower than the system's inherent fluctuations.
image


ygwpz (dismissed):

ucm/integration/vllm/ucm_connector.py
```diff
@@ -143,12 +143,11 @@ def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
 
         self.metrics_config = self.launch_config.get("metrics_config_path", "")
         if self.metrics_config:
-            self.stats_logger = UCMStatsLogger(
+            self.stats_logger = PrometheusStatsLogger(
```
ygwpz: i suggest to split ucm_connector change into another pr.

flesher0813: Split ucm_connector change to https://github.com/ModelEngine-Group/unified-cache-management/pull/658

ygwpz (dismissed):

ucm/shared/test/case/metrics/metrics_perfomance_test.cc
```diff
@@ -0,0 +1,177 @@
...
+
+    // Print result
+    std::cout << "\n===== Test Results: thread num=" << thread_count
+              << ", get interval=" << get_interval_ms << " =====" << std::endl;
+    std::cout << "UpdateStats Total Calls: " << result_.total_calls << std::endl;
+    std::cout << "Backend GetStatsAndClear Calls: " << result_.total_get_calls << std::endl;
+    std::cout << "Total Running Time: " << result_.total_time_s << " s" << std::endl;
+    std::cout << "Avg UpdateStats Time: " << result_.avg_time_us << " us" << std::endl;
+    std::cout << "Avg GetAndClear Time: " << result_.avg_get_time_us << " us" << std::endl;
+}
```
mag1c-h: This test is not a valid unit test, it lacks result checking.

flesher0813: Remove it out of ut

PR #685, opened 2026-01-27
[refactor] Rename global rank and remove broadcast function

harrisonyhq:
# Purpose

In previous code, self.global_rank is assigned to represent global tp rank, and it might cause some misunderstanding, so rename it to self.tp_rank. 

Also load only first rank is currently unavaliable, so remove it.

# Modifications 

Remove broadcast function and rename self.global_rank.

# Test

How was this patch tested?


examples/ucm_config_example.yaml
```diff
@@ -14,8 +14,6 @@ ucm_connectors:
       storage_backends: "/mnt/test"
       io_direct: false
 
-load_only_first_rank: false
```
mag1c-h, line 17: The `load_only_first_rank` parameter is referenced in many documents and test cases, please update accordingly.

harrisonyhq: Already updated 👍

PR #686, opened 2026-01-27
[feat] rerope feature for vllm0.11.0

xinSky00:
# Purpose

Adapting ReRoPE for vLLM v0.11.0

# Modifications 

1. Added git apply patch files for ReRoPE feature adaptation;
2. Included environment variable setup instructions for vLLM v0.11.0 in the documentation

# Test

image

### DIFF
--- a/docs/source/user-guide/sparse-attention/cacheblend.md
+++ b/docs/source/user-guide/sparse-attention/cacheblend.md
@@ -39,7 +39,7 @@ CacheBlend reduces TTFT by 2.2 ~ 3.3× and increases throughput by 2.8 ~ 5× und
 ### Installation
 
 Blend is part of the UCM Sparse Attention module. For installation instructions, please refer to the [UCM's top-level README](https://github.com/ModelEngine-Group/unified-cache-management). Once UCM is installed, Blend is naturally supported by running the following example python scripts.
-
+Currently, Blend only supports CUDA GPUs due to missing kernel implementations on other backends.
 ```bash
 export ENABLE_SPARSE=TRUE
 export DATA_DIR=/home/data/kv_cache
@@ -50,7 +50,7 @@ python <ucm-repo>/examples/offline_inference_blend.py
 
 ### Basic Usage
 Similar to UCM's `offline_inference_esa.py` examples. We only need to specify `ucm_sparse_method` to be `Blend` and specify meta config, as shown below.
-
+You can refer to the offline_inference_blend.py example to get started.
 ```python
 ...
 ktc = KVTransferConfig(
@@ -63,7 +63,7 @@ ktc = KVTransferConfig(
                     "ucm_connector_name": "UcmNfsStore",
                     "ucm_connector_config": {
                         "storage_backends": data_dir,
-                        "kv_block_size": 33554432,
+                        "use_direct": False,
                     },
                 }
             ],
@@ -77,7 +77,6 @@ ktc = KVTransferConfig(
                     },
                 }
             },
-            "use_layerwise": True,
         },
     )
 ...
--- a/ucm/integration/vllm/blend_connector.py
+++ b/ucm/integration/vllm/blend_connector.py
@@ -18,7 +18,6 @@ from ucm.integration.vllm.ucm_connector import (
     UCMDirectConnector,
 )
 from ucm.logger import init_logger
-from ucm.shared.metrics import ucmmonitor
 from ucm.sparse.blend.blockwise_rope import block_wise_rope_forward
 
 if TYPE_CHECKING:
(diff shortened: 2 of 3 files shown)


ucm/integration/vllm/patch/0.11.0/vllm-adapt-rerope.patch
```diff
@@ -0,0 +1,1676 @@
...
+                     attn_metadata = attn_metadata[self.layer_name]
+                 self_kv_cache = self.kv_cache[forward_context.virtual_engine]
+-                self.impl.forward(self,
+-                                  query,
+-                                  key,
+-                                  value,
+-                                  self_kv_cache,
+-                                  attn_metadata,
+-                                  output=output)
++                if envs.VLLM_USE_REROPE:
```
Infinite666, line 53: All of these places that only need to pass `query2` and `key2` but no using them, can just pass by `(query2=query2, key2=key2)`. It's unnecessary to write a branch for `if envs.VLLM_USE_REROPE`, only at the functions that actually use `query2` and `key2`, you can add a branch for `if query2 is not None and key2 is not None`.

ucm/integration/vllm/patch/0.11.0/vllm-adapt-rerope.patch
```diff
@@ -0,0 +1,1676 @@
...
++                                    query,
++                                    key,
++                                    value,
++                                    self_kv_cache,
++                                    attn_metadata,
++                                    output=output)
+             else:
+-                torch.ops.vllm.unified_attention_with_output(
+-                    query, key, value, output, self.layer_name)
++                if envs.VLLM_USE_REROPE:
```
Infinite666, line 74: The same logic as above, and so as all the below logic like `if envs.VLLM_USE_REROPE: attention(xxx, query2=query2, key2=key2) else: attention(xxx,)`, these code can be simplified.

ucm/integration/vllm/patch/0.11.0/vllm-adapt-rerope.patch
```diff
@@ -0,0 +1,1676 @@
...
++    query2: Optional[torch.Tensor] = None,
++    key2: Optional[torch.Tensor] = None,
+     output_scale: Optional[torch.Tensor] = None,
+     output_block_scale: Optional[torch.Tensor] = None,
+ ) -> None:
+diff --git a/vllm/attention/ops/triton_unified_attention_rerope.py b/vllm/attention/ops/triton_unified_attention_rerope.py
+new file mode 100644
+index 000000000..3028d2902
+--- /dev/null
++++ b/vllm/attention/ops/triton_unified_attention_rerope.py
```
Infinite666: Is there any difference between this file and `ucm\sparse\rerope\triton_unified_attention_rerope.py`? Why cannot just use that file?

ucm/integration/vllm/patch/0.11.0/vllm-adapt-rerope.patch
```diff
@@ -0,0 +1,1676 @@
...
+ 
+@@ -183,8 +186,30 @@ class Qwen2Attention(nn.Module):
+     ) -> torch.Tensor:
+         qkv, _ = self.qkv_proj(hidden_states)
+         q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size], dim=-1)
+-        q, k = self.rotary_emb(positions, q, k)
+-        attn_output = self.attn(q, k, v)
++
++        if envs.VLLM_USE_REROPE:
++            attn_metadata = get_forward_context().attn_metadata
```
Infinite666: Code between attn_metadata = get_forward_context().attn_metadata and del k2 can be extracted as a function, since you're using the exact same code for multiple times.

ucm/integration/vllm/patch/0.11.0/vllm-adapt-rerope.patch
xinSky00: This branch takes into account the independent operation of sparse attention and rerope. In sparse attention, flash_attn implementation is running and there is no setting for query2 and key2, so query2 and key2 are not compatible in impl.forward here.

ucm/integration/vllm/patch/0.11.0/vllm-adapt-rerope.patch
xinSky00: The two files are the same. The file ucm\sparse\rerope\triton_unified_attention_rerope.py is generated for the rerope call in the monkey patch form. While the file vllm/attention/ops/triton_unified_attention_rerope.py is for intrusive modification in vllm.

ucm/integration/vllm/patch/0.11.0/vllm-adapt-rerope.patch
xinSky00: OK. I will consider this suggestion.

PR #691, opened 2026-01-28
[Bugfix] Modify worker_id set to separate different worker

flesher0813:
## Purpose
Use get_world_group().rank to mark worker's connector metrics, use engine_id to mark scheduler's connector metrics.
## Modifications 
When using DP, metrics could not figure out which worker the stats belong to.
## Test
image

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -147,10 +147,18 @@ class UCMDirectConnector(KVConnectorBase_V1):
 
         self.metrics_config = self.launch_config.get("metrics_config_path", "")
         if self.metrics_config:
+            worker_id = (
+                get_world_group().rank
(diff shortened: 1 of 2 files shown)


ygwpz (dismissed):

ucm/integration/vllm/ucm_connector.py
```diff
@@ -149,7 +149,7 @@ def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
         if self.metrics_config:
             self.stats_logger = PrometheusStatsLogger(
                 vllm_config.model_config.served_model_name,
-                self.global_rank,
+                os.getpid(),
```
ygwpz: if multinode scenarios maybe duplicated

ygwpz (approved):

PR #702, opened 2026-01-29
[Opt]Support passing parameters by overriding default values through environment variables

Potterluo:
# Purpose
This Pull Request introduces dynamic environment variable substitution and automatic type inference to the configuration utility. This is specifically designed to facilitate seamless configuration overrides in CI/CD pipelines and containerized environments.

# Modifications 
Updated ConfigUtils to support the ${VAR_NAME:-default_value} syntax. This allows the application to prioritize system environment variables while falling back to YAML defaults if the variables are unset.
## Key Changes
1. Environment Variable Substitution
Regex Pattern: Implemented \$\{(\w+)(?::-([^}]*))?\} to capture variable names and optional default values.
Special Character Support: The parser is designed to handle hyphens and special symbols within default values (e.g., postgresql-primary.svc.cluster.local). It splits only at the first :- sequence.
Recursive Processing: The utility now performs a deep-traversal of the YAML dictionary/list structure to ensure all nested strings are processed.

2. Automatic Type Conversion
To maintain data integrity, the utility now automatically casts strings to their native Python types:
Boolean: Converts "true" to True and "false"to False.
Numeric: Automatically detects and converts strings to int or float.

# Test
image

### DIFF
--- a/examples/metrics/grafana.json
+++ b/examples/metrics/grafana.json
@@ -124,7 +124,7 @@
           "expr": "rate(ucm:interval_lookup_hit_rates_sum{model_name=\"$model_name\"}[$__rate_interval])\n/\nrate(ucm:interval_lookup_hit_rates_count{model_name=\"$model_name\"}[$__rate_interval])",
           "hide": false,
           "instant": false,
-          "legendFormat": "Average",
+          "legendFormat": "worker-{{worker_id}}",
           "range": true,
           "refId": "A"
         }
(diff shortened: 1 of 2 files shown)


test/common/config_utils.py
```diff
@@ -1,86 +1,195 @@
+import logging
```
Wwwzff, line 1: import logger from ucm module maybe

test/common/config_utils.py
```diff
@@ -1,86 +1,195 @@
...
 import threading
-from typing import Any, Dict
+from typing import Any, Dict, Optional, Union
 
 import yaml
 
+logger = logging.getLogger(__name__)
+
+
+def _parse_string_type(value: Any) -> Any:
```
Wwwzff, line 12: try to use ast.literal_eval instead of building another wheel

PR #717, opened 2026-01-31
[bugfix] Adaptive modification of llmperf

Menglths:
## Purpose
Enhance the LLM performance test framework by introducing configurable inference behavior and improving flexibility across different inference engines.
This update enables engine-specific execution logic, refines token input handling, and aligns configuration management with existing utilities, improving usability, extensibility, and test consistency.
## Modifications 
- Added new configuration options llm_type and enable_clear_hbm to support custom inference engine selection and optional HBM clearing.
- Introduced sglang inference logic with two prefill runs followed by a single performance test run.
- Introduced mindie inference logic by disabling HBM clearing and enabling multi-round hit-rate testing.
- Updated configuration loading to use the existing config_utils tool.
- Added configurable console output for improved runtime visibility.
- Updated token input rules to support the new inference workflows.
## Test
All test have passed.


test/common/llmperf/utils/token_benchmark.py
```diff
@@ -22,6 +24,28 @@
...
+    using token-id sampling instead of text concatenation.
+    """
+    num_tokens = sample_random_positive_int(mean_tokens, stddev_tokens)
+
+    text = tokenizer_tool.get_some_tokens(
+        num_tokens=num_tokens,
+        seed=seed,
+    )
+
+    actual_tokens = tokenizer_tool.count_tokens(text, include_special=False)
```
Wwwzff: take special tokens into count to get a more accurate tokennum

test/common/llmperf/utils/token_benchmark.py
```diff
@@ -9,6 +10,7 @@
 from typing import Any, Dict, List, Optional, Tuple
 
 import pandas as pd
+from common.llm_connection.token_counter import HuggingFaceTokenizer
```
Wwwzff, line 13: may be a better way to use AutoTokenizer imported from Transformers directly

PR #728, opened 2026-02-03
[Fix] Monkey-patch load failure handling for CUDA and Ascend

NaganooMei:
## Purpose
Fix load failure issues in vLLM integration on both CUDA and Ascend by applying a monkey-patch workaround.
This prevents runtime failures during model loading under specific environments.

## Modifications
- Introduce monkey-patch logic for load failure handling
- Apply unified handling for CUDA and Ascend backends
- Keep changes isolated to integration layer without affecting core logic

## Test
image
image


qyh111 (dismissed):

ucm/integration/vllm/patch/apply_patch.py
```diff
@@ -114,7 +116,7 @@ def apply_all_patches() -> None:
         _patches_applied = True
         logger.info(f"All vLLM patches applied successfully for version {version}")
     except Exception as e:
-        logger.error(f"Failed to apply vLLM patches: {e}")
+        logger.error(f"Failed to apply vLLM patches: {e}", exc_info=True)
```
mag1c-h: Log compatibility is questionable.

ygwpz (approved):

PR #744, opened 2026-02-11
[Opt]Split the thread pool into separate load and dump pools to prevent them from interfering with each other.

qyh111:
## Purpose
Posix tasks (including load and dump) share the same thread pool. So they may interfere with each other.Occasionally, an excessive number of dump tasks prevents load tasks from being executed.
## Modifications 
Split the thread pool into separate load and dump pools to prevent them from interfering with each other.
## Test
Qwen32B tp4 can be pulled up and receive request successfully

### DIFF
--- a/ucm/store/posix/cc/trans_queue.cc
+++ b/ucm/store/posix/cc/trans_queue.cc
@@ -35,9 +35,15 @@ Status TransQueue::Setup(const Config& config, TaskIdSet* failureSet, const Spac
     shardSize_ = config.shardSize;
     nShardPerBlock_ = config.blockSize / config.shardSize;
     ioDirect_ = config.ioDirect;
-    auto success = pool_.SetNWorker(config.dataTransConcurrency)
-                       .SetWorkerFn([this](auto& ios, auto&) { Worker(ios); })
+    auto success = loadPool_.SetNWorker(config.dataTransConcurrency)
+                       .SetWorkerFn([this](auto& ios, auto&) { LoadWorker(ios); })
                        .Run();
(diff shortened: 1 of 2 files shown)


ucm/store/posix/cc/trans_queue.cc
```diff
@@ -52,27 +58,41 @@ void TransQueue::Push(TaskPtr task, WaiterPtr waiter)
         ios.emplace_back<IoUnit>({task->id, task->type, std::move(shard), waiter});
```
mag1c-h: Since the transmission task has been added to an explicit queue, it is no longer necessary to retain the task type field in each shard.

examples/deployments/scripts/vllm/run_vllm_dp.sh
```diff
@@ -36,14 +36,15 @@ start_server() {
...
     if [[ "$ucm_enable" == "true" ]]; then
         [[ -z "$ucm_config_yaml_path" ]] && {
             echo "ERROR: ucm_config_yaml_path not set but ucm_enable=true" >&2
             exit 1
         }
-        LOG_FILE="${vllm_log_path}/vllm_ucm.log"
+        LOG_FILE="${vllm_log_path}/vllm_ucm_${LOG_TIMESTAMP}.log"
     else
-        LOG_FILE="${vllm_log_path}/vllm.log"
+        LOG_FILE="${vllm_log_path}/vllm_${LOG_TIMESTAMP}.log"
```
mag1c-h: A PR should not include modifications unrelated to the topic.

PR #750, opened 2026-02-12
[Feat] adapt dcp&pcp

flesher0813:
## Purpose
This PR provides adaptation for the PCP (Prefill Context Parallel) / DCP (Decode Context Parallel) features in the vllm Ascend implementation. For a detailed introduction to these features, please refer to https://docs.vllm.ai/projects/ascend/en/latest/user_guide/feature_guide/context_parallel.html.

When the CP (Context Parallel) feature is enabled, each device is allocated a corresponding current_rank, calculated as: **current_rank = dcp_world_size * pcp_rank + dcp_rank**. The tokens of a request are interleaved and distributed across each card following the rule of **[current_rank :: cp_kv_cache_interleave_size * cp_world_size]**, cp_world_size calculated as: **cp_world_size = dcp_world_size * pcp_world_size**. In scenarios where KV pooling and PD disaggregation are used, cp_kv_cache_interleave_size is equal to block_size (cp_kv_cache_interleave_size == block_size).

For UCM, this means that blocks are interleaved across each card in the form of [current_rank :: cp_world_size].

## Modifications 
Added a new connector class **UCMCPConnector** that inherits from **UCMLayerwiseConnector**.
- ucm/integration/sglang/ucm_connector.py

## Notice

- For MLA scenarios, the number of blocks saved with CP enabled is the same as the number of identical TP; for GQA scenarios, the number of blocks saved with CP enabled is the same as tp // dcp_size.
- After CP is enabled, the number of saved blocks satisfies: blocks // cp_world_size * cp_world_size.

## Test
Tested with offline_inference.py and online test using deepseek-v2-lite and Qwen3-30B-A3B models base on vLLM-Ascend
 v0.13.0

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -749,13 +749,12 @@ class UCMLayerWiseConnector(UCMDirectConnector):
 class UCMCPConnector(UCMLayerWiseConnector):
     def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
         super().__init__(vllm_config, role)
-        # [k_task, k_task, ...] for mla
-        # [k_task, v_task/rope_task, k_task, vtask/rope_task, ...] for dsa or gqa
         self.use_layerwise = (
             self._vllm_config.kv_transfer_config.kv_connector_extra_config.get(
                 "use_layerwise", False
             )
         )
+
         try:
             from vllm.distributed import get_dcp_group, get_pcp_group
         except ImportError as e:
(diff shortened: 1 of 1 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -746,6 +746,199 @@ def wait_for_save(self) -> None:
         self.is_save = False
 
 
+class UCMCPConnector(UCMLayerWiseConnector):
+    def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
+        super().__init__(vllm_config, role)
+        self.use_layerwise = (
```
ygwpz, line 752: Do we really need to declare a use_layerwise variable again in the subclass? The parent class already has one.

flesher0813: Really need, because the UCMLayerWiseConnector directly sets self.use_layerwise = True instead of reading from the config file.

PR #764, opened 2026-02-27
feat(test): Support result saving functionality for multiple storage

Potterluo:
# PR: feat(test): Support result saving functionality for multiple storage backends

## Description
This PR introduces a robust, multi-backend storage system for test results. It replaces the legacy database utility functions with a unified interface that supports **Local File**, **MongoDB**, and **PostgreSQL** backends. This enhancement allows users to configure and switch between different storage solutions seamlessly based on their infrastructure needs.

### New Features
- **Multi-Backend Support**: Implemented concrete storage adapters for:
  - `localFile`: Saves results to the local filesystem (JSON/CSV).
  - `mongodb`: Stores results in MongoDB collections.
  - `postgresql`: Stores results in PostgreSQL tables.
- **Automatic Type Inference**: Added logic to automatically infer and convert data types during the saving process, ensuring schema consistency across different backends.
- **Resilience Mechanisms**: Implemented connection management pools and automatic retry mechanisms specifically for MongoDB and PostgreSQL to handle transient network issues.

### Refactoring
- **Unified Interface**: Updated `capture_utils` to utilize the new generic storage backend interface, decoupling business logic from specific storage implementations.
- **Legacy Cleanup**: Removed obsolete database utility functions and consolidated them into the new general result-saving mechanism.
- **Configuration Overhaul**: Modified the configuration file structure to support dynamic multi-storage backend definitions (e.g., selecting the active backend and providing specific connection strings).

## Configuration Example
The configuration file now supports a `storage` section:

```yaml
results:
  - localFile: # Save the storage results to a local file. The default formats are jsonl and csv.
      path: "./results"
  - postgresql:
      host: "localhost"
      port: 5432
      dbname: "ucm_test"
      user: "postgres"
      password: "123456"
      retry: 3
  - mongodb:
      host: "127.0.0.1"
      port: 27017
      dbname: "myapp"
      user: "root"
      password: "123456"
      authSource: "admin"
      retry: 3
```

## Testing
image

### DIFF
--- a/docs/source/developer-guide/deepdive_ucm.md
+++ b/docs/source/developer-guide/deepdive_ucm.md
@@ -1,0 +1,31 @@
+# Deep Dive into UCM
+
+To better understand the **architecture and design** of **Unified Cache Management (UCM)**, we recommend exploring the project on **DeepWiki**, an AI-powered tool for technical documentation.
+
+## 📖 DeepWiki Documentation
+
+**Explore UCM on DeepWiki:**
+
+👉 [https://deepwiki.com/ModelEngine-Group/unified-cache-management](https://deepwiki.com/ModelEngine-Group/unified-cache-management)
+
+DeepWiki provides a structured overview of the project, including:
+
+- The overall architecture of UCM
+- Responsibilities and interactions of core modules
+- Key data structures and algorithms
+- Source code explanations for important components
+- The project’s directory structure and design principles
+
+Instead of duplicating all technical details in this documentation, DeepWiki serves as the central place for exploring UCM’s architecture and source code. It provides a developer-friendly view that connects architecture, modules, and implementation.
+
+If you want to:
+
+- Understand the design philosophy behind UCM
+- Explore how major components are implemented
+- Navigate the project source code more efficiently
+
+We recommend starting with the DeepWiki page above.
+
+## 🤝 Contributing
+
+Interested in UCM? After exploring the documentation, you're welcome to contribute by opening issues or submitting pull requests.
--- a/docs/source/index.md
+++ b/docs/source/index.md
@@ -66,6 +66,7 @@ user-guide/rerope/rerope
 :caption: Developer Guide
 :maxdepth: 1
 developer-guide/contribute
+developer-guide/deepdive_ucm
 developer-guide/add_metrics
 developer-guide/extending_store
 :::
--- a/docs/source/user-guide/metrics/metrics.md
+++ b/docs/source/user-guide/metrics/metrics.md
@@ -183,6 +183,3 @@ histograms:
     buckets: [1, 5, 10, 20, 50, 100, 200, 500, 1000]
   # ... other metric configurations
 ```
-
----
-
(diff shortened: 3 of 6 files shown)


test/common/db_utils.py
```diff
@@ -1,113 +1,109 @@
...
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
-from peewee import AutoField, DateTimeField, Model, TextField
-from playhouse.reflection import Introspector
-
-# Lazy imports for database components
+# 延迟导入
 peewee = None
```
dante159753, line 9: try not directly use global var, it's hard to maintain, use a consistent entrance for these lazy modules, like a function:
``` python
def getDbDeps():
   # import modules into global var or a global object
   # check before import, only import once
   return depsObj

# use it
getDbDeps().peewee
```
this way you don't need _ensure_peewee_imported

test/.gitignore
```diff
@@ -11,4 +11,5 @@ $null
 .*
 *.log
 start.bat
+*demo.py
```
dante159753, line 14: why we need this

test/common/capture_results/mongodb.py
```diff
@@ -0,0 +1,235 @@
...
+# Database Configuration
+# ============================================================================
+logger = logging.getLogger(__name__)
+
+
+# Timezone constant: UTC+8
+UTC_8 = timezone(timedelta(hours=8))
+
+# Module-level state
+_test_build_id: Optional[str] = None
```
dante159753: don't use glocal vars, maintain them in config obj, or create another env obj conform to singleton design pattern, when you need to change the logic, you don't need to change all files

test/common/capture_results/mongodb.py
```diff
@@ -0,0 +1,235 @@
...
+_db_config: Optional[Dict[str, Any]] = None
```
dante159753, line 14: same as above, you can reuse config obj

test/.gitignore
Potterluo: > why we need this

because I have some test files locally that I don't want to be submitted, that's all. I think it's general-purpose.

test/common/db_utils.py
Potterluo: > try not directly use global var, it's hard to maintain, use a consistent entrance for these lazy modules, like a function:
> 
> ```python
> def getDbDeps():
>    # import modules into global var or a global object
>    # check before import, only import once
>    return depsObj
> 
> # use it
> getDbDeps().peewee
> ```
> 
> this way you don't need _ensure_peewee_imported

a great suggestion, I will try to use him

test/common/capture_results/mongodb.py
Potterluo: > don't use glocal vars, maintain them in config obj, or create another env obj conform to singleton design pattern, when you need to change the logic, you don't need to change all files

Okay, I violated the design rules. This is indeed far too redundant.I will modify the relevant content and replace them with env obj.

test/common/capture_results/mongodb.py
Potterluo: > same as above, you can reuse config obj

you right，i think they should be in capture_utils.py,

PR #768, opened 2026-02-28
[opt]Add event to sync

qyh111:
## Purpose
This PR adds a config switch `enable_event_sync` to control whether dump tasks wait on a prerequisite compute event.
## Modifications 
Add `enable_event_sync`plumbing in `ucm_connector.py`.
Pass event handle to `dump_data` only when sync is enabled; otherwise pass 0.
Add `CopyStream::WaitEvent(void* event)` to apply wait on all internal copy streams.
Update `DumpQueue` to issue prerequisite wait once via `CopyStream`, ensuring all streams are covered.
Make `Trans::Stream::WaitEvent` pure virtual and implement it in SimuStream (no-op).
## Test
image

image
image

### DIFF
--- a/ucm/shared/trans/stream.h
+++ b/ucm/shared/trans/stream.h
@@ -50,9 +50,6 @@ public:
 
     virtual Status AppendCallback(std::function<void(bool)> cb) = 0;
     virtual Status Synchronized() = 0;
-
-    /** Wait for compute-stream event before D2H. Event ptr is platform-specific
-     * (cudaEvent_t or aclrtEvent). */
     virtual Status WaitEvent(void* event) = 0;
 };
 


ucm/store/cache/cc/dump_queue.cc
```diff
@@ -96,8 +96,17 @@ Status DumpQueue::DumpOneTask(CopyStream& stream, TaskPtr task)
         auto handle = buffer_->Get(shard.owner, shard.index);
         if (!handle.Owner()) { continue; }
         if (!handle.Ready()) {
+            auto cacheStream = stream.NextStream();
+            if (task->desc.compute_event_handle != 0) {
+                auto s = cacheStream->WaitEvent(
```
mag1c-h: A task has only one event, so it is not necessary to execute a wait in each shard.

ucm/store/detail/type/types.h
```diff
@@ -62,6 +62,8 @@ struct Shard {
 struct TaskDesc : std::vector<Shard> {
     using vector::vector; /* Inherit all ctors */
     std::string brief;    /* Description of Task */
+    /** Optional: compute-stream event handle for dump. Cache stream waits before D2H. */
+    uintptr_t compute_event_handle{0};
```
mag1c-h: Naming style does not meet requirements

mag1c-h: This parameter represents the handle to the prerequisite events that the task execution depends on. Using `prerequisiteHandle` might be better.

ucm/integration/vllm/device.py
```diff
@@ -0,0 +1,107 @@
...
+    @abstractmethod
+    def destroy_event_handles(self):
+        pass
+
+class CudaDevice(Device):
+    def __init__(self, device_id: int = 0):
+        super().__init__(device_id)
+
+    def init_device(self):
+        torch.cuda.set_device(self.device_id)
```
mag1c-h: The context has already been set up where this class is called, so it does not need to be initialized again.

ucm/integration/vllm/device.py
```diff
@@ -0,0 +1,107 @@
...
+"""
+from typing import Optional
+from abc import ABC, abstractmethod
+import torch
+from vllm.platforms import current_platform
+from ucm.logger import init_logger
+
+logger = init_logger(__name__)
+
+class Device(ABC):
```
mag1c-h, line 21: The code for runtime difference adaptation already exists, it's best to put the new parts together with them.

ucm/shared/trans/stream.h
```diff
@@ -50,8 +50,16 @@ class Stream {
 
     virtual Status AppendCallback(std::function<void(bool)> cb) = 0;
     virtual Status Synchronized() = 0;
+
+    /** Wait for compute-stream event before D2H. Event ptr is platform-specific
+     * (cudaEvent_t or aclrtEvent). No-op when event is nullptr. */
+    virtual Status WaitEvent(void* event)
+    {
```
mag1c-h: It might be better to define it as a pure virtual function, emphasizing that all inheritors must implement it.

qyh111: Fix it

ucm/integration/vllm/device.py
qyh111: Now all runtime difference adaptation have been moved to device.py

ucm/integration/vllm/device.py
qyh111: fix it

ucm/store/detail/type/types.h
qyh111: fix it

ucm/store/cache/cc/dump_queue.cc
qyh111: fix it

ucm/shared/trans/stream.h
```diff
@@ -50,8 +50,12 @@ class Stream {
 
     virtual Status AppendCallback(std::function<void(bool)> cb) = 0;
     virtual Status Synchronized() = 0;
+
+    /** Wait for compute-stream event before D2H. Event ptr is platform-specific
+     * (cudaEvent_t or aclrtEvent). */
```
mag1c-h: The transfer library only provides basic capabilities, so it's inappropriate to put business-related comments here.

qyh111: OK. I have removes it.

PR #776, opened 2026-03-03
[opt] supports loading non-built-in connectors

mag1c-h:
supports loading non-built-in connectors

 ```yaml
 ucm_connectors:
   - ucm_connector_name: "CustomStore"
     ucm_connector_module_path: "custom.store.connector"
     ucm_connector_config:
       # ... custom config
 ```

### DIFF
--- a/ucm/integration/sglang/ucm_connector.py
+++ b/ucm/integration/sglang/ucm_connector.py
@@ -69,6 +69,7 @@ def _load_extra_config_from_yaml_env() -> Optional[Dict[str, Any]]:
 
 @dataclass
 class UnifiedCacheStoreConfig:
+    module_path: str
     name: str
     config: Dict[str, Any]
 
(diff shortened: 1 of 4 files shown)


ucm/integration/sglang/ucm_connector.py
```diff
@@ -98,6 +99,7 @@ def load_from_config(
 
         ucm_cfg = kvc.get("ucm_connector_config")
         name = kvc.get("ucm_connector_name")
+        module_path = kvc.get("ucm_connector_module_path")
```
pyxyzc, line 102: Is the ucm_connector_module_path parameter required or optional? If it is required, maybe we need to add a value check.

mag1c-h: The parameter is optional. If it is None, it means that the built-in connector is used.

PR #777, opened 2026-03-04
[Feature] Garbage Collection

UESTC-AHao:
## Purpose
Support Garbage Collection：

Hotness Updates: Each Scheduler records blocks hit by a Lookup and asynchronously updates the last time of the corresponding files.

GC Process Enablement: Executed only on Schedulers. In Data Parallel (DP) scenarios, it runs exclusively on the DP0 Scheduler, controlled by the configuration item posix_gc_enable.

GC Initial Capacity: The total storage capacity is derived from the user configuration posix_capacity_gb, and the maximum number of storable files is calculated based on individual file sizes.

GC Trigger Conditions: Triggered if either condition is met: a) Timer expiration; b) The file count still exceeds the threshold after the previous GC round.

GC Execution Detection: Randomly samples 10% of subdirectories and uses multi-threading to count files in parallel, checking whether the average file count per subdirectory exceeds the threshold.

GC Execution Method: Processes all subdirectories in parallel using multiple threads, reclaiming the 10% oldest (least recently accessed) files within each subdirectory.
## Modifications 
ucm_connectors:
  - ucm_connector_name: "UcmPipelineStore"
    ucm_connector_config:
      store_pipeline: "Cache|Posix"
      storage_backends: "./data"
      
      [REQUIRED] Total storage capacity in GB (must set when GC enabled)
      posix_capacity_gb: 
      
      [CONFIGURABLE] Enable garbage collection (default: false)
      posix_gc_enable: true
      
      [CONFIGURABLE] GC trigger threshold ratio (default: 0.7)
      posix_gc_trigger_threshold_ratio: 0.7
      
      [CONFIGURABLE] Percentage of files to recycle per GC run (default: 0.1)
      posix_gc_recycle_percent: 0.1
      
      [CONFIGURABLE] Number of GC worker threads (default: 16)
      posix_gc_concurrency: 16

      [CONFIGURABLE] Timeof GC check interval (default: 30)
      posix_gc_check_interval_sec: 30

      [CONFIGURABLE] The maximum number of files that can be processed per shard (default: 1000)
      posix_gc_max_recycle_count_per_shard: 1000

      [CONFIGURABLE] Shard sample ratio (default: 0.1)
      posix_gc_shard_sample_ratio: 0.1

## Test
423459bc-48d5-42e7-a5b1-8c90baa081a2


ucm/store/posix/cc/hotness_tracker.cc
```diff
@@ -0,0 +1,108 @@
...
+                std::this_thread::yield();
+            } else {
+                std::this_thread::sleep_for(std::chrono::microseconds(100));
+                spinCount = 0;
+            }
+            continue;
+        }
+        spinCount = 0;
+        while (!consumeQueue_.empty()) {
+            utime(consumeQueue_.front().c_str(), nullptr);
```
sumingZero: I noticed the utime system call is used without checking its return value. Should we consider adding error handling here

ucm/store/posix/cc/space_layout.h
```diff
@@ -39,9 +40,12 @@ class SpaceLayout {
     Status Setup(const Config& config);
     std::string DataFilePath(const Detail::BlockId& blockId, bool activated) const;
     Status CommitFile(const Detail::BlockId& blockId, bool success) const;
+    std::vector<std::string> RelativeRoots() const;
+    const std::vector<std::string>& GetStorageBackends() const { return storageBackends_; }
+    std::string PropertyFilePath() const;
+    std::string CapacityDir() const;
```
qyh111: It seems that CapacityDir() and PropertyFilePath() defined but not used

ucm/store/posix/cc/shard_gc.cc
```diff
@@ -0,0 +1,311 @@
...
+
+    for (size_t t = 0; t < threadsCount; ++t) {
+        threads.emplace_back([&, t, shardsPerThread]() {
+            size_t start = t * shardsPerThread;
+            size_t end = std::min(start + shardsPerThread, sampleCount);
+            size_t localCount = 0;
+
+            for (size_t i = start; i < end; ++i) {
+                size_t shardIdx = shardIndices[i];
+                for (const auto& backend : backends_) {
```
qyh111: Why scan multiple backends here? All backends points to the same remote file system

ucm/store/posix/cc/shard_gc.cc
```diff
@@ -0,0 +1,311 @@
...
+        deletedCount = std::make_shared<std::atomic<size_t>>(0);
+        waiter = std::make_shared<Latch>();
+    } catch (const std::exception& e) {
+        running_.store(false);
+        return Status::OutOfMemory();
+    }
+
+    waiter->Set(totalTasks);
+
+    for (const auto& backend : backends_) {
```
qyh111: Why traverse all backends here? I think handle the first backend is enough

ucm/store/posix/cc/shard_gc.cc
```diff
@@ -0,0 +1,311 @@
...
+
+size_t ShardGarbageCollector::ScanAndCollectOldestFiles(const std::string& shardPath,
+                                                        double recyclePercent,
+                                                        std::vector<FileInfo>& filesToDelete)
+{
+    DIR* dir = opendir(shardPath.c_str());
+    if (!dir) { return 0; }
+
+    std::vector<FileInfo> allFiles;
+    allFiles.reserve(1024);
```
qyh111: Could you clarify what 1024 stands for here?

ucm/store/posix/cc/shard_gc.cc
```diff
@@ -0,0 +1,311 @@
...
+    if (asyncWorker_.joinable()) { asyncWorker_.join(); }
+    asyncWorker_ = std::thread([this] {
+        size_t totalDeleted = 0;
+        while (true) {
+            auto result = Execute();
+            if (result) {
+                size_t deleted = result.Value();
+                totalDeleted += deleted;
+                UC_DEBUG("GC deleted {} files", deleted);
+                UC_DEBUG("GC total deleted {} files", totalDeleted);
```
qyh111: These two debug log can probably be combined into one log

ucm/store/cache/cc/buffer_manager.h
```diff
@@ -71,16 +71,21 @@ class BufferManager {
...
         missIdx.reserve(num);
+        hitBlk.reserve(num);
         StopWatch sw;
         for (size_t i = 0; i < num; ++i) {
             uint8_t hit = buffer_->Exist(blocks[i], 0);
             results.push_back(hit);
-            if (hit) { continue; }
+            if (hit) {
+                hitBlk.push_back(blocks[i]);
+                continue;
```
qyh111: It seems there is no further processing after obtaining hitBlk. Is this intended?

ucm/store/posix/cc/hotness_tracker.cc
```diff
@@ -0,0 +1,108 @@
...
+    gc_ = gc;
+    maxFileCount_ = maxFileCount;
+    thresholdRatio_ = thresholdRatio;
+}
+
+void HotnessTracker::Touch(const Detail::BlockId& blockId)
+{
+    auto filePath = layout_->DataFilePath(blockId, false);
+    std::lock_guard<std::mutex> lock(queueMtx_);
+    produceQueue_.push_back(std::move(filePath));
```
qyh111: In multi-TP scenarios, the scheduler only looks up files from rank 0. Consequently, files from other ranks are at risk of being garbage collected.

ucm/store/posix/cc/space_layout.h
UESTC-AHao: I'll delete them; they are the old methods.

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: I see. Since multiple backends correspond to the same filesystem (a shared directory), I only need to operate on the first backend.

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: Yes，handle the first backend is enough

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: 1024 was a hardcoded empirical value; I have changed it to an actual value calculated based on the maximum capacity, which is more reasonable.

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: I will optimize the DEBUG logging.

ucm/store/cache/cc/buffer_manager.h
UESTC-AHao: I will fix it.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -287,6 +292,20 @@ def _create_store(
             config["shard_size"] = kv_cache_layout.shard_size * self.blocks_per_chunk
             config["block_size"] = kv_cache_layout.block_size * self.blocks_per_chunk
             config["local_rank_size"] = self.tp_size if self.is_mla else 1
+        else:
+            config_base = self.block_size * self.element_size * self.head_size
+            config["block_size"] = (
+                config_base
+                * self.num_layers
+                * (1 if self.is_mla else self.num_head * 2)
+                * self.blocks_per_chunk
```
mag1c-h, line 337: Will the different initialization methods of Scheduler and Worker for BlockSize lead to inconsistencies in the final results?

ucm/store/posix/cc/global_config.h
```diff
@@ -31,6 +31,7 @@ namespace UC::PosixStore {
 
 struct Config {
     std::vector<std::string> storageBackends{};
+    std::string uniqueId{};
```
mag1c-h: unused config item.

ucm/store/posix/cc/global_config.h
```diff
@@ -43,6 +44,12 @@ struct Config {
     size_t commitConcurrency{4};
     size_t timeoutMs{30000};
     size_t dataDirShardBytes{3};
+    bool posixGcEnable{false};
+    double posixGcRecyclePercent{0.1};
+    size_t posixGcConcurrency{16};
+    size_t posixGcCheckInterval{30};
```
mag1c-h: It would be better to include the time unit in the variable name.

ucm/store/test/e2e/cache_on_posix_test.py
```diff
@@ -139,9 +139,15 @@ def main():
     config["io_direct"] = True
     config["posix_data_trans_concurrency"] = 32
     config["posix_lookup_concurrency"] = 32
+    test_batch_number = 512
+    config["posix_gc_enable"] = True
+    config["posix_capacity_gb"] = 100
+    config["posix_gc_trigger_threshold_ratio"] = 0.7
+    config["posix_gc_recycle_percent"] = 0.1
+    config["posix_gc_concurrency"] = 16
+    config["posix_gc_check_interval"] = 30
```
mag1c-h: Write test cases specifically for GC, do not modify the original test cases.

ucm/store/test/e2e/posixstore_embed.py
```diff
@@ -51,6 +51,15 @@ def setup(
...
     config["io_direct"] = io_direct
     config["device_id"] = 0 if worker else -1
+
+    # GC配置
+    config["posix_gc_enable"] = True
+    config["posix_capacity_gb"] = 100
+    config["posix_gc_trigger_threshold_ratio"] = 0.7
+    config["posix_gc_recycle_percent"] = 0.1
+    config["posix_gc_concurrency"] = 16
+    config["posix_gc_check_interval"] = 30
```
mag1c-h: Write test cases specifically for GC, do not modify the original test cases.

ucm/store/posix/cc/hotness_tracker.cc
```diff
@@ -0,0 +1,108 @@
...
+        std::lock_guard<std::mutex> lock(gcMtx_);
+        gcCv_.notify_all();
+    }
+    if (utimeWorker_.joinable()) { utimeWorker_.join(); }
+    if (gcWorker_.joinable()) { gcWorker_.join(); }
+}
+
+Status HotnessTracker::Setup(const SpaceLayout* layout, size_t gcCheckIntervalSeconds)
+{
+    if (!layout) { return Status::InvalidParam("layout is null"); }
```
mag1c-h: The interface calls within the module do not require redundant parameter checks.

ucm/store/posix/cc/hotness_tracker.h
```diff
@@ -0,0 +1,75 @@
...
+    HotnessTracker& operator=(const HotnessTracker&) = delete;
+    ~HotnessTracker();
+
+    Status Setup(const SpaceLayout* layout, size_t gcCheckIntervalSeconds);
+    void SetGCTrigger(ShardGarbageCollector* gc, size_t maxFileCount, double thresholdRatio);
+    void Touch(const Detail::BlockId& blockId);
+
+private:
+    void UtimeWorkerLoop();
+    void GCCheckLoop();
```
mag1c-h: The logic for capacity checks and GC execution should not be implemented in the HotnessTracker module.

ucm/store/posix/cc/hotness_tracker.cc
```diff
@@ -0,0 +1,108 @@
...
+    layout_ = layout;
+    gcCheckIntervalSeconds_ = gcCheckIntervalSeconds;
+    stop_.store(false);
```
mag1c-h: The default value is always false, so there is no need to set it explicitly.

ucm/store/posix/cc/hotness_tracker.h
```diff
@@ -0,0 +1,75 @@
...
+#include <deque>
+#include <mutex>
+#include <string>
+#include <thread>
+#include "space_layout.h"
+#include "type/types.h"
+
+namespace UC::PosixStore {
+
+class ShardGarbageCollector;
```
mag1c-h: Why not just include the header file?

ucm/store/posix/cc/posix_store.cc
```diff
@@ -97,6 +156,21 @@ class PosixStore : public StoreV1 {
...
+    {
+        if (!gcEnable_) { return; }
+        gcMgr_.Trigger();
+    }
+    Expected<size_t> ExecuteGC()
+    {
+        if (!gcEnable_) { return Status::Error("GC is not enabled"); }
+        return gcMgr_.Execute();
+    }
+    bool IsGCRunning() const
```
mag1c-h: These three functions don't have a caller in sight, so what's the point of defining them?

ucm/store/posix/cc/posix_store.cc
```diff
@@ -32,7 +34,10 @@ namespace UC::PosixStore {
 class PosixStore : public StoreV1 {
     SpaceManager spaceMgr_;
     TransManager transMgr_;
+    ShardGarbageCollector gcMgr_;
+    HotnessTracker hotnessTracker_;
```
mag1c-h: Both heat updates and garbage collection should belong to the SpaceManager submodule, not be listed alongside it.

ucm/store/posix/cc/shard_gc.cc
```diff
@@ -0,0 +1,297 @@
...
+    maxFileCount_ = maxFileCount;
+    thresholdRatio_ = thresholdRatio;
+}
+
+void ShardGarbageCollector::Trigger()
+{
+    std::lock_guard<std::mutex> lock(mtx_);
+    if (running_.load()) { return; }
+    if (asyncWorker_.joinable()) { asyncWorker_.join(); }
+    asyncWorker_ = std::thread([this] {
```
mag1c-h: There is a threadpool running for GC, so why we need an asyncWorker_?

ucm/store/posix/cc/shard_gc.cc
```diff
@@ -0,0 +1,297 @@
...
+{
+    maxFileCount_ = maxFileCount;
+    thresholdRatio_ = thresholdRatio;
+}
+
+void ShardGarbageCollector::Trigger()
+{
+    std::lock_guard<std::mutex> lock(mtx_);
+    if (running_.load()) { return; }
+    if (asyncWorker_.joinable()) { asyncWorker_.join(); }
```
mag1c-h: This process will continue to block until the GC is complete. Is this as expected?

ucm/store/posix/cc/hotness_tracker.cc
```diff
@@ -0,0 +1,108 @@
...
+void HotnessTracker::UtimeWorkerLoop()
+{
+    constexpr size_t kSpinLimit = 16;
+    size_t spinCount = 0;
+    while (!stop_.load()) {
+        {
+            std::lock_guard<std::mutex> lock(queueMtx_);
+            consumeQueue_.swap(produceQueue_);
+        }
+        if (consumeQueue_.empty()) {
```
mag1c-h: Since `consumeQueue_` is only used within the current function, wouldn't it be better to use a local variable instead of a class member variable?

ucm/store/posix/cc/hotness_tracker.cc
mag1c-h: Storing only the BlockId instead of the full file path can reduce memory usage.

ucm/store/posix/cc/posix_store.cc
```diff
@@ -43,14 +48,68 @@ class PosixStore : public StoreV1 {
...
+                            config.posixCapacityGb, thresholdFilesPerShard);
+                }
+                gcMgr_.SetGCThreshold(maxFileCount, config.posixGcTriggerThresholdRatio);
+                hotnessTracker_.SetGCTrigger(&gcMgr_, maxFileCount,
+                                             config.posixGcTriggerThresholdRatio);
+                if (gcMgr_.ShouldTrigger(maxFileCount, config.posixGcTriggerThresholdRatio)) {
+                    gcMgr_.Trigger();
+                }
+            }
+        }
```
mag1c-h: The effective code length of a function should ideally not exceed 50 lines, otherwise the code readability will drop sharply.

ucm/store/posix/cc/posix_store.cc
```diff
@@ -43,14 +48,68 @@ class PosixStore : public StoreV1 {
...
+                        size_t minCapacityGb =
+                            (minCapacityBytes + 1024ULL * 1024ULL * 1024ULL - 1) /
+                            (1024ULL * 1024ULL * 1024ULL);
+                        return Status::InvalidParam(
+                            "posix_capacity_gb({}) is too small, GC cannot recycle any "
+                            "files. Minimum recommended: {}GB",
+                            config.posixCapacityGb, minCapacityGb);
+                    }
+                    UC_INFO("GC enabled: capacityGb={}, thresholdFilesPerShard={}",
+                            config.posixCapacityGb, thresholdFilesPerShard);
```
mag1c-h: Parameter checking and adaptation logic for different conditions should not be in PosixStore::Setup.

ucm/store/posix/cc/shard_gc.cc
```diff
@@ -0,0 +1,297 @@
...
+    DIR* dir = opendir(shardPath.c_str());
+    if (!dir) { return 0; }
+
+    std::vector<FileInfo> allFiles;
+    allFiles.reserve(maxFileCount_ / layout_->RelativeRoots().size());
+
+    struct dirent* entry;
+    while ((entry = readdir(dir)) != nullptr) {
+        if (entry->d_name[0] == '.') { continue; }
+        if (strstr(entry->d_name, ".tmp") != nullptr) { continue; }
```
mag1c-h: All objects that begin with a period (.) should be skipped.

ucm/store/posix/cc/shard_gc.cc
```diff
@@ -0,0 +1,297 @@
...
+    for (const auto& file : allFiles) { heap->Push(file); }
+
+    filesToDelete.reserve(recycleNum);
+    while (!heap->Empty()) {
+        filesToDelete.push_back(heap->Top());
+        heap->Pop();
+    }
+
+    return totalFiles;
+}
```
mag1c-h: The current implementation has a problem: if the directory contains a large number of files, `allFiles` consumes a significant amount of memory.

We should define the Heap with a maximum number of files to be reclaimed, then loop through all files and add them to the Heap, while simultaneously counting the total number of files. Finally, based on the actual total number of files, determine the number of files to be reclaimed, pop the actual number from the Heap, and perform the reclamation. If the total number in the Heap has not yet reached the threshold, then trigger the next iteration from the beginning.

ucm/store/posix/cc/shard_gc.h
```diff
@@ -0,0 +1,101 @@
...
+struct MtimeComparator {
+    bool operator()(const FileInfo& lhs, const FileInfo& rhs) const
+    {
+        return lhs.mtime > rhs.mtime;
+    }
+};
+
+Detail::BlockId HexToBlockId(const char* hexStr);
+
+std::string BlockIdToHex(const Detail::BlockId& blockId);
```
mag1c-h: These two function declarations should not be placed in the header file.

ucm/store/posix/cc/space_layout.h
```diff
@@ -39,9 +40,10 @@ class SpaceLayout {
     Status Setup(const Config& config);
     std::string DataFilePath(const Detail::BlockId& blockId, bool activated) const;
     Status CommitFile(const Detail::BlockId& blockId, bool success) const;
+    std::vector<std::string> RelativeRoots() const;
+    const std::vector<std::string>& GetStorageBackends() const { return storageBackends_; }
```
mag1c-h: Why do we need this public interface? I don't see it being called anywhere.

ucm/integration/vllm/ucm_connector.py
UESTC-AHao: It usually won't lead to inconsistent final results; inconsistency is only possible when the model has been modified/tampered with.

ucm/store/posix/cc/global_config.h
UESTC-AHao: fixed

ucm/store/posix/cc/global_config.h
UESTC-AHao: fixed

ucm/store/test/e2e/cache_on_posix_test.py
UESTC-AHao: fixed

ucm/store/test/e2e/posixstore_embed.py
UESTC-AHao: fixed

ucm/store/posix/cc/hotness_tracker.cc
UESTC-AHao: fixed

ucm/store/posix/cc/hotness_tracker.cc
UESTC-AHao: fixed

ucm/store/posix/cc/hotness_tracker.h
UESTC-AHao: I will transfer them to the SpaceLayout and GC modules respectively.

ucm/store/posix/cc/hotness_tracker.h
UESTC-AHao: fixed

ucm/store/posix/cc/posix_store.cc
UESTC-AHao: fixed

ucm/store/posix/cc/posix_store.cc
UESTC-AHao: fixed

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: fixed

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: fixed

ucm/store/posix/cc/hotness_tracker.cc
UESTC-AHao: fixed

ucm/store/posix/cc/hotness_tracker.cc
UESTC-AHao: fixed

ucm/store/posix/cc/posix_store.cc
UESTC-AHao: fixed

ucm/store/posix/cc/posix_store.cc
UESTC-AHao: I will move it to the GC module.

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: fixed

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: This makes it easier to control memory.I will fix it.

ucm/store/posix/cc/shard_gc.h
UESTC-AHao: fixed

ucm/store/posix/cc/space_layout.h
UESTC-AHao: fixed

ucm/store/posix/cc/posix_store.cc
```diff
@@ -43,14 +43,17 @@ class PosixStore : public StoreV1 {
...
         s = spaceMgr_.Setup(config);
         if (s.Failure()) [[unlikely]] { return s; }
-        transEnable_ = config.deviceId >= 0;
+
         if (transEnable_) {
             s = transMgr_.Setup(config, spaceMgr_.GetLayout());
             if (s.Failure()) [[unlikely]] { return s; }
         }
-        ShowConfig(config);
         return Status::OK();
```
mag1c-h, line 54: Why should we modify this function?

ucm/store/posix/cc/posix_store.cc
mag1c-h, line 217: Newly added configuration items need to be logged in the startup log.

ucm/store/posix/cc/shard_gc.cc
```diff
@@ -0,0 +1,162 @@
...
+    if (!success) { return Status::Error("failed to start gc thread pool"); }
+
+    UC_INFO(
+        "ShardGC setup: recyclePercent={}, concurrency={}, maxFileCount={}, "
+        "thresholdRatio={}, checkIntervalSec={}",
+        config_.posixGcRecyclePercent, config_.posixGcConcurrency, maxFileCount_,
+        config_.posixGcTriggerThresholdRatio, config_.posixGcCheckIntervalSec);
+
+    gcCheckWorker_ = std::thread(&ShardGarbageCollector::GCCheckLoop, this);
+    while (!stop_.load() && ShouldTrigger()) { Execute(); }
```
mag1c-h: Why do we need to perform a recycling operation in Setup?

ucm/store/posix/cc/shard_gc.cc
```diff
@@ -0,0 +1,162 @@
...
+        gcPool_.Push({ShardTaskContext::Type::SAMPLE, shard, waiter, &sampledFiles});
+    }
+
+    waiter->Wait();
+
+    size_t avgFilesPerShard = sampledFiles.load() / sampleShards.size();
+    size_t thresholdFilesPerShard = maxFileCount_ / layout_->shards_.size();
+
+    UC_INFO("GC sampling: avgFiles/shard={}, threshold={}", avgFilesPerShard,
+            static_cast<size_t>(thresholdFilesPerShard * config_.posixGcTriggerThresholdRatio));
```
mag1c-h: This log may be printed many times in a short period of time.

ucm/store/posix/cc/space_layout.cc
```diff
@@ -144,4 +169,100 @@ std::string SpaceLayout::StorageBackend(const Detail::BlockId& blockId) const
...
+
+    return count;
+}
+
+std::vector<Detail::BlockId> SpaceLayout::GetOldestFiles(const std::string& shard,
+                                                         double recyclePercent,
+                                                         size_t maxRecycleCount) const
+{
+    std::string shardPath = storageBackends_.front();
+    if (shardPath.back() != '/') { shardPath += '/'; }
```
mag1c-h: This is redundant, storageBackends_ is guaranteed to end with '/' during Setup.

ucm/store/posix/cc/space_layout.h
```diff
@@ -31,14 +31,22 @@
 namespace UC::PosixStore {
 
 class SpaceLayout {
+private:
     std::vector<std::string> storageBackends_;
     bool dataDirShard_;
     size_t dataDirShardBytes_;
 
 public:
+    std::vector<std::string> shards_;
```
mag1c-h: Why should we public this?

ucm/store/posix/cc/space_manager.h
```diff
@@ -53,7 +58,7 @@ class SpaceManager {
     const SpaceLayout* GetLayout() const { return &layout_; }
 
 private:
-    uint8_t Lookup(const Detail::BlockId* block);
+    uint8_t Lookup(const Detail::BlockId& block);
```
mag1c-h: Why would we need to replace pointers with references?

ucm/store/posix/cc/posix_store.cc
UESTC-AHao: No modification is needed here after removing redundant logs. fixed

ucm/store/posix/cc/posix_store.cc
UESTC-AHao: fixed

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: fixed.Initial sampling should not block setup; it should be executed in a background thread.

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: fixed

ucm/store/posix/cc/space_layout.cc
UESTC-AHao: fixed

ucm/store/posix/cc/space_layout.h
UESTC-AHao: fixed

ucm/store/posix/cc/space_manager.h
UESTC-AHao: fixed

ucm/store/posix/cc/global_config.h
```diff
@@ -44,6 +44,14 @@ struct Config {
...
     size_t timeoutMs{30000};
     size_t dataDirShardBytes{3};
+    bool posixGcEnable{true};
+    double posixGcRecyclePercent{0.1};
+    size_t posixGcConcurrency{16};
+    size_t posixGcCheckIntervalSec{30};
+    size_t posixCapacityGb{0};
+    double posixGcTriggerThresholdRatio{0.7};
+    size_t posixGcMaxRecycleCountPerShard{1000};
+    double posixGcShardSampleRatio{0.1};
```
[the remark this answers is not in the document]
UESTC-AHao: The code uses the condition gcEnable_ && posixCapacityGb > 0, so when posixCapacityGb = 0, neither the heat tracker nor the GC manager will be initialized at all.

ucm/store/posix/cc/space_manager.cc
```diff
@@ -29,8 +29,15 @@ namespace UC::PosixStore {
...
 {
+    gcEnable_ = config.posixGcEnable;
     auto s = layout_.Setup(config);
     if (s.Failure()) [[unlikely]] { return s; }
+    if (gcEnable_ && config.posixCapacityGb > 0) {
+        s = hotnessTracker_.Setup(&layout_);
+        if (s.Failure()) [[unlikely]] { return s; }
+        s = gcMgr_.Setup(&layout_, config);
+        if (s.Failure()) [[unlikely]] { return s; }
+    }
```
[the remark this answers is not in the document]
UESTC-AHao: U r right, i will fix it

ucm/store/posix/cc/posix_store.cc
```diff
@@ -157,6 +166,33 @@ class PosixStore : public StoreV1 {
         } else {
             return Status::InvalidParam("invalid io engine({})", config.ioEngine);
         }
+        if (config.posixGcEnable && config.posixCapacityGb > 0) {
+            if (config.posixGcRecyclePercent <= 0 || config.posixGcRecyclePercent > 1.0) {
+                return Status::InvalidParam("invalid gc recycle percent({})",
+                                            config.posixGcRecyclePercent);
+            }
```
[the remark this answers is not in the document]
UESTC-AHao: Same as the issue mentioned above.fixed.

ucm/store/posix/cc/hotness_tracker.cc
```diff
@@ -0,0 +1,82 @@
...
+                std::this_thread::sleep_for(std::chrono::microseconds(100));
+                spinCount = 0;
+            }
+            continue;
+        }
+        spinCount = 0;
+        while (!consumeQueue.empty()) {
+            auto filePath = layout_->DataFilePath(consumeQueue.front(), false);
+            utime(filePath.c_str(), nullptr);
+            consumeQueue.pop_front();
```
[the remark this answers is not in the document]
UESTC-AHao: Updating mtime can indicate either the heat of a file that was just written or the heat of a file that was just looked up. The PR description is incorrect, and I will fix it.

ucm/store/posix/cc/shard_gc.cc
```diff
@@ -0,0 +1,151 @@
...
+    layout_ = layout;
+    config_ = config;
+    auto s = ValidateAndInitCapacity();
+    if (s.Failure()) { return s; }
+    auto success = gcPool_.SetWorkerFn([this](ShardTaskContext& ctx, auto&) { ProcessTask(ctx); })
+                       .SetNWorker(config_.posixGcConcurrency)
+                       .Run();
+    if (!success) { return Status::Error("failed to start gc thread pool"); }
+    gcCheckWorker_ = std::thread(&ShardGarbageCollector::GCCheckLoop, this);
+    return Status::OK();
```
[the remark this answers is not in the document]
UESTC-AHao: fixed

PR #778, opened 2026-03-04
[Bugfix] Correct MLA Threshold Gating for GSA

wangwenxin0312:
## Purpose
Add seq_lens retrieval across MLA/non-MLA and prefill/decode stages to determine whether GSA should be enabled.
## Modifications 
1、Added **get_seq_lens()** to centralize seq_lens retrieval.
- Non-MLA: use attn_metadata.seq_lens.
- MLA decode: use attn_metadata.decode.seq_lens.
- MLA prefill:
  - CUDA: use prefill.chunked_context.seq_lens.
  - NPU: use prefill.seq_lens.
- Added safe getattr guards to avoid attribute errors.

2、Location to modify self.gsa_enabled
## Test
python examples/offline_inference_gsaondevice.py
image

### DIFF
--- a/ucm/sparse/gsa_on_device/gsa_on_device.py
+++ b/ucm/sparse/gsa_on_device/gsa_on_device.py
@@ -783,10 +783,12 @@ class GSAOnDevice(UcmSparseBase):
                     )
                 else:  # NPU
                     self.cache_k_hash_gqa_npu(key, k_hash, attn_metadata)
+
+        if not self.gsa_enabled:
+            return query, key, value, output
+
         if self.is_mla:
             if phase == "decode":
-                if not self.gsa_enabled:
-                    return query, key, value, output
                 if self.is_cuda:
                     self.update_decode_topk_mla_cuda(
                         is_rollback_layer,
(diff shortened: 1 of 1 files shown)


ucm/sparse/gsa_on_device/gsa_on_device.py
```diff
@@ -1049,7 +1078,18 @@ def build_sparse_meta(
         if isinstance(attn_metadata, dict):
             attn_metadata = next(iter(attn_metadata.values()))
 
-        seq_lens = attn_metadata.seq_lens
+        self.has_decode = False
+        self.gsa_enabled = False
+        seq_lens = self.get_seq_lens(attn_metadata, self.is_mla, self.is_cuda)
```
Infinite666: No need to pass self variables into self method.

ucm/sparse/gsa_on_device/gsa_on_device.py
```diff
@@ -783,6 +785,8 @@ def attention_begin(
                     self.cache_k_hash_gqa_npu(key, k_hash, attn_metadata)
         if self.is_mla:
             if phase == "decode":
+                if not self.gsa_enabled:
```
Infinite666: Why don't we code like
```
if not self.gsa_enabled:
    return query, key, value, output
if self.is_mla:
    if phase == "decode":
        ...
else:
    if self.has_decode:
        ...
```

PR #782, opened 2026-03-06
[CI] replacing offline e2e test with online e2e test

dante159753:
## Purpose
replacing offline e2e test with online e2e test

## Modifications 
replacing offline e2e test with online e2e test
add lock logic for gpu test, ensure multiple agents won't use same gpu, increase success rate


test/common/online_inference_utils.py
```diff
@@ -0,0 +1,662 @@
+"""
```
Wwwzff, line 1: there's already a wrapper for online chats test/common/llm_connection/openai_connector.py

dante159753: remove it and reuse the existing one

PR #789, opened 2026-03-09
[Build] update gsa build scripts

Infinite666:
## Purpose
Update gsa build scripts for ascend ops to adpat to non-Ascend NPU environments.

## Modifications 
1. Add ascend-a3 platform, use it to tell if current build is for A3 environment.
2. Modify content for multiple build scripts:
`build.sh`: Entrance, setup environments and call other scripts.
`build_ops.sh`: Build for ascend ops, output *.run file.
`build_wheel.sh`: Build for python wheel, output *.whl file.
`install.sh`: Install .run and .whl packages.

## Test
Tested on dev environment.


ucm/sparse/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
 add_subdirectory(esa)
 add_subdirectory(gsa)
 add_subdirectory(gsa_on_device)
-add_subdirectory(kvstar)
+# add_subdirectory(kvstar)
```
ygwpz: Delete it directly.

PR #795, opened 2026-03-10
[Feat] Adding an environment variable input

Menglths:
## Purpose
This change focuses on clarifying token statistics, and cleaning up redundant logic in the benchmarking function.
## Modifications 
- Removed or reduced redundant variables and unused logic in the benchmarking process.
- Added the function of passing test parameters based on environment variables.
## Test
All tests have been passed.

### DIFF
--- a/test/common/llmperf/run_inference.py
+++ b/test/common/llmperf/run_inference.py
@@ -111,13 +111,13 @@ def run_test_cases(
         time.sleep(2)
 
         # ---------- prefill_2 ----------
-        prefill1_args = dict(args)
-        prefill1_args["mean_input"] = prefill_input
-        prefill1_args["mean_output"] = 2
+        prefill2_args = dict(args)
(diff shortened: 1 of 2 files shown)


test/suites/E2E/test_uc_performance.py
Potterluo, line 1: I think we should use English in code, for us to cooperate better

test/common/llmperf/run_inference.py
```diff
@@ -111,21 +111,14 @@ def run_sglang(args, prefill_input):
...
-        prefill2_args = dict(args)
-        prefill2_args.update(
-            {
-                "mean_input": 200,
-                "mean_output": 2,
-                "max_completed": 1,
-                "concurrent": 1,
-            }
-        )
+        prefill1_args = dict(args)
```
yuanzhg078: Why this comment is prefill_2, but use prefill_1 in code？

PR #800, opened 2026-03-10
[opt] Add GSA switch in config.properties

wangwenxin0312:
## Purpose
Add GSA switch in config.properties


examples/deployments/scripts/vllm/config.properties
```diff
@@ -89,6 +89,11 @@ enable_ascend_scheduler=false
 export ENABLE_UCM_PATCH=1
 ucm_enable=true
 use_layerwise=false
+
+# Enable GSA-on-device
+# VLLM_HASH_ATTENTION=1
```
Infinite666: Should need export at the front.

PR #821, opened 2026-03-16
[feat] log with rate limit

Lijiachen1018:
## Purpose
to implement a rate limit logger that suppress duplicate messages in a short time

## Modifications 
add a rate limit filter to the self-defined spdlog sink

### Usage of rate limit interface on python side
```python
def test_log(test: str):
    logger.info_once("test log info once")
    logger.info_limit("test log info limit %s", test)

for i in range(5):
    test_log(f"test{i}")
```

image

### will print when last log time has reached its  time to live which set to 10s in testing and 60s on production
```python
    import time
    for i in range(3):
        if i == 2:
            time.sleep(11)
        logger.info_limit("test message %d", i)
```

image

###  Usage of rate limit interface on CPP side
```cpp
        for (int i = 0;i < 5;i++) {
            UC_INFO_LIMIT("test message {}", i);
        }
```
image

### performance test
We run two performance tests on comparing the latency of duplicate filtering log interface and the plain log interface, by logging same messages / randomly generated message 2000 times.
image

### DIFF
--- a/ucm/logger.py
+++ b/ucm/logger.py
@@ -105,7 +105,7 @@ class Logger(logging.Logger):
             return msg % args
         return msg
 
-    def log(self, levelno, message, *args, exc_info=None, scope=None):
+    def log(self, levelno, message, *args, exc_info=None, scope=None, rate_limit=False):
         level = LevelMap[levelno]
         frame = inspect.currentframe()
         caller_frame = frame.f_back.f_back
@@ -116,6 +116,9 @@ class Logger(logging.Logger):
         if exc_info:
             exc_text = self.format_exception(exc_info)
             msg = msg + "\n" + exc_text
+        if rate_limit:
+            ucmlogger.log_rate_limit(level, file, func, line, msg)
+            return
         ucmlogger.log(level, file, func, line, msg)
 
     @staticmethod
@@ -147,6 +150,15 @@ class Logger(logging.Logger):
 
     def exception(self, message: str, *args: Hashable, **kwargs: Hashable):
         self.log(logging.ERROR, message, *args, **kwargs, exc_info=True)
+
+    def info_limit(self, message: str, *args, **kwargs):
+        self.log(logging.INFO, message, *args, **kwargs, rate_limit=True)
+
+    def warning_limit(self, message: str, *args, **kwargs):
+        self.log(logging.WARNING, message, *args, **kwargs, rate_limit=True)
+
+    def debug_limit(self, message: str, *args, **kwargs):
+        self.log(logging.DEBUG, message, *args, **kwargs, rate_limit=True)
 
 
 def init_logger(name: str = "UC") -> Logger:
--- a/ucm/shared/infra/logger/cc/spdlog_logger.cc
+++ b/ucm/shared/infra/logger/cc/spdlog_logger.cc
@@ -31,9 +31,14 @@
 #include <spdlog/spdlog.h>
 #include "compress_rotate_file_sink.h"
 #include "logger.h"
-
 namespace UC::Logger {
-
+constexpr uint64_t LIMIT_THRESHOLD_MS = 60000;
+constexpr uint32_t RATE_LIMIT_MAX_LOGS_PER_WINDOW = 3;
+constexpr uint32_t kRateLimitCountBits = 2;
+constexpr uint64_t kRateLimitCountMask = (1u << kRateLimitCountBits) - 1u;
+constexpr size_t kHashMixMagic = 0x9e3779b97f4a7c15ULL;
+constexpr size_t kHashShiftLeft = 12;
+constexpr size_t kHashShiftRight = 4;
 static spdlog::level::level_enum SpdLevels[] = {spdlog::level::debug, spdlog::level::info,
                                                 spdlog::level::warn, spdlog::level::err,
                                                 spdlog::level::critical};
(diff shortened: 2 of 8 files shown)


ucm/logger.py
```diff
@@ -133,17 +136,14 @@ def format_exception(e):
             s = s[:-1]
         return s
 
-    @lru_cache
```
Wwwzff, line 136: We might keep the old interfaces for their "print once" behavior, and add new interfaces for "print throttled".

ucm/shared/infra/logger/logger.cc
```diff
@@ -33,6 +33,15 @@ void Log(Level lv, std::string file, std::string func, int line, std::string msg
                               std::move(msg));
 }
 
+void LogRateLimit(Level lv, std::string file, std::string func, int line, std::string msg,
+                  std::string ori_msg)
+{
+    if (Logger::GetInstance().Filter(std::move(ori_msg))) {
```
Wwwzff: it's better to use func and line for generating hashkey and deduplication

ucm/shared/infra/logger/logger.h
```diff
@@ -32,6 +32,8 @@
 namespace UC::Logger {
 
 void Log(Level lv, std::string file, std::string func, int line, std::string msg);
+void LogRateLimit(Level lv, std::string file, std::string func, int line, std::string msg,
+                  std::string ori_msg);
```
mag1c-h: What does the ori_msg parameter mean?

ucm/shared/infra/logger/logger.h
```diff
@@ -46,9 +48,14 @@ bool isEnabledFor(Level lv);
 }  // namespace UC::Logger
 #define UC_SOURCE_LOCATION {__FILE__, __FUNCTION__, __LINE__}
 #define UC_LOG(lv, fmt, ...) UC::Logger::Log(lv, UC_SOURCE_LOCATION, FMT_STRING(fmt), ##__VA_ARGS__)
+#define UC_LOG_LIMIT(lv, fmt, ...) \
+    UC::Logger::LogRateLimit(lv, UC_SOURCE_LOCATION, FMT_STRING(fmt), ##__VA_ARGS__)
```
mag1c-h, line 61: The calling method does not match the definition, and it seems that the newly added macro has not been validated.

ucm/shared/infra/logger/logger.h
Lijiachen1018: message template

ucm/shared/infra/logger/logger.h
Lijiachen1018: fixed

ucm/shared/infra/logger/cc/spdlog_logger.cc
```diff
@@ -45,6 +45,39 @@ void Logger::Log(Level&& lv, SourceLocation&& loc, std::string&& msg)
...
 }
 
+bool Logger::Filter(std::string&& msg)
+{
+    using Clock = std::chrono::steady_clock;
+    const auto now = Clock::now();
+
+    const std::size_t key = std::hash<std::string>{}(msg);
+
+    std::lock_guard<std::mutex> lg(this->mutex_);
```
mag1c-h: This global mutex lock will likely reduce the performance of front-end callers.

ucm/shared/infra/logger/cc/spdlog_logger.cc
```diff
@@ -45,6 +44,26 @@ void Logger::Log(Level&& lv, SourceLocation&& loc, std::string&& msg)
...
+
+bool Logger::Filter(std::string&& msg)
+{
+    // using Clock = std::chrono::steady_clock;
+    uint64_t now = get_current_time_ms();
+    size_t slot_idx = std::hash<std::string>{}(msg) % HASH_SLOT_NUM;
+    SlotData& slot = hash_slots_[slot_idx];
+    // auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(now - slot.last_time);
+    if (now - slot.last_time <= LIMIT_THRESHOLD_MS) { return false; }
+    std::lock_guard<std::mutex> lock(slot.mtx);
```
Wwwzff: maybe use atomic value for saving timestamp and make it totally lock-free

ucm/shared/infra/logger/cc/spdlog_logger.cc
```diff
@@ -45,6 +44,26 @@ void Logger::Log(Level&& lv, SourceLocation&& loc, std::string&& msg)
...
+    auto now = std::chrono::system_clock::now();
+    auto ms = std::chrono::time_point_cast<std::chrono::milliseconds>(now);
+    return ms.time_since_epoch().count();
+}
+
+bool Logger::Filter(std::string&& msg)
+{
+    // using Clock = std::chrono::steady_clock;
+    uint64_t now = get_current_time_ms();
+    size_t slot_idx = std::hash<std::string>{}(msg) % HASH_SLOT_NUM;
```
Wwwzff: maybe make `std::hash<std::string>{}(msg) ` a static private member of the logger class.

ucm/shared/infra/logger/cc/spdlog_logger.cc
Lijiachen1018: fixed

ucm/shared/infra/logger/cc/spdlog_logger.cc
Lijiachen1018: fixed

ucm/shared/infra/logger/cc/spdlog_logger.cc
```diff
@@ -45,6 +44,33 @@ void Logger::Log(Level&& lv, SourceLocation&& loc, std::string&& msg)
...
+    std::hash<std::string_view> h;
+    size_t x = h(fv);
+    x ^= static_cast<size_t>(line) + 0x9e3779b97f4a7c15ULL + (x << 12) + (x >> 4);
+    x ^= h(ori_fmt) + 0x9e3779b97f4a7c15ULL + (x << 12) + (x >> 4);
+    const size_t slot_idx = x % HASH_SLOT_NUM;
+    std::atomic<uint64_t>& last_time = hash_slots_[slot_idx].last_time;
+    uint64_t last = last_time.load(std::memory_order_relaxed);
+    for (;;) {
+        if (now - last <= LIMIT_THRESHOLD_MS) { return false; }
+        if (last_time.compare_exchange_weak(last, now, std::memory_order_relaxed,
```
Wwwzff: maybe use compare_exchange_strong instead to avoid infinite loops

ucm/shared/infra/logger/cc/spdlog_logger.h
```diff
@@ -68,7 +72,16 @@ class Logger {
...
 
+    bool FilterCallSite(const char* file, int line, std::string_view ori_fmt);
+
 private:
+    struct SlotData {
+        std::atomic<uint64_t> last_time{0};
+    };
+
+    std::array<SlotData, HASH_SLOT_NUM> hash_slots_;
+    std::chrono::seconds rate_limit_window_{std::chrono::seconds(60)};
```
Wwwzff: unused param

ucm/shared/infra/logger/cc/spdlog_logger.cc
```diff
@@ -31,9 +31,11 @@
 #include <spdlog/spdlog.h>
 #include "compress_rotate_file_sink.h"
 #include "logger.h"
-
 namespace UC::Logger {
-
+constexpr uint64_t LIMIT_THRESHOLD_MS = 10000;
```
Wwwzff: should be 60s as you've mentioned in description

ucm/shared/infra/logger/cc/spdlog_logger.h
Lijiachen1018: fixed

ucm/shared/infra/logger/cc/spdlog_logger.cc
Lijiachen1018: fixed

ucm/shared/infra/logger/cc/spdlog_logger.cc
```diff
@@ -45,6 +47,32 @@ void Logger::Log(Level&& lv, SourceLocation&& loc, std::string&& msg)
     this->logger_->log(spdlog::source_loc{loc.file, loc.line, loc.func}, level, std::move(msg));
 }
 
+inline uint64_t get_current_time_ms()
```
mag1c-h: This function name does not conform to the overall naming convention.

ucm/shared/infra/logger/cc/spdlog_logger.cc
```diff
@@ -45,6 +47,32 @@ void Logger::Log(Level&& lv, SourceLocation&& loc, std::string&& msg)
...
+{
+    auto now = std::chrono::steady_clock::now();
+    auto ms = std::chrono::time_point_cast<std::chrono::milliseconds>(now);
+    return ms.time_since_epoch().count();
+}
+
+bool Logger::FilterCallSite(const char* file, int line, std::string_view ori_fmt)
+{
+    uint64_t now = get_current_time_ms();
+    const std::string_view fv(file ? file : "");
```
mag1c-h: In what scenarios might `file` be nullptr?

ucm/shared/infra/logger/cc/spdlog_logger.cc
```diff
@@ -45,6 +47,32 @@ void Logger::Log(Level&& lv, SourceLocation&& loc, std::string&& msg)
...
+    std::hash<std::string_view> h;
+    size_t x = h(fv);
+    x ^= static_cast<size_t>(line) + kHashMixMagic + (x << kHashShiftLeft) + (x >> kHashShiftRight);
+    x ^= h(ori_fmt) + kHashMixMagic + (x << kHashShiftLeft) + (x >> kHashShiftRight);
+    const size_t slot_idx = x % HASH_SLOT_NUM;
+    std::atomic<uint64_t>& last_time = hash_slots_[slot_idx].last_time;
+    uint64_t last = last_time.load(std::memory_order_relaxed);
+    if (now - last <= LIMIT_THRESHOLD_MS) { return false; }
```
mag1c-h: Perhaps it would be better to allow three log entries to be output within one minute?

ucm/shared/infra/logger/logger.h
```diff
@@ -32,23 +32,41 @@
...
 #define UC_DEBUG(fmt, ...) UC_LOG(UC::Logger::Level::DEBUG, fmt, ##__VA_ARGS__)
 #define UC_INFO(fmt, ...) UC_LOG(UC::Logger::Level::INFO, fmt, ##__VA_ARGS__)
 #define UC_WARN(fmt, ...) UC_LOG(UC::Logger::Level::WARN, fmt, ##__VA_ARGS__)
 #define UC_ERROR(fmt, ...) UC_LOG(UC::Logger::Level::ERROR, fmt, ##__VA_ARGS__)
-
+#define UC_DEBUG_LIMIT(fmt, ...) UC_LOG_LIMIT(UC::Logger::Level::DEBUG, fmt, ##__VA_ARGS__)
+#define UC_INFO_LIMIT(fmt, ...) UC_LOG_LIMIT(UC::Logger::Level::INFO, fmt, ##__VA_ARGS__)
+#define UC_WARN_LIMIT(fmt, ...) UC_LOG_LIMIT(UC::Logger::Level::WARN, fmt, ##__VA_ARGS__)
+#define UC_ERROR_LIMIT(fmt, ...) UC_LOG_LIMIT(UC::Logger::Level::ERROR, fmt, ##__VA_ARGS__)
```
mag1c-h: We should define the default macro as the frequency-limiting interface, and then add a non-frequency-limiting macro for special purposes.

ucm/shared/infra/logger/cc/spdlog_logger.cc
```diff
@@ -45,6 +49,46 @@ void Logger::Log(Level&& lv, SourceLocation&& loc, std::string&& msg)
...
+    const std::string_view fv(file ? file : "");
+    std::hash<std::string_view> h;
+    size_t x = h(fv);
+    x ^= static_cast<size_t>(line) + kHashMixMagic + (x << kHashShiftLeft) + (x >> kHashShiftRight);
+    x ^= h(ori_fmt) + kHashMixMagic + (x << kHashShiftLeft) + (x >> kHashShiftRight);
+    const size_t slot_idx = x % HASH_SLOT_NUM;
+    std::atomic<uint64_t>& slot = hash_slots_[slot_idx].rate_limit_state;
+
+    uint64_t s = slot.load(std::memory_order_relaxed);
+    const uint64_t window_start = s >> 16;
```
mag1c-h: Reserving 16 bits for `LimitCount` is too much.

ucm/shared/infra/logger/cc/spdlog_logger.cc
Lijiachen1018: fixed

ucm/shared/infra/logger/cc/spdlog_logger.cc
Lijiachen1018: fixed

ucm/shared/infra/logger/cc/spdlog_logger.cc
Lijiachen1018: fixed

ucm/shared/infra/logger/logger.h
Lijiachen1018: fixed

ucm/shared/infra/logger/cc/spdlog_logger.cc
Lijiachen1018: fixed

PR #825, opened 2026-03-16
[feat] log compatible with vllm v0.16.0

Lijiachen1018:
## Purpose
log compatible with vllm v0.16.0

## Modifications 
stub function `current_formatter_type`

### DIFF
--- a/ucm/integration/vllm/patch/logger_patch.py
+++ b/ucm/integration/vllm/patch/logger_patch.py
@@ -6,3 +6,4 @@ def patch_logger(mod):
     from ucm import logger
 
     patch_or_inject(mod, "init_logger", logger.init_logger)
+    patch_or_inject(mod, "current_formatter_type", logger.current_formatter_type)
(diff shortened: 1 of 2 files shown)


ucm/integration/vllm/patch/logger_patch.py
```diff
@@ -6,3 +6,4 @@ def patch_logger(mod):
     from ucm import logger
 
     patch_or_inject(mod, "init_logger", logger.init_logger)
+    patch_or_inject(mod, "current_formatter_type", logger.current_formatter_type)
```
ygwpz, line 9: Will there be a problem in version 0.11.0?

Wwwzff (approved):

ucm/integration/vllm/patch/logger_patch.py
Wwwzff: > Will there be a problem in version 0.11.0?

The interface doesn't show up until version 0.14.1, so it should be okay to run on older versions.

PR #829, opened 2026-03-16
[Fix] Fix thread pool timeout lifecycle and posix prefix lookup completion

Tarrei:
## Purpose

Fix the timeout lifecycle in `ThreadPool` and make `PosixStore` prefix lookup use that timeout path safely.

The main issue was that a timed-out worker could still continue running after the timeout callback, while the caller-side latch could already be released. In `SpaceManager` prefix lookup, that created an inconsistent completion signal and unsafe lifetime assumptions for shared lookup context.

### ThreadPool
- Updated `ThreadPool` timeout handling to avoid detaching timed-out worker threads.
- Added explicit worker state tracking:
  - `Idle`
  - `Running`
  - `TimeoutRequested`
  - `Exited`
- Changed monitor behavior:
  - timeout detection now only notifies and requests worker stop
  - worker cleanup happens only after the worker actually exits
- Made `stop_` an `std::atomic<bool>` to remove the cross-thread data race between destructor, worker threads, and monitor thread.
- Kept the existing public `SetWorkerFn(...)` / `SetWorkerTimeoutFn(...)` interface unchanged for compatibility.

### PosixStore SpaceManager
- Adjusted prefix lookup timeout handling so the timeout callback only updates shared status and no longer decrements the waiter.
- Kept waiter completion on the worker normal path only, so each lookup task completes exactly once.
- Reduced lookup task fan-out from fixed `NWorker()` tasks to `min(num, NWorker())` tasks to avoid unnecessary worker submissions for small requests.

## Test
image
image

### DIFF
--- a/ucm/shared/infra/thread/thread_pool.h
+++ b/ucm/shared/infra/thread/thread_pool.h
@@ -53,15 +53,12 @@ class ThreadPool {
         bool StopRequested() const noexcept { return this->flag_->load(std::memory_order_relaxed); }
     };
 
-    enum class WorkerState { Idle, Running, TimeoutRequested, Exited };
-
     struct Worker {
         ssize_t tid;
         std::thread th;
         StopToken stop;
         std::weak_ptr<Task> current;
         std::atomic<std::chrono::steady_clock::time_point> tp{};
-        std::atomic<WorkerState> state{WorkerState::Idle};
     };
 
 public:
@@ -72,10 +69,9 @@ public:
     {
         {
             std::lock_guard<std::mutex> lock(this->taskMtx_);
-            this->stop_.store(true, std::memory_order_relaxed);
+            this->stop_ = true;
             this->cv_.notify_all();
         }
-        for (auto& worker : this->workers_) { worker->stop.RequestStop(); }
         if (this->monitor_.joinable()) { this->monitor_.join(); }
         for (auto& worker : this->workers_) {
             if (worker->th.joinable()) { worker->th.join(); }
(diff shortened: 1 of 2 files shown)


ucm/store/posix/cc/space_manager.cc
```diff
@@ -130,6 +131,5 @@ void SpaceManager::OnLookupPrefixTimeout(PrefixLookupContext& ctx)
     auto ok = Status::OK().Underlying();
     auto timeout = Status::Timeout().Underlying();
     ctx.status->compare_exchange_weak(ok, timeout, std::memory_order_acq_rel);
-    ctx.waiter->Done();
```
mag1c-h, line 133: `LookupOnPrefix` might get blocked in the kernel due to underlying network issues. In this case, `OnLookupPrefixTimeout` needs to update the state and wake up the `waiter`, otherwise, the `waiter` will wait indefinitely.
The issue of resources being prematurely released, does not exist because `LookupOnPrefix` itself holds a smart pointer to the context.

ucm/shared/infra/thread/thread_pool.h
```diff
@@ -160,49 +164,68 @@ class ThreadPool {
...
+            worker->stop.RequestStop();
+        }
+    }
+
+    size_t CleanupExitedWorkers()
+    {
+        for (auto it = this->workers_.begin(); it != this->workers_.end();) {
+            auto state = (*it)->state.load(std::memory_order_relaxed);
+            if (state == WorkerState::Exited) {
+                if ((*it)->th.joinable()) { (*it)->th.join(); }
```
mag1c-h: Worker threads may never reach the `Exited` state due to being suspended. In addition, new threads are only started after the old threads exit. If all Workers are in a suspended state, there will be no available threads in the thread pool.

PR #849, opened 2026-03-21
[fix]Add timeout in PosixStore working thread

qyh111:
## Purpose

- Prevent Posix load/dump worker threads from blocking indefinitely when storage (e.g. a hung mount) stalls synchronous I/O, by wiring transfer pools to the existing `ThreadPool` worker-timeout path (`config.timeoutMs`).

## Modifications

- **`ucm/store/posix/cc/trans_queue.{h,cc}`**  
  - `SetWorkerTimeoutFn` on `loadPool_` and `dumpPool_` using `config.timeoutMs`.  
  - `OnIoUnitTimeout`: insert `ios.owner` into `failureSet_` (if not already present) and `ios.waiter->Done()` so other shards bail out and the task latch can drain; hung workers are detached per `ThreadPool` monitor behavior.

- **`ucm/shared/infra/thread/latch.h`**  
  - Add `WaitForDuration(timeoutMs)`: wait up to `timeoutMs` **from call time** for `counter_ == 0`, unlike `WaitFor` (remaining budget from `startTp`).

- **`ucm/store/detail/template/task_wrapper.h`**  
  - On `WaitFor(timeoutMs_)` failure: `failureSet_.Insert(taskId)` then drain with `WaitForDuration(drainSliceMs)` in a loop; periodic `UC_WARN` uses `taskId`.

### DIFF
--- a/ucm/store/posix/cc/trans_queue.cc
+++ b/ucm/store/posix/cc/trans_queue.cc
@@ -38,25 +38,24 @@ Status TransQueue::Setup(const Config& config, TaskIdSet* failureSet, const Spac
     auto success =
         loadPool_.SetNWorker(config.dataTransConcurrency)
             .SetWorkerFn([this](auto& ios, auto&) { LoadWorker(ios); })
-            .SetWorkerTimeoutFn([this](IoUnit& ios, ssize_t tid) { OnIoUnitTimeout(ios, tid); },
+            .SetWorkerTimeoutFn([this](IoUnit& ios, ssize_t tid) { OnIoUnitTimeout(ios); },
                                 config.timeoutMs)
             .Run();
     if (!success) [[unlikely]] {
         return Status::Error(fmt::format("workers({}) start failed", config.dataTransConcurrency));
     }
-    success =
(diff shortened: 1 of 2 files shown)


ucm/store/posix/cc/trans_queue.cc
```diff
@@ -35,21 +35,33 @@ Status TransQueue::Setup(const Config& config, TaskIdSet* failureSet, const Spac
...
+            .SetWorkerTimeoutFn([this](IoUnit& ios, ssize_t tid) { OnIoUnitTimeout(ios, tid); },
+                                config.timeoutMs)
+            .Run();
     if (!success) [[unlikely]] {
         return Status::Error(fmt::format("workers({}) start failed", config.dataTransConcurrency));
     }
     return Status::OK();
 }
 
+void TransQueue::OnIoUnitTimeout(IoUnit& ios, ssize_t tid)
```
mag1c-h: tid is an unused parameter.

qyh111: fixed

PR #852, opened 2026-03-21
[feat] Store provides the ability to configure CPU affinity.

mag1c-h:
## Purpose

Store provides the ability to configure CPU affinity, Engine connector can configure the CPU cores used when Dumping and Loading of the KVCache as needed.

## Modifications 

- New config item: `cpu_affinity_cores`, type list[int]
- Each thread in the Store performs an affinity configuration once it starts
- Configuration failures are logged but do not affect existing functionality

### DIFF
--- a/ucm/store/cache/cc/dump_queue.cc
+++ b/ucm/store/cache/cc/dump_queue.cc
@@ -23,6 +23,7 @@
  * */
 #include "dump_queue.h"
 #include "logger/logger.h"
+#include "thread/cpu_affinity.h"
 
 namespace UC::CacheStore {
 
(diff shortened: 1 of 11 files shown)


ucm/store/cache/cc/dump_queue.cc
```diff
@@ -66,6 +68,8 @@ void DumpQueue::DispatchStage(std::promise<Status>& started)
     auto s = stream.Setup(deviceId_, streamNumber_);
     started.set_value(s);
     if (s.Failure()) [[unlikely]] { return; }
+    s = CpuAffinity::SetCpuAffinity4CurrentThread(cpuAffinityCores_);
+    if (!cpuAffinityCores_.empty() && s.Failure()) { UC_WARN("Failed({}) to set affinity.", s); }
```
qyh111: Why not skip setting affinity when cpuAffinityCores_ is empty？If store doesn't setup with cpuAffinityCores_, this warning log will show here, is this expected?

mag1c-h: fixed

ygwpz (approved):

PR #854, opened 2026-03-23
[feat] Add NUMA-aware CPU core split for vllm worker and store threads

wangwenxin0312:
## Purpose
Improve CPU affinity by assigning worker and store cores based on NUMA locality for CUDA and Ascend devices.

## Modifications 
- Detect NUMA node from CUDA PCI info
- Support Ascend visible-device mapping
- Split NUMA cpulist into worker/store groups
## Test
Enable by setting the environment variable **VLLM_CPU_AFFINITY**.

### DIFF
--- a/ucm/integration/vllm/device.py
+++ b/ucm/integration/vllm/device.py
@@ -45,69 +45,6 @@ class Device(ABC):
         """
         pass
 
-    def _fallback_cpu_affinity(self, local_rank: int) -> Optional[str]:
-        """
-        Fallback path for both CUDA and NPU:
-        split current allowed CPUs by local_rank within visible devices.
-        """
-        try:
-            cores = sorted(os.sched_getaffinity(0))
-            if not cores:
-                return None
-
-            if current_platform.is_cuda_alike():
-                visible = os.environ.get("CUDA_VISIBLE_DEVICES")
-                total_devices = (
-                    len([x.strip() for x in visible.split(",") if x.strip()])
-                    if visible
-                    else torch.cuda.device_count()
-                )
-                dev_name = "cuda"
-            elif current_platform.device_type == "npu":
-                visible = os.environ.get("ASCEND_RT_VISIBLE_DEVICES") or os.environ.get(
-                    "ASCEND_VISIBLE_DEVICES"
-                )
-                total_devices = (
-                    len([x.strip() for x in visible.split(",") if x.strip()])
-                    if visible
-                    else torch.npu.device_count()
-                )
-                dev_name = "npu"
-            else:
-                return None
-
-            if total_devices <= 0 or local_rank < 0 or local_rank >= total_devices:
-                logger.warning(
-                    f"[CPU Affinity] invalid fallback split: "
-                    f"{dev_name} local_rank={local_rank}, total_devices={total_devices}"
-                )
-                return None
-
-            base = len(cores) // total_devices
-            extra = len(cores) % total_devices
-
-            start = local_rank * base + min(local_rank, extra)
-            length = base + (1 if local_rank < extra else 0)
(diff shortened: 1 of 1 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -329,8 +442,25 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
+            if enable_affinity
+            else (None, None)
+        )
+
+        self.store = self._create_store(self.kv_cache_layout, store_cores)
         self.device = create_device()
+
+        if worker_cores:
+            try:
+                os.sched_setaffinity(0, worker_cores)
```
mag1c-h, line 349: Does setting affinity in `register_kv_caches` affect all threads in Worker?

Infinite666: Seems it won't affect the threads that created before `register_kv_caches`, let's take this as an issue and deal with it later.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -300,9 +303,121 @@ def _create_store(
...
+            else:
+                visible = os.environ.get("ASCEND_RT_VISIBLE_DEVICES") or os.environ.get(
+                    "ASCEND_VISIBLE_DEVICES"
+                )
+
+                device_id = local_rank
+                if visible:
+                    dev_list = [int(x.strip()) for x in visible.split(",") if x.strip()]
+                    if local_rank < len(dev_list):
+                        device_id = dev_list[local_rank]
```
Infinite666: Why is it only the NPU environment that needs to obtain the device from visible devices list? The `local_rank` is the same as physical device id in cuda but not in ascend?

ucm/integration/vllm/ucm_connector.py
```diff
@@ -300,9 +303,121 @@ def _create_store(
...
+
+                if hasattr(self, "npu_to_numa"):
+                    numa_node = self.npu_to_numa.get(device_id)
```
Infinite666: Missing definition of `npu_to_numa`?

ucm/integration/vllm/ucm_connector.py
wangwenxin0312: CUDA can directly use local_rank because **torch.cuda.get_device_properties(local_rank)** returns the real PCI information of the underlying physical GPU.

ucm/integration/vllm/ucm_connector.py
wangwenxin0312: Thank you for your suggestion. Have removed it.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -329,8 +440,25 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
         }
         self.first_layer_id = next(iter(self.layer_name_to_id.values()))
 
-        self.store = self._create_store(self.kv_cache_layout)
+        is_cuda = self._vllm_config.device_config.device_type == "cuda"
```
qyh111: I think current_platform.is_cuda_alike() is better

ucm/integration/vllm/ucm_connector.py
```diff
@@ -300,9 +303,117 @@ def _create_store(
...
+
+        worker_cores, store_cores = [], []
+        pci_bus_id, numa_node = None, None
+
+        try:
+            # =========================
+            # CUDA path
+            # =========================
+            if is_cuda:
+                prop = torch.cuda.get_device_properties(local_rank)
```
qyh111: `ucm/integration/vllm/device.py`  already exits for runtime difference adaptation already exists, it's better to put the new parts together with them.

PR #857, opened 2026-03-23
[CI] Add gsa online test

dante159753:
## Purpose
Add gsa online test, same test logic as offline test

## Modifications 
translate offline test case to online test, refine the online test framework

### DIFF
--- a/docs/source/index.md
+++ b/docs/source/index.md
@@ -49,7 +49,6 @@ Paper list:
 getting-started/quickstart_vllm
 getting-started/quickstart_vllm_ascend
 getting-started/quickstart_sglang
-getting-started/quickstart_mindie_llm
 getting-started/kv_cache_calculator
 :::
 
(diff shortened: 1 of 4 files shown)


test/common/online_inference_utils.py
```diff
@@ -292,6 +297,65 @@ def __exit__(self, exc_type, exc_val, exc_tb) -> None:
...
+            executor.submit(_send_request, i, req): i for i, req in enumerate(requests)
+        }
+
+        # Collect results as they complete
+        for future in as_completed(future_to_index):
+            index, response = future.result()
+            results[index] = response
+
+    # Ensure all results are filled (should not happen if no exception)
+    return [resp for resp in results if resp is not None]
```
yuanzhg078: Could filtering out None responses break index alignment with standard_answers and lead to incorrect evaluation results?

dante159753: good point, return the results list, and add check before that

Wwwzff (approved):
LGTM

PR #865, opened 2026-03-25
[opt] Enhance NPU CPU affinity resolution with NUMA fallback

wangwenxin0312:
## Purpose
NUMA info: use real topology to map devices to NUMA nodes.  NPU -> PCIe -> NUMA -> cpulist
NUMA info v2: Fallback: evenly distribute devices across NUMA nodes based on lscpu when topology information is unavailable.
## Modifications 
ucm/integration/vllm/device.py

### DIFF
--- a/ucm/integration/vllm/device.py
+++ b/ucm/integration/vllm/device.py
@@ -23,33 +23,6 @@ from ucm.logger import init_logger
 logger = init_logger(__name__)
 
 
-def execute_command(cmd_list):
-    with subprocess.Popen(
-        cmd_list, shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE
-    ) as p:
-        out, err = p.communicate(timeout=1000)
-    res = out.decode()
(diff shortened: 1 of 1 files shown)


ucm/integration/vllm/device.py
```diff
@@ -21,6 +23,33 @@
...
+    with subprocess.Popen(
+        cmd_list, shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    ) as p:
+        out, err = p.communicate(timeout=1000)
+    res = out.decode()
+    return res
+
+
+@dataclass
+class DeviceInfo:
```
Infinite666: 1. Make it a inner class for `NpuDevice` since just it will only be used by `NpuDevice` and avoid ambiguity between gpu.
2. (Optional) Maybe `numa_id`, `pcie_info` and other alike info can also be added in this class, so that we can get rid of so many different dicts.

PR #870, opened 2026-03-26
[refactor] Remove NVML-based CPU affinity setting

wangwenxin0312:
## Purpose
Move CPU affinity control.
## Modifications 
ucm/shared/trans/cuda/cuda_device.cc
Remove SetCpuAffinity implementation that sets CPU affinity via NVML.

### DIFF
--- a/ucm/shared/trans/cuda/cuda_device.cc
+++ b/ucm/shared/trans/cuda/cuda_device.cc
@@ -22,7 +22,6 @@
  * SOFTWARE.
  * */
 #include <cuda_runtime.h>
-#include <nvml.h>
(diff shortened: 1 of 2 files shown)


ucm/shared/trans/cuda/cuda_device.cc
mag1c-h: Remove the relevant header files and libraries linked in CMake as well.

wangwenxin0312: Thanks for your suggestion. Have removed it.

mag1c-h (approved):

PR #871, opened 2026-03-27
[Feat] adapt for DSA model on CUDA platform

sumingZero:
## Purpose
Adapt KV cache layout handling for DSA models on CUDA platform and make layer-wise KV cache ordering explicit.

## Modifications
- Add support for DSA model KV cache layout on CUDA, where additional per-layer index cache is present.
- Keep existing tensor pointer / stride extraction logic unchanged for supported KV cache tensor formats.

## Test
- Tested GLM5 on H100.

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -18,6 +18,7 @@ from vllm.distributed.kv_transfer.kv_connector.v1.base import (
     KVConnectorRole,
 )
 from vllm.distributed.parallel_state import get_tp_group, get_world_group
+from vllm.distributed.utils import get_pp_indices
 from vllm.model_executor.models.utils import extract_layer_index
 from vllm.platforms import current_platform
 from vllm.v1.core.sched.output import SchedulerOutput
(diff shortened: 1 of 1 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -74,13 +74,26 @@ def __init__(
...
         if self.pp_size > 1 and self.num_hidden_layers <= 0:
             raise ValueError("num_hidden_layers must be > 0 when pp_size > 1")
+        self.layer_name_to_id = {
+            name: self._extract_layer_index(name) for name in kvcaches.keys()
+        }
         self._build_layout(kvcaches)
 
+    @staticmethod
+    def _extract_layer_index(layer_name: str) -> Optional[int]:
+        """
```
qyh111: use from vllm.model_executor.models.utils import extract_layer_index maybe better

ucm/integration/vllm/ucm_connector.py
```diff
@@ -74,13 +74,26 @@ def __init__(
         )
         if self.pp_size > 1 and self.num_hidden_layers <= 0:
             raise ValueError("num_hidden_layers must be > 0 when pp_size > 1")
+        self.layer_name_to_id = {
+            name: self._extract_layer_index(name) for name in kvcaches.keys()
```
qyh111: I think self.layer_name_to_id  already exist

ucm/integration/vllm/ucm_connector.py
```diff
@@ -74,13 +74,26 @@ def __init__(
...
+
     def _build_layout(self, kvcaches):
-        raw_ptr_rows = []
-        stride_rows = []
+        raw_ptr_rows = [[] for _ in range(self.num_hidden_layers)]
+        stride_rows = [[] for _ in range(self.num_hidden_layers)]
 
-        for _, kv_layer in kvcaches.items():
+        for layer_name, kv_layer in kvcaches.items():
             ptrs = []
```
qyh111, line 96: Have you consider PP parallel ?

ucm/integration/vllm/ucm_connector.py
sumingZero: Modified

ucm/integration/vllm/ucm_connector.py
sumingZero: Yes, I moved it to KVCacheLayout

ucm/integration/vllm/ucm_connector.py
sumingZero: You are right. I didn't account for PP. Fixed now by correcting the layer id to local layer id.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -74,13 +75,17 @@ def __init__(
...
+        self.layer_name_to_id = {
+            name: extract_layer_index(name) for name in kvcaches.keys()
+        }
+        self.first_layer_id = next(iter(self.layer_name_to_id.values()))
         self._build_layout(kvcaches)
 
     def _build_layout(self, kvcaches):
-        raw_ptr_rows = []
-        stride_rows = []
+        raw_ptr_rows = [[] for _ in range(self.num_hidden_layers)]
```
flesher0813: 这会扩充成所有层ptrs，而不是每个pp组分到的层数

sumingZero: You're right. I've fixed the code to calculate the number of layers for the current worker.

PR #874, opened 2026-03-27
[feat] Add tensor parallel size support and update GPU memory utilization for online inference tests

hmy98213:
## Modifications 

host changed from "0.0.0.0" to "127.0.0.1"
gpu_mem default set to 60000, 6000 is not enough (tests always fail on H20)
added the tensor_parallel_size parameter and a test with tensor_parallel_size parameter = 2

### DIFF
--- a/test/suites/E2E/test_online_inference.py
+++ b/test/suites/E2E/test_online_inference.py
@@ -22,12 +22,19 @@ import yaml
 from common.online_inference_utils import hbm_ssd_mixed_test
 from common.path_utils import get_path_relative_to_test_root, get_path_to_model
 
+os.environ["ENABLE_UCM_PATCH"] = "1"
+
+# These online E2E cases start a fresh vLLM instance with UCM enabled and a
+# relatively large max_model_len, so the previous 6 GB hint translated to an
+# unrealistically small --gpu-memory-utilization (~6%) on H20 GPUs.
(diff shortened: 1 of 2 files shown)


test/suites/E2E/test_online_inference.py
```diff
@@ -22,12 +22,19 @@
 from common.online_inference_utils import hbm_ssd_mixed_test
 from common.path_utils import get_path_relative_to_test_root, get_path_to_model
 
+os.environ["ENABLE_UCM_PATCH"] = "1"
```
yuanzhg078, line 25: Already set in dockerfile. Maybe no need  in test here.

test/suites/E2E/test_online_inference.py
```diff
@@ -202,3 +209,76 @@ def test_online_accuracy_hbm_ssd_mixed_pp(
...
+
+        ucm_storage_dir = "/tmp/ucm_cache"
+        served_model_name = model_name
+        tokenizer_path = f"/home/models/{model_name}"
+        model_path = get_path_to_model(model_name, config)
+
+        # Build UCM config
+        ucm_config = {
+            "use_layerwise": (
+                use_layerwise if ucm_connector_name != "UcmNfsStore" else False
```
flesher0813, line 246: Do we still need this check for only setting _use_layerwise_ to True

PR #877, opened 2026-03-30
[CI] add test set config

dante159753:
## Purpose

- add pytest cases without starting server
- add test set cases
- add support for clear hbm

### DIFF
--- a/examples/ucm_config_example.yaml
+++ b/examples/ucm_config_example.yaml
@@ -17,6 +17,7 @@ ucm_connectors:
       storage_backends: "/mnt/test"
       io_direct: false
       # cache_buffer_capacity_gb: 256
+      # posix_capacity_gb: 1024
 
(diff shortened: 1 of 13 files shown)


test/test_set/test_set.yaml
```diff
@@ -0,0 +1,236 @@
...
+            ucm_connector_config:
+              store_pipeline: "Cache|Posix"
+              storage_backends: "/mnt/kvcache-local"
+              io_direct: false
+              cache_buffer_capacity_gb: 32
+
+        # When you use UcmNfsStore, you should set enable_event_sync to false.
+        enable_event_sync: true
+      environment:
+        - ENABLE_UCM_PATCH: "1"
```
yuanzhg078, line 64: Maybe you should set **VLLM_SERVER_DEV_MODE=1**here to use reset method

dante159753: added, it will work after vllm 0.14.0

PR #883, opened 2026-03-31
[BugFix] Fix the bug where non-DP0 processes fail to update file hotness in DP scenarios.

UESTC-AHao:

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -336,10 +336,11 @@ class UCMDirectConnector(KVConnectorBase_V1):
                 * (1 if self.is_mla else self.num_head * 2)
                 * self.blocks_per_chunk
             )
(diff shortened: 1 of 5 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -336,10 +336,10 @@ def _create_store(
                 * (1 if self.is_mla else self.num_head * 2)
                 * self.blocks_per_chunk
             )
-        # GC only enabled for Scheduler with data_parallel_rank == 0
-        dp_rank = self._vllm_config.parallel_config.data_parallel_rank
-        if self._role == KVConnectorRole.WORKER or dp_rank != 0:
-            config["posix_gc_enable"] = False
+        config["posix_hotness_tracker_enable"] = self._role != KVConnectorRole.WORKER
```
sumingZero, line 340: dp_rank is not defined

PR #884, opened 2026-03-31
[BugFix] Fix the bug where non-DP0 processes fail to update file hotness in DP scenarios.

UESTC-AHao:

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -336,10 +336,10 @@ class UCMDirectConnector(KVConnectorBase_V1):
                 * (1 if self.is_mla else self.num_head * 2)
                 * self.blocks_per_chunk
             )
-        # GC only enabled for Scheduler with data_parallel_rank == 0
-        dp_rank = self._vllm_config.parallel_config.data_parallel_rank
-        if self._role == KVConnectorRole.WORKER or dp_rank != 0:
-            config["posix_gc_enable"] = False
+        dp_rank = self._vllm_config.parallel_config.rank
+        config["posix_gc_enable"] = (
+            self._role != KVConnectorRole.WORKER and dp_rank == 0
+        )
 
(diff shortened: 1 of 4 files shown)


ucm/store/posix/cc/global_config.h
```diff
@@ -44,13 +44,14 @@ struct Config {
     size_t commitConcurrency{4};
     size_t timeoutMs{30000};
     size_t dataDirShardBytes{3};
+    bool posixHotnessTrackerEnable{true};
```
mag1c-h: Updating the hotness of blocks should be something that every scheduler should do by default, and should not be controlled by a specific switch.

UESTC-AHao: fixed

ucm/integration/vllm/ucm_connector.py
```diff
@@ -336,10 +336,10 @@ def _create_store(
...
                 * self.blocks_per_chunk
             )
-        # GC only enabled for Scheduler with data_parallel_rank == 0
-        dp_rank = self._vllm_config.parallel_config.data_parallel_rank
-        if self._role == KVConnectorRole.WORKER or dp_rank != 0:
-            config["posix_gc_enable"] = False
+        dp_rank = self._vllm_config.parallel_config.rank
+        config["posix_gc_enable"] = (
+            self._role != KVConnectorRole.WORKER and dp_rank == 0
+        )
```
mag1c-h, line 342: Why should we modify this? If the user has disabled GC in their configuration, this will override the user's configuration to enable it.

UESTC-AHao: posix_gc_enable serves as an internal flag to determine whether a process is permitted to initiate Garbage Collection (GC). GC will only be enabled if posix_gc_enable is true and the user has configured posixCapacityGb.  as the actual user-facing switch for enabling GC is posixCapacityGb.

PR #892, opened 2026-04-03
[Feat]Add Chunk size

qyh111:
## Test
image
image

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -63,12 +63,12 @@ class RequestDispatchMeta:
 
 class KVCacheLayout:
     def __init__(
-        self, kvcaches, use_layerwise: bool, vllm_config: "VllmConfig"
+        self, kvcaches, launch_config: dict, vllm_config: "VllmConfig"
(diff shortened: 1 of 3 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -422,12 +425,15 @@ def get_num_new_matched_tokens(
 
         total_hit_block_num = hbm_hit_block_num + external_hit_blocks
 
-        external_hit_tokens = external_hit_blocks * self.block_size
+        external_hit_tokens = 0
+        if external_hit_blocks > 0:
+            remainder = num_computed_tokens % self.chunk_size
+            external_hit_tokens = external_hit_blocks * self.chunk_size - remainder
```
flesher0813, line 431: Should this be external_hit_tokens = (external_hit_blocks - remainder) * self.chunk_size?

sumingZero, line 431: Does this imply that because the initial chunk in UCM may contain blocks already present in HBM, we must deduct that overlap to accurately compute the number of external hit tokens?

PR #895, opened 2026-04-03
[opt] Prevent Load starvation by reserving buffers for high-priority operations

mag1c-h:
## Purpose
This PR introduces a reserved buffer mechanism to prevent Load operations from being starved when Dump operations consume all available buffers slowly.
In scenarios where Dump operations process buffers slowly, they can occupy all buffers, causing Load operations to wait indefinitely. By reserving a portion of buffers exclusively for Load operations, we ensure system responsiveness and prevent deadlocks.

## Modifications
- Added loadExclusiveBufferNumber configuration parameter (default 1024) to specify the number of buffers reserved for Load operations.
- Modified buffer allocation logic to respect reserved buffers: ordinary allocations cannot use the last loadExclusiveBufferNumber buffers, while Load operations (with allowReserved=true) can access all buffers.
- Updated LoadQueue::DispatchOneTask to request buffers with allowReserved=true to ensure Load tasks can always get buffers.
- Added comprehensive tests for reserved buffer behavior.

## Test
- Added GetReservedNode test to verify reserved buffer allocation logic.
- Existing tests pass with updated configurations.

## Configuration
New config parameter: `cache_load_exclusive_buffer_number` (default: 1024)

### DIFF
--- a/ucm/store/cache/cc/trans_buffer.cc
+++ b/ucm/store/cache/cc/trans_buffer.cc
@@ -70,12 +70,13 @@ protected:
         int32_t deviceId{-1};
         size_t nodeSize{0};
         size_t totalSize{0};
+        size_t reservedNumber{0};
     };
     BaseConfig base_;
 
 public:
-    BufferStrategy(int32_t deviceId, size_t nodeSize, size_t totalSize)
-        : base_({deviceId, nodeSize, totalSize})
+    BufferStrategy(int32_t deviceId, size_t nodeSize, size_t totalSize, size_t reservedNumber)
+        : base_({deviceId, nodeSize, totalSize, reservedNumber})
     {
     }
     virtual ~BufferStrategy() = default;
@@ -86,7 +87,7 @@ public:
     virtual void NodeLock(size_t iNode) = 0;
     virtual void NodeUnlock(size_t iNode) = 0;
     virtual size_t& FirstAt(size_t iBucket) = 0;
-    virtual size_t FetchNode() = 0;
+    virtual size_t FetchNode(bool allowReserved) = 0;
     virtual void* DataAt(size_t iNode) = 0;
     virtual BufferMetaNode* MetaAt(size_t iNode) = 0;
 };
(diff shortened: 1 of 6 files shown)


ucm/store/cache/cc/trans_buffer.cc
```diff
@@ -536,10 +545,11 @@ size_t TransBuffer::FindAt(size_t iBucket, const Detail::BlockId& blockId, size_
...
-size_t TransBuffer::Alloc(const Detail::BlockId& blockId, size_t shardIdx, size_t iBucket)
+size_t TransBuffer::Alloc(const Detail::BlockId& blockId, size_t shardIdx, size_t iBucket,
+                          bool allowReserved)
 {
     for (;;) {
-        auto iNode = strategy_->FetchNode();
+        auto iNode = strategy_->FetchNode(allowReserved);
         auto meta = strategy_->MetaAt(iNode);
         strategy_->NodeLock(iNode);
         if (meta->reference > 0) {
```
UESTC-AHao, line 555: If all buffers are occupied here, will it lead to busy-waiting? Is it necessary to yield the CPU?

ucm/store/cache/cc/trans_buffer.cc
```diff
@@ -410,11 +413,12 @@ class SharedBufferStrategy : public BufferStrategy {
...
     void NodeUnlock(size_t iNode) override { header_->nodeLocks[iNode].Unlock(); }
     size_t& FirstAt(size_t iBucket) override { return header_->buckets[iBucket]; }
-    size_t FetchNode() override
+    size_t FetchNode(bool allowReserved) override
     {
+        const auto limit = header_->nNode - (allowReserved ? 0 : base_.reservedNumber);
         header_->lock.Lock();
-        auto iNode = header_->freeHead++;
-        if (header_->freeHead == nNode_) { header_->freeHead = 0; }
+        const auto iNode = header_->freeHead;
```
UESTC-AHao, line 420: Is it necessary to prioritize allocation from the reserved reservedFreeHead?

mag1c-h: No, the reserved slots are only for special cases, not for priority use.

ucm/store/cache/cc/trans_buffer.cc
mag1c-h: No, once a slot is referenced, the next slot is immediately sought. Each slot has a corresponding mutex lock, and the locking process yields the CPU.

ygwpz (approved):

PR #901, opened 2026-04-08
[Fix] Duplicate metrics registration

flesher0813:
## Purpose
Fix the error register metrics twice when scheduler and worker share the same process

## Modifications 
Modify ucm/observability.py to ensure one process can only register once:  move metrics_mapping to be global variable

## Test
Start server with the following command, start server success and get metrics success 

> vllm serve /models/Qwen3-1.7B --served-model-name Qwen3-1.7B --block-size 128 --tensor-parallel-size 1 --data-parallel-size 1 --gpu-memory-utilization 0.87 --trust-remote-code --port 7850 --kv-transfer-config '{"kv_connector":"UCMConnector","kv_connector_module_path":"ucm.integration.vllm.ucm_connector","kv_role":"kv_both","kv_connector_extra_config":{"use_layerwise":"true","UCM_CONFIG_FILE":"/vllm-workspace/unified-cache-management/examples/ucm_config_example.yaml"}}'

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -262,7 +262,7 @@ class UCMDirectConnector(KVConnectorBase_V1):
         self.metrics_config = self.launch_config.get("metrics_config_path", "")
         if self.metrics_config:
             worker_id = (
-                get_world_group().rank
+                f"{self.engine_id}_{get_world_group().rank}"
                 if role == KVConnectorRole.WORKER
                 else self.engine_id
             )
--- a/ucm/observability.py
+++ b/ucm/observability.py
@@ -36,6 +36,10 @@ from ucm.shared.metrics import ucmmetrics
 
 logger = init_logger(__name__)
 
+# Store metric mapping: metric_name -> Union[Counter, Gauge, Histogram]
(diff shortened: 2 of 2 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -262,7 +262,7 @@ def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
         self.metrics_config = self.launch_config.get("metrics_config_path", "")
         if self.metrics_config:
             worker_id = (
-                get_world_group().rank
+                f"{self.engine_id}_{get_world_group().rank}"
```
Wwwzff, line 265: wonder if it's unique under DP circumstances

flesher0813: Yes, it would be unique id for DP, code like this
`def get_engine_name_with_dp(self, engine_name, dp_rank): 
        return f"{engine_name}_dp{dp_rank}"`

ygwpz (approved):

ucm/integration/vllm/ucm_connector.py
harrisonyhq: In case of P/D separation, will prefill worker and decode worker share a same engine id within a DP group?

PR #902, opened 2026-04-08
[CI] refactor dockerfile

dante159753:
## Purpose
make the dockerfile support production usage, it can install from source or from built package.

## Modifications 
add build script, build package first in dockerfile, then install from the built package

## Test
validate through PR gate, also add mindie and sglang compilation in PR gate pipeline


docker/Dockerfile.ucm-sglang-cuda-v0.5.5
```diff
@@ -0,0 +1,33 @@
+# Set to other image if needed
+ARG IMAGE_SOURCE="lmsysorg"
+ARG IMAGE_NAME_VERSION="sglang:v0.5.5.post3"
```
yuanzhg078, line 3: Maybe should update version to latest sglang release.

dante159753: not sure if it works with latest sglang, maybe that's a seperate work and can be analyzed later

Wwwzff (dismissed):

ygwpz (approved):

PR #903, opened 2026-04-08
[Bugfix] Fix raise with string literals in UCMBlendConnector

SuperMarioYL:
## Purpose

Fix three `raise "string"` statements in `UCMBlendConnector` that produce `TypeError` at runtime instead of the intended error messages.

## Modifications

In Python 3, `raise "string"` does not raise an exception with that message — it raises a `TypeError: exceptions must derive from BaseException`. This means callers never see the actual diagnostic text.

| Line | Before | After |
|------|--------|-------|
| 136 | `raise "UCMBlendConnector init failed, ..."` | `raise ValueError("UCMBlendConnector init failed, ...")` |
| 330 | `raise "Please call setup model first."` | `raise RuntimeError("Please call setup model first.")` |
| 339 | `raise "get cos_sin_cache from model failed! ..."` | `raise RuntimeError("get cos_sin_cache from model failed! ...")` |

- `ValueError` for the config validation (invalid configuration input)
- `RuntimeError` for the precondition and model compatibility checks

## Test

- Verified `black --check` and `isort --check` pass on the modified file.
- No new runtime behavior — only the exception type changes from `TypeError` to the correct type with the intended message.

### DIFF
--- a/ucm/integration/vllm/blend_connector.py
+++ b/ucm/integration/vllm/blend_connector.py
@@ -133,7 +133,7 @@ class UCMBlendConnector(UCMDirectConnector):
             self.enable_blend = True
             self.chunk_end_token_id = blend_config["chunk_end_token_id"]
         else:
-            raise "UCMBlendConnector init failed, please check your config"
+            raise ValueError("UCMBlendConnector init failed, please check your config")
 
         self.requests_blend_meta: dict[str, BlendRequestMeta] = {}
         self.cos_sin_cache: torch.Tensor = None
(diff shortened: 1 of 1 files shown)


SuperMarioYL:
The `test-e2e-sparse-offline-gpu` CI failure appears to be an infrastructure issue on the self-hosted runner — the log shows a pip ReadTimeoutError connecting to files.pythonhosted.org and a missing Python3 interpreter, both of which are unrelated to the code changes in this PR. Could a maintainer trigger a CI re-run when the runner is available?

SuperMarioYL:
Gentle ping — this has been open since early April with no review yet.

It's a small bug fix: `UCMBlendConnector` had `raise` statements on bare string literals, which raises `TypeError: exceptions must derive from BaseException` instead of the intended error.

The two red checks are `test-e2e-sparse-gpu` / `test-e2e-sparse-offline-gpu` on the self-hosted runner, and they last ran in April against a base that has since moved a long way. As I noted at the time, the logs showed a pip `ReadTimeoutError` and a missing Python 3 interpreter — infrastructure, not this diff. Happy to rebase onto current `develop` so they can be re-run against something fresh, if a maintainer can kick them off.

PR #905, opened 2026-04-09
[Bugfix] Fix PrometheusStatsLogger crash on re-instantiation and non-daemon thread

SuperMarioYL:
## Purpose

Fix two functional bugs in `PrometheusStatsLogger` that cause crashes in multi-engine scenarios and prevent clean process shutdown.

### Bug 1: `AttributeError` on re-instantiation

When `_metric_mappings` is already populated (e.g. multi-engine deployment or hot-reload), `__init__` returns early at line 71 **before** initializing any instance attributes (`is_running`, `thread`, `config`, `log_interval`, etc.).

Any subsequent call to `shutdown()` or the `__del__` destructor raises `AttributeError: 'PrometheusStatsLogger' object has no attribute 'is_running'`.

**Fix:** Move `self.config`, `self.log_interval`, `self.is_running`, and `self.thread` initialization **before** the early-return guard. The second instance becomes a safe no-op that can be cleanly shut down.

### Bug 2: Non-daemon `update_stats_loop` thread blocks process exit

The background thread is created without `daemon=True`. If `shutdown()` is never explicitly called (e.g. crash, unclean teardown, `__del__` not invoked by GC), this thread keeps the Python process alive indefinitely.

**Fix:** Set `daemon=True` on the thread. Also guard `shutdown()` with `if self.thread is not None` to handle the re-instantiation case safely.

## Test

- `black --check` and `isort --check --profile=black` pass on the modified file.
- Both bugs are deterministic: Bug 1 triggers on any second instantiation, Bug 2 triggers on any unclean shutdown.

### DIFF
--- a/test/test_ucm_connector_metrics.py
+++ b/test/test_ucm_connector_metrics.py
@@ -90,7 +90,7 @@ class FakeHistogram(FakeMetric):
 
 
 class FakeThread:
-    def __init__(self, target):
+    def __init__(self, target, daemon=None):
         self.target = target
         self.started = False
         self.joined = False
@@ -1312,7 +1312,7 @@ def test_multiproc_logger_respects_consumer_switch():
     )
     logger = observability.PrometheusStatsLogger("model-a", "worker-0", "unused.yaml")
 
-    assert not hasattr(logger, "thread")
+    assert logger.thread is None
     assert FakeCounter.created == {}
 
 
--- a/ucm/observability.py
+++ b/ucm/observability.py
@@ -55,6 +55,12 @@ class PrometheusStatsLogger:
         Load metrics config from YAML file (config_path),
         register metrics using prometheus_client, and start a thread to get updated metrics.
         """
+        # Always initialize the worker thread attribute so that shutdown()
+        # and __del__() are safe even when metric registration is skipped
+        # via one of the early-returns below.
+        self.thread = None
+        self.is_running = False
+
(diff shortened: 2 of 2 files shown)


SuperMarioYL:
The `test-e2e-pc-gpu`, `test-e2e-sparse-gpu`, and `test-e2e-sparse-offline-gpu` CI failures show network proxy errors and a missing Python3 interpreter on the self-hosted runner — this appears to be an infrastructure issue rather than a code regression. Could a maintainer trigger a CI re-run when the runner environment is stable?

SuperMarioYL:
Rebased onto `develop` (this PR's base) and re-authored the fix against the refactored `__init__` (upstream #901/#995 added two more early-return paths). In ca9c6269:

- `__init__` now sets `self.thread = None` and `self.is_running = False` **before** any early-return, so `shutdown()` / `__del__()` are safe regardless of which early-return is hit.
- the worker thread is created with `daemon=True`, so it won't block process exit if `shutdown()` isn't called explicitly.
- `shutdown()` guards `self.thread.join()` with `if self.thread is not None:`.

Verified all three bugs are still live on current `develop` (the early-returns skip attribute setup; `shutdown()` calls `self.thread.join()` unguarded; the thread has no `daemon=True`). The test's `FakeThread` now accepts the `daemon` kwarg and the early-return assertion was updated to `logger.thread is None`. Happy to address any further feedback.

PR #907, opened 2026-04-09
[opt] Support full-graph replay for GSA-enabled sparse attention

wangwenxin0312:
## Purpose
This PR supports GSA sparse attention to CUDA full-graph mode and adds a runtime fallback for unsupported MLA branches.
## Modifications 
- Graph Construction
  - Add build_dummy_sparse_meta in dummy_run to construct GSA-related metadata during graph capture.
  - Ensure sparse paths are included in the captured full graph.
  - Capture additional GSA-specific graphs via maybe_capture_extra_cudagraphs after initial graph capture.

- Runtime Sparse Meta Build
  - In execute_model, build runtime sparse metadata before dispatch.

- GSA-aware Batch Descriptor Dispatch
  - Replace default BatchDescriptor with GsaBatchDescriptor when GSA is enabled.

- Runtime Mode Adjustment
  - Add GSA-specific runtime mode adjustment after dispatch.
  - For MLA + gsa_enabled + pc_hit + decode_only:
    - force cudagraph_runtime_mode = CUDAGraphMode.NONE
    - fall back to default BatchDescriptor
  - This avoids replaying graphs with mismatched execution paths.

- Preserve GSA descriptor dispatch for all other compatible cases.
## Test
image


ucm/sparse/gsa_on_device/gsa_on_device.py
```diff
@@ -994,6 +1025,47 @@ def attention_finished(
...
+        self.gsa_enabled = False
+        self.has_decode = False
+        self.decode_only = False
+        self.has_pc_hit = False
+
+    def build_dummy_sparse_meta(self, attn_metadata) -> bool:
+        """Set up all internal GSA state required before a GSA-on graph
+        capture _dummy_run.
+        """
+        from ucm.sparse.gsa_on_device.hamming_topk import update_seq_lens
```
Infinite666: All of the imports of `update_seq_lens` can be moved to the very front of the file.

ucm/sparse/gsa_on_device/gsa_on_device.py
```diff
@@ -994,6 +1025,47 @@ def attention_finished(
...
+        self.num_reqs = seq_lens.size(0)
+
+        # ── flags checked by attention_begin ──────────────────────────────
+        self.gsa_enabled = True
+        self.has_decode = True
+        self.decode_only = True
+        self.has_pc_hit = False
+
+        # ── ori (full-table) buffers: use dummy large seq_lens ────────────
+        if not self.gsa_enabled:
```
Infinite666: What's the usage of this branch since `self.gsa_enabled` was just set to True?

ucm/sparse/gsa_on_device/gsa_on_device.py
```diff
@@ -1236,25 +1293,57 @@ def get_seq_lens(self, attn_metadata):
...
-            attn_metadata.seq_lens,
-            topk_token=self.hash_topk_tokens,
-            block_size=self.block_size,
-        )
+            _topk = update_seq_lens(
+                attn_metadata.seq_lens,
+                topk_token=self.hash_topk_tokens,
+                block_size=self.block_size,
+            )
+            self.cg_buf_topk_seq_lens_qwen[: self.num_reqs].copy_(_topk)
```
Infinite666, line 1328: Why only GQA's seq_lens tensor needs buffer but MLA doesn't?

ucm/sparse/gsa_on_device/gsa_on_device.py
```diff
@@ -1379,6 +1468,128 @@ def build_sparse_meta(
...
+            batch_descriptor.num_tokens,
+            batch_descriptor.uniform_decode,
+            gsa_enabled=True,
+        )
+        full_keys = cudagraph_dispatcher.cudagraph_keys[CUDAGraphMode.FULL]
+        pw_keys = cudagraph_dispatcher.cudagraph_keys[CUDAGraphMode.PIECEWISE]
+
+        if (
+            gsa_bd in full_keys
+            or gsa_bd.non_uniform in full_keys
```
Infinite666: Why does non uniform decode can be in FULL?

ucm/sparse/gsa_on_device/gsa_on_device.py
```diff
@@ -1379,6 +1468,128 @@ def build_sparse_meta(
...
+    def resolve_batch_descriptor(
+        self,
+        batch_descriptor: BatchDescriptor,
+        cudagraph_dispatcher: Any,
+    ) -> BatchDescriptor:
+        """Return a GSA-specific batch descriptor when a matching graph exists."""
+        if not self.gsa_enabled:
+            return batch_descriptor
+
+        if self.is_mla and self.has_pc_hit and self.decode_only:
```
Infinite666: This is to prevent the scenario where the length is 1 but it is actually a prefill right? What about the more general scenarios like `self.has_pc_hit` is True or `self.decode_only` is False?

ucm/sparse/gsa_on_device/gsa_on_device.py
wangwenxin0312: Thank you for the suggestion. Have moved.

ucm/sparse/gsa_on_device/gsa_on_device.py
wangwenxin0312: Thank you for the suggestion. Have removed.

ucm/sparse/gsa_on_device/gsa_on_device.py
wangwenxin0312: The implementation of the MLA is located in `maybe_init_cudagraph_buffers_for_topk`

ucm/sparse/gsa_on_device/gsa_on_device.py
wangwenxin0312: The condition gsa_bd.non_uniform in full_keys is unnecessary and has been removed.
FULL graphs are only captured when uniform_decode=True.

ucm/sparse/gsa_on_device/gsa_on_device.py
wangwenxin0312: For mixed prefill-decode batches, or for MLA decode with has_pc_hit=True or pure_prefill, we keep using the original batch_descriptor so the batch follows the default graph dispatch path.

Infinite666: OK, so can it just be simply written like below?
```
if self.has_pc_hit:
    return batch_descriptor
```

wangwenxin0312: Not sufficient. Even when `has_pc_hit=False`, we still need to fall back to the original batch_descriptor for pure prefill batches or mixed prefill-decode batches.

Infinite666 (approved):

PR #917, opened 2026-04-16
[Feat] Support recording hit rate

flesher0813:
## Purpose
Support logging requests' hit rate with no real dumping/loading.
Combined with the existing **Fake Store** — which only stores metadata in /dev/shm without actual KV load/dump operations — add a new **UCMLiteConnector** in ucm_connector.py to support per-request hit ratio and overall total hit ratio logging.

## Modifications 
- Add **UCMLiteConnector** in ucm_connector.py, and refactor the redundant functions _get_num_new_match_tokens_ and __generate_dispatch_meta_ in **UCMCPConnector** into the parent class to simplify the codebase.
- Add a new switch **use_lite** in **ucm_config_example.yaml** to control using of **UCMLiteConnector**.

## How to use it
Only need to set use_lite=true in **ucm_config_example.yaml** to use it, default using 10GB to set buffer_number.

## Test
Tested with vllm benchmark and not enable prefix caching
model | tp | input len | output len | prompts | use lite | concurrency | hit rate | ttft | tpot | e2el
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
QwQ-32B | 4 | 4000 | 100 | 100 | yes | 100 | 0 | 6989.23 | 81.36 | 15044.29
QwQ-32B | 4 | 4000 | 100 | 100 | yes | 100 | 0 | 7186.33 | 81.39 | 15244
QwQ-32B | 4 | 4000 | 100 | 100 | yes | 100 | 0 | 7274.58 | 81.36 | 15328.94
QwQ-32B | 4 | 4000 | 100 | 100 | yes | 100 | 1 | 7148.75 | 81.19 | 15186.64
QwQ-32B | 4 | 4000 | 100 | 100 | yes | 100 | 1 | 7159.6 | 81.28 | 15206.64
QwQ-32B | 4 | 4000 | 100 | 100 | yes | 100 | 1 | 7164.8 | 81.31 | 15214.25
QwQ-32B | 4 | 4000 | 100 | 100 | no | 100 | 0 | 7019.12 | 81.41 | 15078.46
QwQ-32B | 4 | 4000 | 100 | 100 | no | 100 | 0 | 7044.49 | 81.37 | 15100.45
QwQ-32B | 4 | 4000 | 100 | 100 | no | 100 | 0 | 7181.16 | 81.34 | 15234.13

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -279,6 +279,9 @@ class UCMDirectConnector(KVConnectorBase_V1):
 
         # invlalid block ids due to load errors
         self._invalid_block_ids: set[int] = set()
+        self.cp_world_size = 1
+        self.hash_block_size = self.block_size
+        self.block_size *= self.cp_world_size
 
     def generate_hash(
         self, block_size: int, token_ids: List[int], parent_block_hash_value: bytes
@@ -400,14 +403,15 @@ class UCMDirectConnector(KVConnectorBase_V1):
         hbm_hit_block_num = num_computed_tokens // self.block_size
 
         ucm_block_ids = self.generate_hash(
-            self.block_size, request.all_token_ids, self._seed
+            self.hash_block_size, request.all_token_ids, self._seed
         )
 
-        external_block_ids = ucm_block_ids[hbm_hit_block_num:]
+        external_block_ids = ucm_block_ids[hbm_hit_block_num * self.cp_world_size :]
         if not external_block_ids:
             return 0, False
         try:
             external_hit_blocks = self.store.lookup_on_prefix(external_block_ids) + 1
+            external_hit_blocks //= self.cp_world_size
         except RuntimeError as e:
             external_hit_blocks = 0
             logger.error(f"request {request.request_id} look up error. {e}")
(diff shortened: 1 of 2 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -1154,6 +1061,43 @@ def get_num_new_matched_tokens(
         return expect_hit_block_num * self.block_size, False
 
 
+class UCMLiteConnector(UCMDirectConnector):
```
ygwpz, line 1064: it seems lite connector does not need to inherit UCMDirectConnector because it only needs to implement get_num_new_matched_tokens() method.

flesher0813: Inheriting from UCMDirectConnector is to adapt to the Fake Store. For temporarily, this part will not be modified.

qyh111 (approved):

ucm/integration/vllm/ucm_connector.py
mag1c-h: Implementing only the get_num_new_matched_tokens method is not enough, because lite also needs to record metadata when saving_kvcache.

mag1c-h (approved):

PR #920, opened 2026-04-16
[Fix] Update KVCacheLayout to set row counts.

hmy98213:
## Purpose

Fix the `IndexError: list index out of range` issue when running `GLM-4.7-W8A8-floatmtp`.

## Modification

Use the number of unique layer IDs, `len(set(self.layer_name_to_id.values()))`, as the row count in KV cache layout construction instead of `self.local_num_hidden_layers`.

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -88,8 +88,10 @@ class KVCacheLayout:
         self._build_layout(kvcaches)
 
     def _build_layout(self, kvcaches):
-        raw_ptr_rows = [[] for _ in range(self.local_num_hidden_layers)]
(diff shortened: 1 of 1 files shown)


qyh111 (approved):

ucm/integration/vllm/ucm_connector.py
```diff
@@ -88,8 +88,10 @@ def __init__(
         self._build_layout(kvcaches)
 
     def _build_layout(self, kvcaches):
-        raw_ptr_rows = [[] for _ in range(self.local_num_hidden_layers)]
```
flesher0813, line 91: Did self.local_num_hidden_layers) useful in other functions? Can we remove it and its related variables?

flesher0813: Fix later

PR #923, opened 2026-04-17
[Feat] Integrate Mooncake into UCM By UcmMooncakeStoreV1

NaganooMei:
## Purpose
- Integrate `UcmMooncakeStoreV1` into UCM/vLLM KV transfer flow, including lookup/load/dump/wait/check path support.
- Wire Mooncake v1 connector registration in store factory and pass runtime config from `UCMConnector`.
- Add Mooncake Prefix Cache user guide `docs/source/user-guide/prefix-cache/mooncakestore.md`.
- Update Prefix Cache docs index to include `mooncakestore`.

## Modifications 
- `ucm/store/mooncakestore/mooncake_connector.py`
  - Refactor and integrate Mooncake v1 backend logic.
  - Support async task execution and task lifecycle checks.
  - Add worker-side NPU buffer registration for RDMA transfer.
- `ucm/integration/vllm/ucm_connector.py`
  - Pass Mooncake-required runtime fields (`device_id`, tensor size metadata, optional register buffers, etc.).
- `ucm/store/factory_v1.py`
  - Register `UcmMooncakeStoreV1` for connector factory creation.
- `docs/source/user-guide/prefix-cache/mooncakestore.md`
  - Add user guide for Mooncake Store v1 usage and launch instructions.
- `docs/source/user-guide/prefix-cache/index.md`
  - Add `mooncakestore` to toctree.
 
## Test
Validated on single-machine Atlas A2 with whole-save whole-load enabled and UCM + Mooncake backend.

### QwQ-32B (TP=2)

| Sequence Length | Concurrency | Hit Rate | Baseline | Read | Improvement |
|---|---:|---:|---:|---:|---:|
| 8k | 10 | 1.0 | 10819.43 | 1238.38 | 88.55% |
| 8k | 10 | 0.8 | 10819.43 | 3270.46 | 69.77% |

### DSV2-Lite (TP=2)

| Sequence Length | Concurrency | Hit Rate | Baseline | Read | Improvement |
|---|---:|---:|---:|---:|---:|
| 8k | 10 | 1.0 | 2911.66 | 859.18 | 70.49% |
| 8k | 10 | 0.8 | 2911.66 | 1356.10 | 53.43% |

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -335,6 +335,12 @@ class UCMDirectConnector(KVConnectorBase_V1):
             config["shard_size"] = kv_cache_layout.shard_size * self.blocks_per_chunk
             config["block_size"] = kv_cache_layout.block_size * self.blocks_per_chunk
             config["local_rank_size"] = self.tp_size if self.is_mla else 1
+            register_buffer_ptrs, register_buffer_sizes = (
+                self._build_register_buffer_regions()
+            )
+            if register_buffer_ptrs:
+                config["register_buffer_ptrs"] = register_buffer_ptrs
+                config["register_buffer_sizes"] = register_buffer_sizes
             if cpu_affinity_cores:
                 config["cpu_affinity_cores"] = list(cpu_affinity_cores)
         else:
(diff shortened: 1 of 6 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -349,6 +355,32 @@ def _create_store(
...
+        sizes: list[int] = []
+        for kv_layer in self.kv_caches.values():
+            for tensor in self._iter_register_buffer_tensors(kv_layer):
+                ptrs.append(int(tensor.data_ptr()))
+                sizes.append(int(tensor.numel() * tensor.element_size()))
+        logger.info(
+            "Mooncake worker register buffers prepared: regions=%s",
+            len(ptrs),
+        )
+        return ptrs, sizes
```
harrisonyhq, line 369: What's the kvcache here in this function? If some requests contains kv caches longer then current kv size, will this buffer still able to hold it?

ucm/integration/vllm/ucm_connector.py
```diff
@@ -353,6 +359,32 @@ def _create_store(
...
+        self, kv_layer: torch.Tensor | Tuple[torch.Tensor, ...]
+    ) -> list[torch.Tensor]:
+        if isinstance(kv_layer, torch.Tensor):
+            if kv_layer.dim() == 5:
+                return [kv_layer[0], kv_layer[1]]
+            if kv_layer.dim() == 3:
+                return [kv_layer]
+            raise ValueError(f"Unsupported kv cache tensor shape: {kv_layer.shape}")
+        if isinstance(kv_layer, tuple):
+            return list(kv_layer)
```
harrisonyhq, line 381: Since this function is called as iterator, you can refator this to return an iterator, rather than a full list, to avoid generate a very long list in case of long context.

ucm/integration/vllm/ucm_connector.py
NaganooMei: The kv_caches here are the full paged KV pool allocated and passed down by vLLM, rather than per-request KV cache. This function is only used to collect/register those backing buffers for Mooncake. If a request exceeds the capacity of the current KV pool, vLLM will handle that case in its own KV cache management path.

ucm/integration/vllm/ucm_connector.py
NaganooMei: the input of this helper is one layer’s KV cache tensor object, and the output is just the tensors expanded by KV dimension / tuple elements, depending on the layout. It is not related to request length.

PR #925, opened 2026-04-20
[Opt] Support setting a threshold for kv_cache storage

flesher0813:
## Purpose
Support setting threshold for kv_cache dump/load. When request's input tokens less than the threshold, ucm would not dump/load.
## Modifications 
- Add a new option **persist_token_threshold** in ucm_config_example.yaml
- Modify ucm_connector.py to support this feat
## Test
Tested with online service.

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -277,7 +277,11 @@ class UCMDirectConnector(KVConnectorBase_V1):
                 f"metrics_config_path: {self.metrics_config}, set worker_id: {worker_id}"
             )
 
-        # invlalid block ids due to load errors
+        self.persist_token_threshold = self.launch_config.get(
+            "persist_token_threshold", 0
+        )
+
+        # invalid block ids due to load errors
         self._invalid_block_ids: set[int] = set()
         self.cp_world_size = 1
         self.hash_block_size = self.block_size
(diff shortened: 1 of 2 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -402,6 +406,10 @@ def get_num_new_matched_tokens(
         assert num_computed_tokens % self.block_size == 0
         hbm_hit_block_num = num_computed_tokens // self.block_size
 
+        # Skip persistence if token count is below the threshold
+        if self.persist_token_threshold > request.num_tokens:
```
qyh111, line 427: Maybe we should add a log here.

flesher0813: Fixed, add a log

harrisonyhq:
Is this compatible with chunk prefill？

ucm/integration/vllm/ucm_connector.py
```diff
@@ -402,6 +406,14 @@ def get_num_new_matched_tokens(
...
+            logger.info_once(
+                f"Skip persistence: req {request.request_id}, "
+                f"input tokens ({request.num_tokens}) < threshold ({self.persist_token_threshold})."
+            )
+            return 0, False
```
mag1c-h, line 432: The hit rate was set to 0 here, thus achieving the goal of skipping the load. So, what strategy was used to skip the dump?

flesher0813:
> Is this compatible with chunk prefill？

Yes, as the judgment relies on the total prompt tokens of the original request. Chunk processing will not lead to incorrect threshold checks.

ucm/integration/vllm/ucm_connector.py
flesher0813: Dump operations depend on the generation of self.requests_meta. Since this logic is skipped, the corresponding dump process will not be executed.

qyh111 (dismissed):

harrisonyhq (dismissed):

qyh111 (approved):

harrisonyhq (approved):

PR #926, opened 2026-04-20
[Fix] Not print hit rate when using 1tp

flesher0813:
## Purpose
Fix the bug that not print external hit rate when using 1tp
## Modifications 
Modify ucm_connector.py
## Test
Tested with online server using command like below, print external hit rate correctly.
```export ASCEND_RT_VISIBLE_DEVICES=1
vllm serve /models/Qwen3-0.6B \
--max-model-len 20000 \
--tensor-parallel-size 1 \
--gpu_memory_utilization 0.87 \
--block_size 128 \
--trust-remote-code \
--host x.x.x.x \
--port xxxx \
--no-enable-prefix-caching \
--kv-transfer-config \
'{
    "kv_connector": "UCMConnector",
    "kv_connector_module_path": "ucm.integration.vllm.ucm_connector",
    "kv_role": "kv_both",
    "kv_connector_extra_config": {"UCM_CONFIG_FILE": "/work-space/ucm_example_config.yaml"}
}'```

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -1096,6 +1096,7 @@ class UCMLiteConnector(UCMDirectConnector):
             ],
             "enable_record_traces": enable_record_traces,
             "persist_token_threshold": persist_token_threshold,
+            "use_lite": True,
         }
         super().__init__(vllm_config, role)
         self.total_block_nums = 0

### CODE REVIEW

File: ucm/integration/vllm/ucm_connector.py [hunk 1]
```diff
@@ -1096,6 +1096,7 @@ def __init__(self, vllm_config, role):
             ],
             "enable_record_traces": enable_record_traces,
             "persist_token_threshold": persist_token_threshold,
+            "use_lite": True,
```
qyh111, line 1099: In which scenario do we need this?  `UCMDirectConnector` doesn't  need `use_lite`

flesher0813: When using 1 tp to start server, scheduler and worker would share the same process. Worker will init first and change vllm_config, it would cause scheduler create UCMDirectConnector but not UCMLiteConnector.

qyh111 (approved):

harrisonyhq (approved):

PR #927, opened 2026-04-20
[Fix] Fix dump err when 0 < len(load_blocks) < fully hit

flesher0813:
## Purpose
Fix the dump failure issue that occurs when len(load_blocks) < fully_hit in the external hit scenario with **use_lite=true**.The failure is caused by len(ucm_block_ids) < len(vllm_block_ids) during wait_for_save.This discrepancy arises because, in **_generate_dispatch_meta**, the **new_tokens** parameter still includes external hits, even though req_meta.token_processed has already accounted for those tokens.

## Modifications 
- Modify func **get_num_new_matched_tokens** of UCMLiteConnector: using hbm_hit_tokens as request_meta.token_processed, as it needs to return 0 hit in external storage.
- Modify func **_generate_dispatch_meta** of UCMLiteConnector: only generating dump block infos when req_meta.token_processed + new_tokens >= total_hit_tokens

## Test
Tested with online llmperf pipeline.

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -1116,8 +1116,18 @@ class UCMLiteConnector(UCMDirectConnector):
             external_hit_blocks = (
                 request_meta.total_hit_block_num - request_meta.hbm_hit_block_num
             )
+            need_dump_blks = request_meta.ucm_block_ids[
+                request_meta.total_hit_block_num :
+            ]
+            shard_indexs = [0] * len(need_dump_blks)
+            total_ptrs = [[0]] * len(need_dump_blks)
+            try:
+                task = self.store.dump_data(need_dump_blks, shard_indexs, total_ptrs)
+                self.store.wait(task)
+            except RuntimeError as e:
+                logger.error(f"request {request.request_id} wait dump task error. {e}")
+            self.requests_meta[request.request_id] = RequestMeta()
 
-            request_meta.total_hit_block_num = request_meta.hbm_hit_block_num
         self.total_hit_block_nums += external_hit_blocks
 
         logger.info(


ucm/integration/vllm/ucm_connector.py
```diff
@@ -1126,6 +1127,39 @@ def get_num_new_matched_tokens(self, request, num_computed_tokens):
...
+                dump_ucm_block_ids = ucm_block_ids[
+                    start_idx * self.cp_world_size : end_idx * self.cp_world_size
+                ]
+                dump_vllm_block_ids = req_meta.vllm_block_ids[start_idx:end_idx]
+            req_meta.token_processed += new_tokens
+
+        return RequestDispatchMeta(
+            (load_ucm_block_ids, load_vllm_block_ids),
+            (dump_ucm_block_ids, dump_vllm_block_ids),
+        )
```
qyh111: I wonder if we can use Fake Store in scheduler，then we don't need to do anything in worker

flesher0813: Yes, changed to this method

ucm/integration/vllm/ucm_connector.py
```diff
@@ -1116,8 +1116,18 @@ def get_num_new_matched_tokens(self, request, num_computed_tokens):
...
+                request_meta.total_hit_block_num :
+            ]
+            shard_indexs = [0] * len(need_dump_blks)
+            total_ptrs = [[0]] * len(need_dump_blks)
+            try:
+                task = self.store.dump_data(need_dump_blks, shard_indexs, total_ptrs)
+                self.store.wait(task)
+            except RuntimeError as e:
+                logger.error(f"request {request.request_id} wait dump task error. {e}")
+            self.requests_meta[request.request_id] = RequestMeta()
```
qyh111, line 1129: 这里是不是直接从self.requests_meta pop出去是不是更好，不需要在维护这个请求了

qyh111 (approved):

ucm/integration/vllm/ucm_connector.py
flesher0813: 这里考虑到可以同时trace+lite，trace打印依赖request.request_id in self.requests_meta的判断，所以这里只选择清空

PR #933, opened 2026-04-24
[Feat] Add per-stage pipeline store and layerwise overlap metrics

dante159753:
## Purpose
Adds observability needed to diagnose pipeline store (Cache|Posix) per-tier performance and to verify that UCMLayerWiseConnector's load/forward/save overlap actually hides backend latency.

## Modifications 

Pipeline Store (C++ side):
  - Cache: per-task load/dump duration + bandwidth, queue wait, dispatch, backend-wait, H2D/D2H, backend-submit durations; lookup hit/miss counters and instantaneous hit rate gauge.
  - Posix: per-IO S2H/H2S duration + bandwidth, queue wait, failure counters.

Layerwise Connector (Python side):
  - wait_blocking_ms: primary signal for overlap health (near 0 = perfect overlap; tracks load_duration = degenerated to serial).
  - inter_wait_interval_ms, next_layer_submit_ms, first_layer_submit_ms, save_submit_ms, save_per_layer_wait_ms, save_tail_total_ms, stalled_layers_total.

Infrastructure:
  - Change metrics library from STATIC to SHARED so cachestore.so, posixstore.so, and ucmmetrics.so share one Metrics singleton. With a STATIC metrics the function-local GetInstance() produced a separate instance in each .so and all C++ UpdateStats() calls from stores were silently dropped.
  - Set INSTALL_RPATH=\$ORIGIN/../../shared/metrics on cachestore.so and posixstore.so; \$ORIGIN on ucmmetrics.so.


ucm/store/cache/cc/load_queue.cc
```diff
@@ -143,6 +154,8 @@ void LoadQueue::TransferOneTask(CopyStream& stream, ShardTask&& task)
         }
         if (!task.waiter) {
             holder_.push_back(std::move(task));
+            UC::Metrics::UpdateStats("cache_load_backend_wait_duration_ms",
```
ygwpz, line 157: **Bug:** `cache_load_backend_wait_duration_ms` is recorded twice in this function.

- First time here (line 157-158) in the `if (!task.waiter)` branch
- Second time at line 168-169 in the normal flow after `stream.Synchronize()`

This causes incorrect metrics data because the same metric is updated from two different execution paths. Should only record once based on the actual path taken.

ucm/store/cache/cc/load_queue.cc
```diff
@@ -151,6 +164,10 @@ void LoadQueue::TransferOneTask(CopyStream& stream, ShardTask&& task)
             UC_ERROR("Failed({}) to sync on stream for task({}).", s, task.taskHandle);
             break;
         }
+        auto tpEnd = NowTime::Now();
+        UC::Metrics::UpdateStats("cache_load_backend_wait_duration_ms",
```
ygwpz, line 168: **Duplicate recording:** This is the second place where `cache_load_backend_wait_duration_ms` is recorded. See my comment at line 157-158 for the first occurrence.

Both paths record the same metric, which will cause data corruption in Prometheus metrics. Consider removing one or restructuring the logic to only record in the appropriate path.

ucm/store/cache/cc/trans_manager.h
```diff
@@ -55,11 +56,23 @@ class TransManager : public Detail::TaskWrapper<TransTask, Detail::TaskHandle> {
         const auto num = t->desc.size();
         const auto size = shardSize_ * num;
         const auto tp = w->startTp;
+        const auto isLoad = t->type == TransTask::Type::LOAD;
         UC_DEBUG("Cache task({},{},{},{}) dispatching.", id, brief, num, size);
-        w->SetEpilog([id, brief = std::move(brief), num, size, tp] {
+        w->SetEpilog([id, brief = std::move(brief), num, size, tp, isLoad] {
             auto cost = NowTime::Now() - tp;
```
ygwpz, line 62: **Potential edge case:** When `cost` is very small (e.g., sub-millisecond), `bwGbps` could become extremely large due to division by a tiny number. Consider adding a minimum threshold for `cost` before calculating bandwidth to avoid unrealistic spikes in metrics.

Example: `if (cost > 1e-6) { bwGbps = ... } else { bwGbps = 0.0; }`

ucm/store/posix/cc/trans_queue.cc
```diff
@@ -94,13 +95,23 @@ void TransQueue::LoadWorker(IoUnit& ios)
     if (ios.firstIo) {
         auto wait = NowTime::Now() - ios.waiter->startTp;
         UC_DEBUG("Posix load task({}) start running, wait {:.3f}ms.", ios.owner, wait * 1e3);
+        UC::Metrics::UpdateStats("posix_load_wait_duration_ms", wait * 1e3);
     }
     if (failureSet_->Contains(ios.owner)) {
         ios.waiter->Done();
         return;
     }
+    auto tp = NowTime::Now();
```
ygwpz, line 104: **Same edge case as trans_manager.h:** When `cost` is very small, `bwGbps` could become extremely large. Consider adding a minimum threshold check before calculating bandwidth to avoid unrealistic spikes.

examples/metrics/metrics_configs.yaml
```diff
@@ -4,23 +4,53 @@ log_interval: 5  # Interval in seconds for logging metrics
...
 
 # Gauge metrics configuration
-# gauge:
-#   - name: "lookup_hit_rate"
-#     documentation: "Hit rate of ucm lookup requests since last log"
-#     multiprocess_mode: "livemostrecent"
+gauge:
+  - name: "cache_lookup_hit_rate"
+    documentation: "Instantaneous Cache stage hit rate from the most recent lookup call"
+    multiprocess_mode: "livemostrecent"
```
ygwpz, line 84: **Missing metric definition:** `layerwise_first_layer_requests` is used in the code (ucm_connector.py:849) but not defined in the histogram configuration here. Should this be a gauge or counter? Please add the corresponding definition.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -852,14 +888,41 @@ def wait_for_layer_load(self, layer_name: str) -> None:
...
+        self._layerwise_prev_wait_end = wait_end
+
+        if self.metrics_config:
+            blocking_ms = (wait_end - wait_start) * 1000
+            # Key metric: near 0 means overlap is perfect; large values mean
+            # loads are the bottleneck.
+            ucmmetrics.update_stats(
+                {
+                    "layerwise_wait_blocking_ms": blocking_ms,
+                    "layerwise_wait_tasks_count": float(n_tasks),
```
ygwpz, line 897: **Missing metric definition:** `layerwise_wait_tasks_count` is used here but not defined in `metrics_configs.yaml`. Please add the corresponding histogram or gauge definition.

ygwpz (changes requested):

ucm/store/cache/cc/load_queue.cc
ygwpz, line 157: This issue is still present. The metric `cache_load_backend_wait_duration_ms` is recorded twice:
- Here at line 157-158 in the early return path
- Again at line 168-169 in the normal path

This causes duplicate metric recording depending on execution path. Consider removing the recording from the early return path, or restructuring to ensure single recording.

ucm/store/cache/cc/trans_manager.h
```diff
@@ -55,11 +56,23 @@ class TransManager : public Detail::TaskWrapper<TransTask, Detail::TaskHandle> {
...
+            auto costMs = cost * 1e3;
+            auto bwGbps = cost > 0 ? static_cast<double>(size) / cost / 1e9 : 0.0;
```
ygwpz, line 64: The edge case suggestion wasn't fully addressed. Current check `cost > 0` still allows division by extremely small numbers (e.g., microseconds), causing unrealistic bandwidth spikes. Consider adding a minimum threshold like `cost > 1e-6` or `cost > 0.001` to prevent unrealistic values.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -827,7 +835,20 @@ def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
             self.request_data.append((request_id, ucm_block_ids, total_ptrs))
 
         if self.need_load:
```
ygwpz, line 833: Multiple `time.perf_counter()` calls are added throughout the critical path. While `perf_counter` is relatively fast, consider batching metrics updates to reduce overhead in high-throughput scenarios. For example, `wait_for_layer_load` now has 4 timing calls per invocation.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -852,14 +888,41 @@ def wait_for_layer_load(self, layer_name: str) -> None:
...
+            blocking_ms = (wait_end - wait_start) * 1000
+            # Key metric: near 0 means overlap is perfect; large values mean
+            # loads are the bottleneck.
+            ucmmetrics.update_stats(
+                {
+                    "layerwise_wait_blocking_ms": blocking_ms,
+                    "layerwise_wait_tasks_count": float(n_tasks),
+                }
+            )
+            if n_tasks > 0 and blocking_ms > 1.0:
```
ygwpz, line 900: The threshold `blocking_ms > 1.0` for counting stalled layers might be too low for high-throughput scenarios. Even small scheduler jitter could trigger this counter frequently. Consider adjusting based on actual workload characteristics.

PR #945, opened 2026-04-30
[Feat] Mooncake & Mooncake|Posix

UESTC-AHao:
## Purpose
UCM + Mooncake & Mooncake|Posix (Ascend 0.18.0)

## Modifications 
unified-cache-management\examples\ucm_mooncake_config.yaml

### DIFF
--- a/examples/ucm_mooncake_config.yaml
+++ b/examples/ucm_mooncake_config.yaml
@@ -1,0 +1,38 @@
+# UCM + Mooncake Configuration Example
+#
+# Prerequisites:
+#   1. mooncake master_service running: mooncake_master --port 50088
+#   2. libmooncakestore.so built and installed
+#   3. NPU device available
+#
+# Usage:
+#   kv_connector_extra_config={"UCM_CONFIG_FILE": "/path/to/ucm_mooncake_config.yaml"}
+
+ucm_connectors:
+  - ucm_connector_name: "UcmPipelineStore"
+    ucm_connector_config:
+      store_pipeline: "Mooncake|Posix"
+      local_hostname: "127.0.0.1"
+      master_server_address: "127.0.0.1:50088"
+      metadata_server: "P2PHANDSHAKE"
+      protocol: "ascend"
+      global_segment_size: 32212254720
+      replica_num: 1
+      local_buffer_size: 1073741824
+      storage_backends: "/mnt/test"
+      io_direct: false
+      # device_name: ""                 # RDMA device, leave empty for ascend
+
+# prerequisiteHandle requires event sync to ensure NPU compute finishes
+# before Mooncake reads device memory for dump.
+enable_event_sync: true
+
+# Enable UCM metrics (optional)
+# metrics_config_path: "/workspace/unified-cache-management/examples/metrics/metrics_configs.yaml"
+
+# Sparse attention (optional)
+# ucm_sparse_config:
+#   GSAOnDevice: {}
+
+# Whether to use layerwise loading/saving (optional, default: True for UCMConnector)
+use_layerwise: true
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -382,6 +382,7 @@ class UCMDirectConnector(KVConnectorBase_V1):
         )
 
         self.store = self._create_store(self.kv_cache_layout, store_cores)
+        self._register_kv_cache_memory()
 
         if worker_cores:
             try:
@@ -392,6 +393,31 @@ class UCMDirectConnector(KVConnectorBase_V1):
 
         if self.device is None:
             raise RuntimeError(f"Unsupported device platform for UCMDirectConnector.")
+
+    def _register_kv_cache_memory(self):
+        for layer_name, kv_layer in self.kv_caches.items():
+            if isinstance(kv_layer, torch.Tensor):
+                if kv_layer.dim() == 5:
+                    num_blocks = kv_layer.shape[1]
+                    block_size = kv_layer[0].shape[1:].numel() * kv_layer.element_size()
+                    total_size = num_blocks * block_size
+                    self.store.register_memory(kv_layer[0].data_ptr(), total_size)
+                    self.store.register_memory(kv_layer[1].data_ptr(), total_size)
+                elif kv_layer.dim() == 3:
+                    num_blocks = kv_layer.shape[0]
+                    total_size = kv_layer.numel() * kv_layer.element_size()
+                    self.store.register_memory(kv_layer.data_ptr(), total_size)
+                else:
+                    raise ValueError(
+                        f"Unsupported kv cache tensor shape: {kv_layer.shape}"
+                    )
+            elif isinstance(kv_layer, Tuple):
+                for tensor in kv_layer:
+                    total_size = tensor.numel() * tensor.element_size()
+                    self.store.register_memory(tensor.data_ptr(), total_size)
+            else:
+                raise TypeError(f"Unsupported kv cache type: {type(kv_layer)}")
+        logger.info(f"Registered {len(self.kv_caches)} layers' kv cache memory")
 
     def get_num_new_matched_tokens(
         self,
--- a/ucm/store/mooncakestore/CMakeLists.txt
+++ b/ucm/store/mooncakestore/CMakeLists.txt
@@ -1,0 +1,50 @@
+find_path(ASCEND_ACL_INCLUDE_DIR NAMES acl/acl.h PATHS /usr/local/Ascend/ascend-toolkit/latest/include NO_DEFAULT_PATH)
+find_library(ASCEND_ACL_LIBRARY NAMES ascendcl PATHS /usr/local/Ascend/ascend-toolkit/latest/lib64 NO_DEFAULT_PATH)
+find_library(MOONCAKE_STORE_LIBRARY NAMES mooncake_store PATHS /vllm-workspace/Mooncake/mooncake-store/lib NO_DEFAULT_PATH)
+
+if(ASCEND_ACL_INCLUDE_DIR AND ASCEND_ACL_LIBRARY AND MOONCAKE_STORE_LIBRARY)
+    file(GLOB_RECURSE SOURCES "./cc/*.cc")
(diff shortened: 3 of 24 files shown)


examples/ucm_mooncake_config.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+ucm_connectors:
+  - ucm_connector_name: "UcmPipelineStore"
+    ucm_connector_config:
+      store_pipeline: "Mooncake|Posix"
+      local_hostname: "127.0.0.1"
+      master_server_address: "127.0.0.1:50088"
+      metadata_server: "P2PHANDSHAKE"
+      protocol: "ascend"
+      global_segment_size: 32212254720
```
sumingZero, line 19: Should we explain the meaning of these parameters

ucm/store/pipeline/connector.py
```diff
@@ -157,6 +157,9 @@ def wait(self, task: Task) -> None:
     def check(self, task: Task) -> bool:
         return self.store_.Check(task.task_id)
 
+    def register_memory(self, base_addr: int, total_size: int) -> None:
+        self.store_.RegisterMemory(base_addr, total_size)
```
sumingZero, line 161: Should the return value of  RegisterMemory be validated

ucm/store/ucmstore_v1.h
```diff
@@ -147,6 +147,19 @@ class StoreV1 {
...
+     *
+     * This method should be called once during initialization to register
+     * the entire KV cache memory range. Subsequent Load/Dump operations
+     * will use addresses within this registered range.
+     *
+     * @param base_addr Base address of the memory region.
+     * @param total_size Total size of the memory region in bytes.
+     * @return Status::OK on success, error code on failure.
+     */
+    virtual Status RegisterMemory(void* base_addr, size_t total_size) = 0;
```
sumingZero, line 161: If RegisterMemory is Mooncake-specific operation, should it be handled internally within MooncakeStore rather than being added as a public interface in StoreV1 base class? The current design forces all other store implementations to provide empty stub methods, violating Interface Segregation Principle.

ucm/store/test/e2e/mooncake_store_test.py
```diff
@@ -0,0 +1,396 @@
...
+def create_mooncake_store(
+    device_id,
+    tensor_size,
+    tensors_per_block,
+    local_hostname,
+    master_address,
+    metadata_server,
+    protocol,
+    global_segment_size,
+    replica_num,
```
yuanzhg078, line 145: **create_mooncake_store()** is called with an extra argument worker_num, but its function signature does not define worker_num

ucm/store/mooncakestore/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+find_path(ASCEND_ACL_INCLUDE_DIR NAMES acl/acl.h PATHS /usr/local/Ascend/ascend-toolkit/latest/include NO_DEFAULT_PATH)
+find_library(ASCEND_ACL_LIBRARY NAMES ascendcl PATHS /usr/local/Ascend/ascend-toolkit/latest/lib64 NO_DEFAULT_PATH)
```
ygwpz, line 2: The path `/vllm-workspace/Mooncake/mooncake-store/lib` is hardcoded here. This won't work in our CI environment where Mooncake is installed at a different location. Could we use an environment variable like `MOONCAKE_STORE_ROOT` instead?

ucm/store/mooncakestore/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+find_library(MOONCAKE_STORE_LIBRARY NAMES mooncake_store PATHS /vllm-workspace/Mooncake/mooncake-store/lib NO_DEFAULT_PATH)
+
+if(ASCEND_ACL_INCLUDE_DIR AND ASCEND_ACL_LIBRARY AND MOONCAKE_STORE_LIBRARY)
+    file(GLOB_RECURSE SOURCES "./cc/*.cc")
+    add_library(mooncakestore SHARED ${SOURCES})
+
+    target_compile_features(mooncakestore PUBLIC cxx_std_20)
```
ygwpz, line 9: Same issue here with the include path. Would be nice if `MOONCAKE_STORE_INCLUDE_DIR` could be passed as a cmake option.

ucm/store/mooncakestore/cc/trans_manager.cc
```diff
@@ -0,0 +1,502 @@
...
+{
+    for (size_t i = 0; i < shards.size(); ++i) {
+        if (results[i] >= 0) { continue; }
+
+        auto& shard = shards[i];
+        size_t totalSize = 0;
+        for (auto sz : shard.sizes) { totalSize += sz; }
+        if (totalSize == 0) { continue; }
+
+        void* buf = bufPool.AcquireWithTimeout(std::chrono::milliseconds(30000));
```
ygwpz, line 208: The 30-second timeout here is hardcoded. In some scenarios we might want a longer timeout for large transfers. Could this be configurable via the Config struct?

ucm/store/mooncakestore/cc/mooncake_store.cc
```diff
@@ -0,0 +1,367 @@
...
+            return Status::Error("RPC Client::Create failed");
+        }
+        rpcClient_ = clientOpt.value();
+        UC_DEBUG("RPC Client setup ok (lookup only)");
+        return Status::OK();
+    }
+
+    std::vector<int> RpcBatchIsExist(const std::vector<std::string>& keys)
+    {
+        if (keys.empty()) { return std::vector<int>(keys.size(), -1); }
```
ygwpz, line 255: When `keys.empty()`, returning `std::vector<int>(keys.size(), -1)` returns an empty vector, which is correct. But the logic feels a bit odd - why check `keys.empty()` and then reference `keys.size()`? Could simplify to just `return {};`

ucm/store/mooncakestore/cc/mooncake_store.cc
```diff
@@ -0,0 +1,367 @@
...
+
+    void BuildShards(const Detail::TaskDesc& desc, TransTask& out)
+    {
+        out.shards.reserve(desc.size());
+        for (const auto& shard : desc) {
+            std::string key = BlockIdToKey(shard.owner) + "_" + std::to_string(shard.index);
+
+            if (shard.addrs.size() > config_.tensorSizeList.size()) {
+                UC_WARN(
+                    "BuildShards: key={} has {} addrs but tensorSizeList has only {}, truncating",
```
ygwpz, line 274: This warning is logged but the operation continues with truncated data. Should this be an error instead? Truncating tensor addresses seems like it could cause correctness issues.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -393,6 +394,31 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
+                    num_blocks = kv_layer.shape[1]
+                    block_size = kv_layer[0].shape[1:].numel() * kv_layer.element_size()
+                    total_size = num_blocks * block_size
+                    self.store.register_memory(kv_layer[0].data_ptr(), total_size)
+                    self.store.register_memory(kv_layer[1].data_ptr(), total_size)
+                elif kv_layer.dim() == 3:
+                    num_blocks = kv_layer.shape[0]
+                    total_size = kv_layer.numel() * kv_layer.element_size()
+                    self.store.register_memory(kv_layer.data_ptr(), total_size)
+                else:
```
ygwpz, line 410: The error message doesn't include `layer_name`. When this happens in production, it's hard to know which layer caused the issue. Could you include the layer name in the message?

ucm/integration/vllm/ucm_connector.py
```diff
@@ -393,6 +394,31 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
+                    raise ValueError(
+                        f"Unsupported kv cache tensor shape: {kv_layer.shape}"
+                    )
+            elif isinstance(kv_layer, Tuple):
+                for tensor in kv_layer:
+                    total_size = tensor.numel() * tensor.element_size()
+                    self.store.register_memory(tensor.data_ptr(), total_size)
```
ygwpz, line 417: Similarly, the type error doesn't mention which layer. Would help debugging to add `layer_name` here too.

ucm/store/pipeline/connector.py
```diff
@@ -209,9 +212,33 @@ def _posix_pipeline_builder(
...
+    pipeline.Stack(
+        "Mooncake", str(store_dir / "mooncakestore/libmooncakestore.so"), config
+    )
+
+
+def _mooncake_posix_pipeline_builder(
+    config: Dict[str, object], pipeline: ucmpipelinestore.PipelineStore
+):
+    store_dir = Path(__file__).resolve().parent.parent
+    posix_config = copy.deepcopy(config)
```
ygwpz, line 228: Why is `tensor_size` only set when `device_id >= 0`? If Posix is used as a fallback backend in non-NPU scenarios, shouldn't tensor_size still be needed?

ucm/store/test/e2e/mooncake_store_test.py
UESTC-AHao: Yes,I'll fix it.

examples/ucm_mooncake_config.yaml
UESTC-AHao: I will add comments to provide brief explanations for these parameters.

ucm/store/mooncakestore/CMakeLists.txt
UESTC-AHao: The current intention is to make it directly usable in the Ascend environment: because the Mooncake source code path in the Vllm-Ascend image is located at /vllm-workspace/Mooncake/mooncake-store/lib (a fixed path). However, considering future scalability and support for the CUDA ecosystem, switching to environment variables is indeed more reasonable.

ucm/store/mooncakestore/CMakeLists.txt
UESTC-AHao: Understood. I will make it more reasonable.

PR #951, opened 2026-05-09
[Feat] Add backend-only cache load mode

dante159753:
## Summary
- add `cache_load_backend_only` to bypass CacheStore lookup/load hits while keeping D2H/H2D staging
- preserve in-flight/shared-buffer load coalescing by only marking idle ready entries not-ready on explicit load-path gets
- add focused CacheStore/TransBuffer tests for the backend-only load behavior

## CacheStore implementation

This option is intentionally implemented inside CacheStore rather than by removing CacheStore from the pipeline. PosixStore only receives host-side addresses from the upper stage; CacheStore is still the stage that owns the pinned/staged host buffer and the D2H/H2D transfers between vLLM device KV cache and the backend store.

`cache_load_backend_only` defaults to `false`. When enabled, it controls whether CacheStore should avoid serving lookup/load hits from its in-memory transfer buffer.

When enabled, `BufferManager` changes lookup behavior:
- `Lookup()` and `LookupOnPrefix()` skip `TransBuffer::Exist()` and query the backend store directly.
- For `device_id == -1`, `BufferManager::Setup()` skips the shared-buffer watcher because scheduler lookup no longer needs to inspect CacheStore's shared buffer.

The backend load behavior is implemented in `TransBuffer::Get()`. `allowReserved` remains only an allocation-policy parameter. LoadQueue now explicitly calls `Get(block, shard, true, true)`, and the backend-only bypass decision is gated by the separate `isLoad` argument. DumpQueue continues to call `Get(block, shard)` and is not treated as a load path.

Inside `TransBuffer::Get()`:
- If there is no existing entry, allocation is unchanged.
- If an existing entry is idle (`Owner() == true`) and ready, backend-only load marks that entry not-ready before returning it. LoadQueue then sees `Owner() && !Ready()` and submits a backend load into the same host buffer.
- If an existing entry is not idle (`Owner() == false`), TransBuffer returns the same handle without clearing readiness. LoadQueue does not submit another backend load for non-owners; it waits for the owner to mark the buffer ready.

## Why this is correct

This keeps the data path needed for Posix performance tests while removing only the CacheStore hit shortcut.

For repeated loads of an already cached idle block, the ready bit is cleared while the returned handle owns the entry. That forces the load to descend to the backend, but it reuses the existing buffer address instead of allocating a fresh buffer. Because ownership is based on `reference == 0`, the code only clears an idle entry; it does not overwrite a buffer that another task is currently using.

For concurrent loads of the same block, only the first caller owns the entry. Later callers get non-owner handles and wait for readiness. This preserves the previous load coalescing behavior, including the shared-buffer/MLA case where multiple workers can request the same block at about the same time. The change therefore avoids the unsafe pattern of every worker independently loading into the same shared address.

For dumps, `isLoad` is false, so dump calls continue to use the existing CacheStore semantics. Dump-side D2H staging and backend submission are not affected.

For lookup, querying the backend directly is the intended semantics of this mode: a block should count as an external hit only if the lower backend reports it. CacheStore may still contain staged data, but it no longer short-circuits the hit decision.

## Tests

Added focused coverage for:
- backend-only lookup bypassing a ready CacheStore buffer and querying the backend instead
- backend-only load reusing an idle ready cache entry while marking it not-ready so backend load is forced
- backend-only reserved gets not bypassing unless the explicit `isLoad` flag is set
- backend-only load preserving in-flight coalescing by returning a non-owner handle for concurrent gets

## Validation
- `git diff --check origin2/develop..HEAD` passed
- no parameter hint comments remain in the changed CacheStore/test files
- commit hooks passed (`codespell`, formatting hooks skipped where no matching files)
- build/tests not run locally because `cmake` and `clang-format-20` are not available in this environment

### DIFF
--- a/docs/source/user-guide/prefix-cache/pipeline_store.md
+++ b/docs/source/user-guide/prefix-cache/pipeline_store.md
@@ -99,6 +99,9 @@ ucm_connectors:
 * **io_direct** (optional, default: `false`):  
   Whether to enable direct I/O.
 
+* **cache_load_backend_only** (optional, default: `false`):
+  Set to `true` to make lookup and load use the backend while keeping Cache Store Device-to-Host and Host-to-Device buffers.
+
 * **stream_number** *(optional, default: 8)*  
   Number of threads used for data transfer between the Host and Storage.
 
--- a/docs/source/user-guide/prefix-cache/pipeline_store.md
+++ b/docs/source/user-guide/prefix-cache/pipeline_store.md
@@ -100,7 +100,12 @@ ucm_connectors:
   Whether to enable direct I/O.
 
 * **cache_load_backend_only** (optional, default: `false`):
-  Set to `true` to make lookup and load use the backend while keeping Cache Store Device-to-Host and Host-to-Device buffers.
+  Set to `true` to make lookup and load bypass Cache Store hits and query/load from the backend store.
+  Cache Store is still kept in the pipeline as the transfer stage: dump still copies KV from device
+  to Cache Store's pinned host buffer before backend storage writes, and load still reads backend
+  data into that host buffer before copying it back to device. This is useful for backend
+  performance testing because it avoids in-memory cache hits without removing the D2H/H2D staging
+  path.
 
 * **stream_number** *(optional, default: 8)*  
   Number of threads used for data transfer between the Host and Storage.
--- a/docs/source/user-guide/prefix-cache/pipeline_store.md
+++ b/docs/source/user-guide/prefix-cache/pipeline_store.md
@@ -99,14 +99,6 @@ ucm_connectors:
 * **io_direct** (optional, default: `false`):  
   Whether to enable direct I/O.
 
-* **cache_load_backend_only** (optional, default: `false`):
-  Set to `true` to make lookup and load bypass Cache Store hits and query/load from the backend store.
-  Cache Store is still kept in the pipeline as the transfer stage: dump still copies KV from device
-  to Cache Store's pinned host buffer before backend storage writes, and load still reads backend
-  data into that host buffer before copying it back to device. This is useful for backend
-  performance testing because it avoids in-memory cache hits without removing the D2H/H2D staging
-  path.
-
 * **stream_number** *(optional, default: 8)*  
   Number of threads used for data transfer between the Host and Storage.
 
--- a/examples/ucm_config_example.yaml
+++ b/examples/ucm_config_example.yaml
@@ -16,6 +16,8 @@ ucm_connectors:
       store_pipeline: "Cache|Posix"
       storage_backends: "/mnt/test"
       io_direct: false
+      # Force lookup/load to use the backend while keeping CacheStore D2H/H2D buffers.
+      # cache_load_backend_only: false
       # cache_buffer_capacity_gb: 256
       # posix_capacity_gb: 1024
 
--- a/examples/ucm_config_example.yaml
+++ b/examples/ucm_config_example.yaml
@@ -16,8 +16,6 @@ ucm_connectors:
       store_pipeline: "Cache|Posix"
       storage_backends: "/mnt/test"
       io_direct: false
-      # Force lookup/load to use the backend while keeping CacheStore D2H/H2D buffers.
-      # cache_load_backend_only: false
       # cache_buffer_capacity_gb: 256
       # posix_capacity_gb: 1024
 
--- a/ucm/store/cache/cc/buffer_manager.h
+++ b/ucm/store/cache/cc/buffer_manager.h
@@ -34,6 +34,7 @@ namespace UC::CacheStore {
 class BufferManager {
     std::unique_ptr<TransBuffer> buffer_{nullptr};
     StoreV1* backend_{nullptr};
+    bool loadBackendOnly_{false};
 
     template <auto LookupFunc>
     auto LookupThrough(const Detail::BlockId* blocks, size_t num)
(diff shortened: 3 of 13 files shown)


ucm/store/cache/cc/buffer_manager.h
```diff
@@ -60,12 +64,14 @@ class BufferManager {
     TransBuffer* GetTransBuffer() { return buffer_ ? buffer_.get() : nullptr; }
     Expected<std::vector<uint8_t>> Lookup(const Detail::BlockId* blocks, size_t num)
     {
-        if (!buffer_) { return LookupThrough<&StoreV1::Lookup>(blocks, num); }
+        if (!buffer_ || loadBackendOnly_) { return LookupThrough<&StoreV1::Lookup>(blocks, num); }
         return LookupFast(blocks, num);
```
ygwpz, line 68: The `loadBackendOnly_` flag bypasses cache lookup, but what about Load operations? Should there be similar handling for load paths, or is the current design intentional?

ucm/store/cache/cc/trans_buffer.cc
```diff
@@ -490,6 +491,9 @@ TransBuffer::Handle TransBuffer::Get(const Detail::BlockId& blockId, size_t shar
     strategy_->BucketLock(iBucket);
     auto iNode = FindAt(iBucket, blockId, shardIdx, owner);
     if (iNode != invalidIndex) {
+        if (bypassHitOnReservedGet_ && allowReserved && owner && Ready(iNode)) {
```
ygwpz: The bypass condition `bypassHitOnReservedGet_ && allowReserved && owner && Ready(iNode)` marks a ready entry as not-ready. This could lead to redundant backend lookups if the data is already available. Is this the intended behavior for backend-only mode?

docs/source/user-guide/prefix-cache/pipeline_store.md
```diff
@@ -99,6 +99,9 @@ ucm_connectors:
 * **io_direct** (optional, default: `false`):  
   Whether to enable direct I/O.
 
+* **cache_load_backend_only** (optional, default: `false`):
```
ygwpz: The description "keeping Cache Store Device-to-Host and Host-to-Device buffers" is vague. Could you clarify what this means operationally? Does it affect D2H/H2D transfer efficiency?

ucm/store/cache/cc/buffer_manager.h
dante159753: This is intentional. `BufferManager` only changes the scheduler-side lookup decision: when `cache_load_backend_only` is enabled, lookup/prefix lookup asks the backend directly instead of consulting CacheStore's in-memory buffer.

The load path is handled one layer lower in `TransBuffer`. `LoadQueue` already calls `Get(block, shard, true)`, and `TransBuffer` uses that reserved-load path to clear `ready` on an idle owner entry. That makes `LoadQueue` see `Owner() && !Ready()` and submit `backend_->Load(...)` while still using CacheStore's host staging buffer for the subsequent H2D copy. Non-owner handles are left unchanged so in-flight/shared-buffer loads still coalesce instead of multiple workers loading the same block concurrently.

ucm/store/cache/cc/trans_buffer.cc
dante159753: Yes, this is the intended behavior for backend-only mode. The goal of this mode is specifically to avoid serving load hits from CacheStore's ready buffer so Posix/backend load performance can be measured.

docs/source/user-guide/prefix-cache/pipeline_store.md
dante159753: Updated in `419b58b` to make this explicit.

Operationally, this mode bypasses CacheStore hits for lookup/load decisions, but CacheStore remains in the pipeline as the transfer stage. Dump still copies KV from device to CacheStore's pinned host buffer before backend writes, and load still reads backend data into that host buffer before H2D back to device. So the mode is meant to avoid in-memory cache hits for backend performance testing without removing the existing D2H/H2D staging path.

examples/ucm_config_example.yaml
```diff
@@ -16,6 +16,8 @@ ucm_connectors:
       store_pipeline: "Cache|Posix"
       storage_backends: "/mnt/test"
       io_direct: false
+      # Force lookup/load to use the backend while keeping CacheStore D2H/H2D buffers.
+      # cache_load_backend_only: false
```
mag1c-h: This configuration item was added for internal testing purposes, so it's best not to include it in the user configuration template.

ucm/store/cache/cc/trans_buffer.cc
```diff
@@ -490,6 +491,9 @@ TransBuffer::Handle TransBuffer::Get(const Detail::BlockId& blockId, size_t shar
...
+            MarkNotReady(iNode);
```
mag1c-h: `allowReserved` is a parameter of the allocation strategy and should not be used to indicate whether it is a Load or Dump path, even if the current use case is that only the Load path will be configured.

examples/ucm_config_example.yaml
dante159753: removed

ucm/store/cache/cc/trans_buffer.cc
dante159753: Agreed. Fixed in `c2e8ffc`: add `isLoad` parameter for the load-path decision.

docs/source/user-guide/prefix-cache/pipeline_store.md
dante159753: Follow-up: removed this user-guide entry in `b48acfc` since the switch is internal-testing only. Kept the operational detail in the PR description instead.

mag1c-h (approved):
LGTM

PR #952, opened 2026-05-09
[Feat] Add layerwise load-ahead window

dante159753:
## Summary

- Add configurable `layerwise_load_ahead` for vLLM layerwise KV loading.
- Keep the default at `1` to preserve the existing one-layer lookahead behavior.
- Refill the load window by local layer order after each layer wait, avoiding duplicate submissions.
- Document the tuning option in the example config and PipelineStore guide.

## Validation

- Manually ran the new scheduling tests because local Python does not have `pytest` installed.
- Ran `python -m compileall ucm\integration\vllm\ucm_connector.py ucm\integration\vllm\tests\test_layerwise_load_ahead.py`.
- Commit hooks passed: codespell, black, isort.

## Notes

`layerwise_load_ahead=1` preserves the previous behavior. Values like `2` or `4` can improve overlap when layer load latency is higher than per-layer inference latency, but should be tuned with Cache Store queue, host buffer, and H2D stream pressure in mind.

### DIFF
--- a/ucm/integration/vllm/tests/test_layerwise_load_ahead.py
+++ b/ucm/integration/vllm/tests/test_layerwise_load_ahead.py
@@ -1,0 +1,352 @@
+from __future__ import annotations
+
+import importlib.util
+import sys
+import types
+from collections import defaultdict
+from pathlib import Path
+
+_MISSING = object()
+_STUBBED_MODULE_NAMES = [
+    "torch",
+    "numpy",
+    "vllm",
+    "vllm.config",
+    "vllm.distributed",
+    "vllm.distributed.kv_transfer",
+    "vllm.distributed.kv_transfer.kv_connector",
+    "vllm.distributed.kv_transfer.kv_connector.v1",
+    "vllm.distributed.kv_transfer.kv_connector.v1.base",
+    "vllm.distributed.parallel_state",
+    "vllm.model_executor",
+    "vllm.model_executor.models",
+    "vllm.model_executor.models.utils",
+    "vllm.platforms",
+    "vllm.v1",
+    "vllm.v1.core",
+    "vllm.v1.core.sched",
+    "vllm.v1.core.sched.output",
+    "ucm",
+    "ucm.integration",
+    "ucm.integration.vllm",
+    "ucm.integration.vllm.device",
+    "ucm.logger",
+    "ucm.observability",
+    "ucm.shared",
+    "ucm.shared.metrics",
+    "ucm.store",
+    "ucm.store.factory_v1",
+    "ucm.store.ucmstore_v1",
+    "ucm.utils",
+    "ucm.sparse",
+    "ucm.sparse.state",
+]
+
+
+def _module(name: str, *, package: bool = False) -> types.ModuleType:
+    module = types.ModuleType(name)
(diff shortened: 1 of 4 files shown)


ucm/integration/vllm/tests/test_layerwise_load_ahead.py
```diff
@@ -0,0 +1,352 @@
...
+    def extract_block_addrs(self, vllm_block_ids, layer_first=False):
+        assert layer_first is True
+        rows = []
+        for local_row in range(self.layer_count):
+            rows.append([[local_row * 1000 + block_id] for block_id in vllm_block_ids])
+        return rows
+
+
+def _make_metadata(*, include_failing_request: bool = False):
+    request_meta = {
```
ygwpz, line 250: Missing test coverage: no test for `load_ahead` > `total_layer_count`. What behavior is expected when the window exceeds available layers? Consider adding a test case for this edge condition.

ucm/integration/vllm/tests/test_layerwise_load_ahead.py
```diff
@@ -0,0 +1,352 @@
...
+    connector.layerwise_load_ahead = load_ahead
+    connector.layer_ids = [10, 11, 12, 13, 14]
+    connector.layer_name_to_id = {
+        f"layer.{layer_id}": layer_id for layer_id in connector.layer_ids
+    }
+    connector.first_layer_id = 10
+    connector.kv_cache_layout = FakeKVCacheLayout(len(connector.layer_ids))
+    connector.tp_rank = 0
+    connector.tp_size = 1
+    connector.is_mla = False
```
ygwpz, line 280: Missing test for validation errors: `_get_layerwise_load_ahead` should raise `ValueError` for negative or non-integer values, but there's no test verifying this behavior.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -836,6 +882,8 @@ def wait_for_layer_load(self, layer_name: str) -> None:
             return
```
ygwpz, line 882: The `should_refill_window` check adds to `_waited_load_layers` before popping from `load_tasks`. If the same layer is revisited (e.g., rollback or MTP paths), this could cause issues. Should the check happen after the pop?

ucm/integration/vllm/ucm_connector.py
```diff
@@ -780,7 +783,47 @@ def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
...
+    def _get_layer_local_row(self, layer_id: int) -> int:
+        layer_id_to_local_row = getattr(self, "layer_id_to_local_row", None)
+        if layer_id_to_local_row is not None:
+            return layer_id_to_local_row[layer_id]
+        return self.layer_ids.index(layer_id)
+
+    def _submit_next_load_layers(
+        self, metadata: "UCMConnectorMetadata", count: int
+    ) -> None:
+        submitted_count = 0
```
ygwpz, line 815: The `_submit_next_load_layers` loop condition `self._next_load_layer_index < len(self.layer_ids)` stops silently when index exceeds. Should there be a debug log or assertion to clarify expected behavior?

PR #955, opened 2026-05-11
[Bugfix] Adapt UCM connector to vLLM kv_cache_config

dante159753:
## Purpose
now latest VLLM requires kv_cache_config in constructor, adapt UCM connector to it, without breaking the execution of older VLLM version.

## Modifications 
add method `_init_kv_connector_base`, check if it receives kv_cache_config param, then call corresponding init method

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -1,12 +1,13 @@
 import copy
 import hashlib
+import inspect
 import math
 import os
 import pickle
 import time
 from collections import defaultdict
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, List, Optional, Tuple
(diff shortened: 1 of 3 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -1088,7 +1135,12 @@ def get_num_new_matched_tokens(
 
 
 class UCMLiteConnector(UCMDirectConnector):
-    def __init__(self, vllm_config, role):
+    def __init__(
+        self,
+        vllm_config,
```
ygwpz, line 1140: Missing type annotations for `vllm_config` and `role` parameters. All other connector classes use `vllm_config: "VllmConfig", role: KVConnectorRole` for consistency.

PR #959, opened 2026-05-14
[CI] Add vLLM Dockerfiles to support deepseek v4

dante159753:
## Purpose
Add vLLM Dockerfiles to support deepseek v4


scripts/build_cuda.sh
```diff
@@ -40,7 +40,7 @@ echo "Build version=$VERSION"
 function check_build_install()
 {
     cd ${KVCACHE_PROJECT_ROOT}
-    # pip3 install -r requirements.txt
+    pip3 install -r requirements.txt
```
ygwpz, line 43: build_cuda.sh now installs requirements.txt to get the `wheel` package, but build_ascend.sh still has this line commented out (line 57). If the Ascend build also needs `wheel` for building packages, this should be uncommented as well for consistency.

dante159753: ascend image come with this package

docker/Dockerfile.ucm-vllm-ascend.a2-v0.20.2
```diff
@@ -0,0 +1,29 @@
+# Set to other image if needed
+ARG IMAGE_SOURCE="quay.io/ascend"
+ARG IMAGE_NAME_VERSION="vllm-ascend:deepseekv4"
```
harrisonyhq, line 3: Why the base image here is deepseek v4, as the dockerfile is for v0.20.2?

dante159753: vllm ascend don't have tag > 0.18，it's recommended to use this image for deepseek

harrisonyhq (approved):

mag1c-h (approved):

PR #960, opened 2026-05-15
Add layerwise & pipeline store metrics to monitor performance

dante159753:
## Purpose
Add layerwise & pipeline store metrics to monitor performance

## Modifications 
add metrics
optimize the performance of metrics lib

## Test

performance compare with develop:
|  | tpot | ttft | throughput(token/s) |
| :--- | :---: | ---: | ---: |
| develop | 0.095 | 5.71 | 9.07 |
| this pr | 0.095 | 5.82(+2%) | 9.01(+0.66%) |


ucm/shared/metrics/cc/domain/metrics.h
```diff
@@ -36,24 +38,56 @@
 #include <unordered_map>
 #include <vector>
 namespace UC::Metrics {
+using MetricId = uint32_t;
+constexpr MetricId INVALID_METRIC_ID = std::numeric_limits<MetricId>::max();
+
+struct CachedMetric {
+    explicit CachedMetric(const char* metricName) : name{metricName} {}
+
+    const char* name;
```
ygwpz: Storing a raw `const char*` pointer here is risky — if the metric name is dynamically constructed or the original string goes out of scope, this pointer could become invalid. Consider using `std::string_view` or storing a copy.

ucm/shared/metrics/cc/api/metrics_api.h
```diff
@@ -33,11 +33,20 @@ void CreateStats(const std::string& name, const std::string& type);
...
 std::tuple<std::unordered_map<std::string, double>, std::unordered_map<std::string, double>,
-           std::unordered_map<std::string, std::vector<double>>>
+           std::unordered_map<std::string, std::vector<double>>, uint64_t>
 GetAllStatsAndClear();
 
 }  // namespace UC::Metrics
-#endif
\ No newline at end of file
+
+#define NAME_TO_METRIC_ID(name)                        \
```
ygwpz: This macro creates a static `CachedMetric` instance. Ensure static initialization order is correct — if this static is accessed before construction, the pointer could be invalid. Consider using function-local static or `constexpr`.

ucm/store/cache/cc/buffer_manager.h
```diff
@@ -43,6 +44,8 @@ class BufferManager {
         auto res = (backend_->*LookupFunc)(blocks, num);
         if (!res) [[unlikely]] { return decltype(res)(res.Error()); }
         UC_DEBUG("Cache lookup({}) in backend costs {:.3f}ms.", num, sw.Elapsed().count() * 1e3);
+        UC::Metrics::UpdateStats(NAME_TO_METRIC_ID("cache_lookup_backend_duration_ms"),
```
ygwpz, line 47: Adding multiple `UpdateStats` calls in the hot path (cache lookup) could impact performance. Consider batching updates or sampling (e.g., update every Nth call) for high-frequency operations.

ci/grafana/grafana_connector.json
```diff
@@ -0,0 +1,2472 @@
...
+        "refresh": 1,
+        "regex": "",
+        "skipUrlSync": false,
+        "sort": 2,
+        "type": "query"
+      },
+      {
+        "current": {
+          "selected": false,
+          "text": "/share/datasets/public_models/Meta-Llama-3-8B-Instruct",
```
ygwpz: This dashboard contains hardcoded paths `/share/datasets/public_models`. Consider using template variables or environment-specific configuration to make this portable across environments.

ucm/shared/metrics/cc/domain/metrics.cc
```diff
@@ -36,82 +36,160 @@ void Metrics::CreateStats(const std::string& name, const std::string& type)
...
+    return it == nameToId_.end() ? INVALID_METRIC_ID : it->second;
+}
+
+MetricId Metrics::ResolveCachedMetric(CachedMetric& metric) const
+{
+    auto id = metric.id.load(std::memory_order_acquire);
+    if (id != INVALID_METRIC_ID) { return id; }
+    if (!isInited_.load(std::memory_order_acquire)) { return INVALID_METRIC_ID; }
+    auto epoch = registerEpoch_.load(std::memory_order_acquire);
+    if (metric.seenEpoch.load(std::memory_order_acquire) == epoch) {
```
ygwpz: The epoch comparison logic should be documented — clarify when `registerEpoch_` advances and what happens if `CachedMetric` was seen in one epoch but registered later.

ucm/shared/metrics/cc/api/metrics_api.h
dante159753: This is already a function-local static inside the lambda, so initialization is thread-safe and does not depend on cross-TU static initialization order.

ucm/store/cache/cc/buffer_manager.h
dante159753: These calls now use cached metric ids and id-keyed buffers, so the hot path avoids string lookup/hash. Batching through a map would reintroduce that path, and sampling would change metric semantics.

ucm/shared/metrics/cc/domain/metrics.cc
dante159753: We prefer to keep this code self-explanatory without an extra comment here. `registerEpoch_` is advanced on successful metric registration, and a cached miss is retried after the epoch changes.

ucm/shared/metrics/cc/domain/metrics.h
dante159753: Changed this to store an owning `std::string`, so dynamically constructed names cannot leave a dangling pointer. The hot path still uses the cached metric id after the first resolve.

ci/grafana/grafana_connector.json
dante159753: Cleared the exported `model_name.current` default values from the dashboards so they are not tied to this environment-specific model path.

ucm/shared/metrics/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 file(GLOB_RECURSE UCMMETRICS_CC_SOURCE_FILES "./cc/*.cc")
-add_library(metrics STATIC ${UCMMETRICS_CC_SOURCE_FILES})
+add_library(metrics SHARED ${UCMMETRICS_CC_SOURCE_FILES})
```
flesher0813, line 2: Does this change necessary? I used to test the former version, it's ok for _store_ and _python connector_ to use one singleton.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -562,14 +562,13 @@ def get_num_new_matched_tokens(
             f"hit hbm: {hbm_hit_block_num * self.cp_world_size}, "
             f"hit external: {external_hit_blocks * self.cp_world_size}"
         )
-        if self.metrics_config:
```
flesher0813, line 564: If there are no other judgment conditions for _self.metrics_config_, shoule we stop storing it as an object member variable?

ucm/shared/metrics/CMakeLists.txt
dante159753: Yes, this is needed for the new store-side C++ metrics. With a static metrics library, cachestore/posixstore and the ucmmetrics pybind module can each own a separate Metrics singleton, so C++ UpdateStats() from store backends may not be visible to Python get_all_stats_and_clear(). Making metrics shared makes that singleton boundary explicit.

ucm/store/cache/cc/buffer_manager.h
```diff
@@ -104,6 +121,8 @@ class BufferManager {
         if (!res) [[unlikely]] { return res.Error(); }
         UC_DEBUG("Cache lookup({}/{}) in backend costs {:.3f}ms.", missBlk.size(), num,
                  sw.Elapsed().count() * 1e3);
+        UC::Metrics::UpdateStats(NAME_TO_METRIC_ID("cache_lookup_backend_duration_ms"),
```
flesher0813, line 124: Will this conversion improve performance? Can we put it into the function? Or we should modify the docs to match this

ucm/integration/vllm/ucm_connector.py
dante159753: Done. `metrics_config` is now kept as a local variable for logger setup only, and the remaining metrics update paths no longer depend on `self.metrics_config`.

ucm/store/cache/cc/buffer_manager.h
dante159753: yes use id instead of string can improve ~30% performance, I will update the doc

ygwpz (approved):

PR #966, opened 2026-05-20
[Bugfix] Rectifying the Exception in Total Throughput Calculation Results

Menglths:
## Purpose
Rectifying the Exception in Full Throughput Calculation Results

## Modifications 
Delete the repeated calculation line from the token_banchmark.py file.

## Test
All tests have passed.


ygwpz (approved):
Reviewed. The fix correctly removes the duplicate `completed_requests.append(metrics)` call that would have caused double-counting of metrics when `num_output_tokens > 0`.

PR #968, opened 2026-05-20
[opt]opt hma

wuhuxiao:
## Purpose

  Optimize and simplify the HMA FAWA connector implementation, reducing duplicated Ascend-specific connector logic while keeping FA/WA KV cache load and dump support in the unified FAWA path.

  ## Modifications

  - Refactored ucm/integration/vllm/hma_connector.py
      - Replaced segment-based KV cache mapping with group metadata and batched pointer extraction.
      - Added unified FA/WA pointer extraction logic for load and dump paths.
      - Simplified request metadata to record load/dump keys, hash ranges, and per-group vLLM block ids directly.
      - Batched dump submission across requests to reduce repeated FA/WA store task creation.
      - Removed the separate UCMAscendFAWAConnector and folded Ascend handling into UCMFAWAConnector.
  - Updated ucm/integration/vllm/ucm_connector.py
      - Simplified FAWA connector selection to use UCMFAWAConnector.can_handle_kv_cache_config(...).
      - Removed separate Ascend FAWA connector routing.

  ## Test
verified both on npu & gpu

### DIFF
--- a/ucm/integration/vllm/hma_connector.py
+++ b/ucm/integration/vllm/hma_connector.py
@@ -1062,13 +1062,14 @@ class UCMFAWAConnector(UCMDirectConnector):
             if self.wa_store is None:
                 raise RuntimeError("WA store is not initialized.")
 
-            total_keys: list[bytes] = []
+            fa_dump_keys: list[bytes] = []
+            wa_dump_keys: list[bytes] = []
             fa_ptr_rows: list[np.ndarray] = []
             wa_ptr_rows: list[np.ndarray] = []
             for request in metadata.request_meta.values():
                 if not request.dump_keys:
                     continue
-                total_keys.extend(request.dump_keys)
+                fa_dump_keys.extend(request.dump_keys)
                 fa_ptr_rows.append(
                     self._extract_fa_ptr(
                         request.dump_keys,
@@ -1077,6 +1078,7 @@ class UCMFAWAConnector(UCMDirectConnector):
                         request.dump_vllm_block_ids,
                     )
                 )
+                wa_dump_keys.extend(request.dump_keys[-1:])
                 wa_ptr_rows.append(
                     self._extract_wa_ptr(
                         request.dump_keys[-1:],
@@ -1084,7 +1086,7 @@ class UCMFAWAConnector(UCMDirectConnector):
                     )
                 )
 
-            if not total_keys:
+            if not fa_dump_keys:
                 return
 
             fa_ptrs = np.vstack(fa_ptr_rows)
@@ -1095,7 +1097,7 @@ class UCMFAWAConnector(UCMDirectConnector):
                 self._submit_dump_task(
                     "FA",
                     self.fa_store,
-                    total_keys,
+                    fa_dump_keys,
                     fa_ptrs,
                     event_handle,
                 )
@@ -1104,7 +1106,7 @@ class UCMFAWAConnector(UCMDirectConnector):
                 self._submit_dump_task(
                     "WA",
                     self.wa_store,
-                    total_keys,
+                    wa_dump_keys,
                     window_ptrs,
                     event_handle,
                 )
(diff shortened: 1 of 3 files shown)


ucm/integration/vllm/hma_connector.py
```diff
@@ -89,24 +75,18 @@ def handle_tensor(
...
-                raise ValueError(
-                    f"KV cache tensor has empty block dimension: {t.shape}"
-                )
-            token_strides.append(t.stride(token_dim) * t.element_size())
-            tensor_sizes_per_token.append(tensor_size // view_tensor_block_size)
-            view_tensor_block_sizes.append(view_tensor_block_size)
-            view_tensors.append(t)
+            tensor_block_size = int(t.shape[token_dim])
+            tensor_token_strides.append(t.stride(token_dim) * t.element_size())
+            tensor_sizes_per_token.append(tensor_size // tensor_block_size)
```
ygwpz, line 81: The validation check for "empty block dimension" (tensor_block_size <= 0) was removed in this refactor. Consider adding back this validation to prevent division by zero and potential runtime errors.

wuhuxiao: The tensor_block_size is inferred from the physical tensor dimensions and is strictly required to be a positive value.

qyh111 (approved):

mag1c-h (approved):

ucm/integration/vllm/hma_connector.py
```diff
@@ -720,71 +304,101 @@ def can_handle_kv_cache_config(
...
+
+        groups = self._kv_cache_config.kv_cache_groups
+        self.fa_group_ids, self.window_group_ids = [], []
+        layer_compress_ratios = getattr(
+            self._vllm_config.model_config.hf_config,
+            "compress_ratios",
+            None,
+        )
+        if layer_compress_ratios is None:
+            raise "current only support DSV4"
```
ygwpz, line 362: 🔴 **Critical:** `raise "current only support DSV4"` is invalid Python syntax. Python does not allow raising a string directly; you must raise an Exception. This should be `raise RuntimeError("current only support DSV4")` or `raise NotImplementedError("current only support DSV4")`. This will cause a TypeError at runtime.

ucm/integration/vllm/hma_connector.py
```diff
@@ -1162,13 +581,17 @@ def _store_tensor_size_list(
...
-            repeat = self.group_tail_blocks[group_id]
-            if repeat is None:
-                repeat = 1
-            for segment_tokens in self.group_window_spans[group_id][:repeat]:
+            meta = self.group_metas[group_id]
+
+            if not meta.tail_tokens:
+                continue
+
+            segment_tokens = meta.tail_tokens // meta.tail_blocks
```
ygwpz, line 589: ⚠️ **Warning:** `segment_tokens = meta.tail_tokens // meta.tail_blocks` could cause division by zero if `meta.tail_blocks` is 0. While `tail_blocks` is calculated with `max(..., 1)` in `_init_group_metas`, the `KVCacheGroupMeta` dataclass has default values that could be 0 if not properly initialized. Consider adding an assertion or validation.

ucm/integration/vllm/hma_connector.py
```diff
@@ -720,71 +304,101 @@ def can_handle_kv_cache_config(
...
+            compress_ratio = getattr(spec, "compress_ratio", 1)
+            token_block_size = kv_cache_spec.block_size
+
+            if self.is_ascend_layout:
+                # for ascend bug, wait for ascend fix
+                token_block_size = kv_cache_spec.block_size * compress_ratio
+
+            if window_size is None:
+                # hash_block_size must be an integral multiple of token_block_size.
+                tail_tokens = self.hash_block_size
```
ygwpz, line 378: 💡 **Suggestion:** `tail_tokens = self.hash_block_size` is assigned but not used in the window_size=None branch before being overwritten by the calculation at line 395. This appears to be unused code or could lead to confusion. Consider removing or clarifying the intent.

PR #973, opened 2026-05-23
[bugfix]split patch between 0.18.0 and 0.18.0rc1 for sfa

qyh111:


ucm/integration/vllm/patch/v0180/vllm_ascend/pc/attention/sfa_v1_0180.py
```diff
@@ -0,0 +1,275 @@
...
+                    fused_kv_no_split, kv_ag_handle = all_gather_async(
+                        torch.cat(
+                            [
+                                k_pe.view(-1, k_pe.shape[-1]),
+                                k_nope.view(-1, k_nope.shape[-1]),
+                                k_li.view(-1, k_li.shape[-1]),
+                            ],
+                            dim=1,
+                        ),
+                        get_tp_group(),
```
ygwpz, line 137: `get_tp_group()` is called here but not imported at the top of the file. Should add `from vllm.distributed.parallel_state import get_tp_group` or verify it's available from another import source.

PR #974, opened 2026-05-23
lite

qyh111:

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -1250,71 +1250,126 @@ class UCMMockConnector(UCMDirectConnector):
         return expect_hit_block_num * self.block_size, False
 
 
-class UCMLiteConnector(UCMDirectConnector):
+class UCMLiteConnector(KVConnectorBase_V1):
     def __init__(
         self,
         vllm_config,
         role,
         kv_cache_config: Optional["KVCacheConfig"] = None,
     ):
-        ucm_config = Config(vllm_config.kv_transfer_config)
-        launch_config = ucm_config.get_config()
-        enable_record_traces = launch_config.get("enable_record_traces", False)
-        persist_token_threshold = launch_config.get("persist_token_threshold", 0)
-        vllm_config.kv_transfer_config.kv_connector_extra_config = {
-            "ucm_connectors": [
-                {
-                    "ucm_connector_name": "UcmPipelineStore",
-                    "ucm_connector_config": {
-                        "store_pipeline": "Fake",
-                        "share_buffer_enable": True,
-                        "buffer_number": 244032232,
-                    },
-                }
-            ],
-            "enable_record_traces": enable_record_traces,
-            "persist_token_threshold": persist_token_threshold,
-            "use_lite": True,
-        }
-        super().__init__(vllm_config, role, kv_cache_config)
+        self.block_size = vllm_config.cache_config.block_size
+        self.hash_block_size = self.block_size
+        self.requests_meta: dict[str, RequestMeta] = {}
         self.total_block_nums = 0
-        self.total_hit_block_nums = 0
+
+        self.request_hasher = RequestHasher(vllm_config, 0)
+        self._seed = self.request_hasher("UCM_HASH_SEED")
+
+        super().__init__(vllm_config, role, kv_cache_config)
+
         logger.info("Init UCMLiteConnector.")
 
     def get_num_new_matched_tokens(self, request, num_computed_tokens):
-        super().get_num_new_matched_tokens(request, num_computed_tokens)
-
-        external_hit_blocks = 0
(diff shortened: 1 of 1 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -1250,71 +1250,81 @@ def get_num_new_matched_tokens(
...
-            total_ptrs = [[0]] * len(need_dump_blks)
-            try:
-                task = self.store.dump_data(need_dump_blks, shard_indexs, total_ptrs)
-                self.store.wait(task)
-            except Exception as e:
-                logger.error(
-                    f"request {request.request_id} wait dump task error. {type(e).__name__}: {e}"
-                )
-            self.requests_meta[request.request_id] = RequestMeta()
+        return 0, False
```
ygwpz, line 1311: This method always returns `(0, False)` meaning no tokens are matched. Is this intentional as a stub implementation, or should there be actual matching logic here?

ucm/integration/vllm/ucm_connector.py
```diff
@@ -1250,71 +1250,81 @@ def get_num_new_matched_tokens(
...
 
-        self.total_hit_block_nums += external_hit_blocks
+    def update_state_after_alloc(
+        self, request: "Request", blocks: "KVCacheBlocks", num_external_tokens: int
+    ):
+        pass
```
ygwpz, line 1316: Empty `pass` implementation. If this is required by the interface, consider adding a docstring explaining why no action is needed, or raise NotImplementedError if it should be implemented.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -1250,71 +1250,81 @@ def get_num_new_matched_tokens(
...
 
-        logger.info(
-            f"req external hit rate: {(external_hit_blocks / req_blocks_num):.2f}, "
-            f"total external hit rate: {(self.total_hit_block_nums / self.total_block_nums):.2f}"
-        )
-        return 0, False
+    def build_connector_meta(
+        self, scheduler_output: SchedulerOutput
+    ) -> KVConnectorMetadata:
+        return UCMConnectorMetadata({})
```
ygwpz: Returns an empty `UCMConnectorMetadata({})`. Is this correct for the lite connector, or should it contain some actual metadata?

ucm/integration/vllm/ucm_connector.py
```diff
@@ -1250,72 +1250,127 @@ def get_num_new_matched_tokens(
...
+            # store minimal RequestMeta for scheduler bookkeeping
+            self.requests_meta[request.request_id] = RequestMeta(
+                ucm_block_ids=ucm_block_ids,
+                hbm_hit_block_num=0,
+                total_hit_block_num=0,
+                num_token_ids=len(request.all_token_ids),
+                token_processed=0,
+            )
+
+            self.total_block_nums += req_blocks_num
```
ygwpz, line 1309: ⚠️ **Warning:** `self.total_block_nums` is only incremented when a new request is added to `requests_meta`. If a request already exists (e.g., after a chunked prefill), this counter won't be updated. This could lead to inconsistent statistics. Consider either incrementing for all requests or clarifying the intended behavior.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -1250,72 +1250,127 @@ def get_num_new_matched_tokens(
...
 
-        logger.info(
-            f"req external hit rate: {(external_hit_blocks / req_blocks_num):.2f}, "
-            f"total external hit rate: {(self.total_hit_block_nums / self.total_block_nums):.2f}"
-        )
         return 0, False
 
+    def update_state_after_alloc(
+        self, request: "Request", blocks: "KVCacheBlocks", num_external_tokens: int
```
ygwpz, line 1314: 💡 **Suggestion:** The stub methods (`update_state_after_alloc`, `start_load_kv`, `wait_for_layer_load`, `save_kv_layer`, `wait_for_save`) should have docstrings explaining that they are intentionally empty for the lite connector, or raise NotImplementedError if they are not meant to be called.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -1250,72 +1250,127 @@ def get_num_new_matched_tokens(
...
+        kv_layer: torch.Tensor,
+        attn_metadata: "AttentionMetadata",
+        **kwargs: Any,
+    ) -> None:
+        pass
+
+    def wait_for_save(self):
+        pass
+
+    def generate_hash(
```
ygwpz, line 1351: ⚠️ **Warning:** The `generate_hash` method uses `self.request_hasher` which could raise exceptions. There's no error handling here. If the hasher fails, the exception will propagate up and could crash the request processing. Consider adding a try-except block or documenting the expected behavior.

PR #976, opened 2026-05-25
[opt] Fix FAWA HMA connector dump handling and optimize MLA TP dump balance.

wuhuxiao:
## Purpose

  Fix FAWA HMA connector dump handling and optimize MLA TP dump balance.

  ## Modifications

  - Add FAWA dump task tracking for tensor-parallel workers.
  - Split FA dump rows across TP ranks and distribute WA dump work by request.
  - Wait pending dump tasks before preemption or request block release.
  - Restore `get_finished` and `request_finished_all_groups` behavior with explicit FAWA handling.
  - Improve KV cache group layout comments and metadata helpers.

  ## Test

  - Verified

### DIFF
--- a/ucm/integration/vllm/hma_connector.py
+++ b/ucm/integration/vllm/hma_connector.py
@@ -1189,14 +1189,15 @@ class UCMFAWAConnector(UCMDirectConnector, SupportsHMA):
     ) -> tuple[set[str] | None, set[str] | None]:
         # Worker side method
         try:
-            finished_request_ids = []
-            for request_ids, dump_tasks in self.tp_dump_tasks.items():
-                if finished_req_ids.intersection(request_ids):
-                    finished_request_ids.append(request_ids)
-                    for dump_task in dump_tasks:
-                        self._wait_dump_task(dump_task)
-            for request_ids in finished_request_ids:
-                self.tp_dump_tasks.pop(request_ids, None)
+            if finished_req_ids:
+                finished_chunk_req_ids = []
+                for request_ids, dump_tasks in self.tp_dump_tasks.items():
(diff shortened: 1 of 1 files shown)


ucm/integration/vllm/hma_connector.py
```diff
@@ -1072,52 +1128,164 @@ def wait_for_save(self) -> None:
...
+            reported through this interface.
+        """
+        return None, None
+
+    def request_finished_all_groups(
+        self,
+        request: "Request",
+        block_ids: tuple[list[int], ...],
+    ) -> tuple[bool, dict[str, object] | None]:
+        """Handle HMA request completion for all KV-cache groups.
```
qyh111: This is a `scheduler` size method, can tasks be waited here?

ucm/integration/vllm/hma_connector.py
```diff
@@ -362,30 +389,31 @@ def _init_group_metas(self) -> None:
             raise "current only support DSV4"
```
ygwpz: This `raise` statement is invalid Python syntax. Strings cannot be raised directly - use `raise ValueError("current only support DSV4")` or similar instead.

ucm/integration/vllm/hma_connector.py
```diff
@@ -743,7 +805,7 @@ def _generate_dispatch_meta(
             load_end = req_meta.total_hit_block_num
```
ygwpz, line 791: This changes `np.arange(load_start, load_end)` to `np.arange(load_start + 1, load_end + 1)`. Please verify this is intentional and document why the boundary indexing shifted by one.

qyh111 (approved):

PR #977, opened 2026-05-25
[Feat] Add ASU SQE protocol implementations

yumingyue624:
## Purpose
Implement ASU transport protocol packet builders for SQE commands and link establishment

## Modifications
* Add SQE classes: Store, Retrieve, BatchStore, BatchRetrieve, Delete, Exist, KeepAlive
* Add link establishment protocol: NegotiateSqe, ConnectSqe, DisconnectSqe

## Test
All 11 unit tests passed (7 SQE pack + 4 link protocol pack)

### DIFF
--- a/ucm/transport/kv/asu/trans/src/link_proto.h
+++ b/ucm/transport/kv/asu/trans/src/link_proto.h
@@ -1,0 +1,163 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2026 Huawei Technologies Co., Ltd. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ * */
+#pragma once
+
+#include <cstdint>
+#include <cstring>
+#include <string>
+#include <vector>
+#include "asu_transport/types.h"
+
+namespace UC::ASU {
+
+constexpr std::uint32_t kMsgHeaderSize = 16;
+constexpr std::uint32_t kMsgVersion = 1;
+
+// LinkProtocolCmd defines the message types for the KV connection protocol.
+enum class LinkProtocolCmd : std::uint8_t {
+    Negotiate = 0,
+    Handshake = 1,
+    HandshakeDone = 3,
+    Disconnect = 4
+};
+
+constexpr std::size_t kNegotiatePayloadSize = 160;  // 32 + 128
+constexpr std::size_t kHandshakePayloadSize = 160;
+constexpr std::size_t kDisconnectPayloadSize = 8;
+
+class MsgHeader {
+public:
+    std::uint32_t crc{0};
+    std::uint8_t ver{kMsgVersion};
+    LinkProtocolCmd cmd{LinkProtocolCmd::Negotiate};
+    std::uint8_t pad[6]{0};
+    std::uint32_t len{0};
+
+    void Pack(std::vector<std::uint8_t>& buffer) const
+    {
+        std::memcpy(&buffer[0], &crc, 4);
+        buffer[4] = ver;
+        buffer[5] = static_cast<std::uint8_t>(cmd);
+        std::memcpy(&buffer[12], &len, 4);
+    }
+
+    static MsgHeader Unpack(const std::uint8_t* data)
+    {
+        MsgHeader h;
+        std::memcpy(&h.crc, &data[0], 4);
+        h.ver = data[4];
+        h.cmd = static_cast<LinkProtocolCmd>(data[5]);
+        std::memcpy(&h.len, &data[12], 4);
+        return h;
+    }
+};
+
+class NegotiateRequest {
+public:
+    std::uint32_t cap{0};
+    std::uint32_t private_len{4};
+    std::uint8_t major_version{1};
+    std::uint8_t minor_version{0};
+    std::uint16_t kato{0};
+};
+
+class HandshakeRequest {
+public:
+    std::uint8_t gid[16]{0};
+    std::uint16_t lid{0};
+    std::uint8_t mtu{0};
+    std::uint8_t total_qp_num{0};
+    std::uint8_t sl{0};
+    std::uint8_t traffic_class{0};
+    std::uint8_t rnr_timer{0};
+    std::uint8_t rnr_retry_cnt{0};
+    std::uint8_t timeout{0};
+    std::uint8_t retry_cnt{0};
+    std::uint8_t qp_rd_atom{0};
+    std::uint8_t rsv{0};
+    std::uint32_t start_psn{0};
+    std::uint32_t qpn[32]{0};
+};
+
+class DisconnectRequest {
+public:
+    std::uint32_t local_qpn{0};
+    std::uint32_t remote_qpn{0};
+};
+
+class NegotiateSqe {
+public:
+    NegotiateSqe() = default;
+
+    const void* Data() const { return buffer.data(); }
+    std::size_t Size() const { return buffer.size(); }
+
+    Status Pack(const NegotiateRequest& req);
+    Status Validate() const;
+
+private:
+    std::vector<std::uint8_t> buffer;
+};
+
+class HandshakeSqe {
+public:
+    HandshakeSqe() = default;
+
+    const void* Data() const { return buffer.data(); }
+    std::size_t Size() const { return buffer.size(); }
+
+    Status Pack(const HandshakeRequest& req);
+    Status Validate() const;
+
+private:
+    std::vector<std::uint8_t> buffer;
+};
+
+class HandshakeDoneSqe {
+public:
+    HandshakeDoneSqe() = default;
+
+    const void* Data() const { return buffer.data(); }
+    std::size_t Size() const { return buffer.size(); }
+
+    Status Pack();
+
+private:
+    std::vector<std::uint8_t> buffer;
+};
+
+class DisconnectSqe {
+public:
+    DisconnectSqe() = default;
+
+    const void* Data() const { return buffer.data(); }
+    std::size_t Size() const { return buffer.size(); }
+
+    Status Pack(const DisconnectRequest& req);
+
+private:
+    std::vector<std::uint8_t> buffer;
+};
+
+}  // namespace UC::ASU
--- a/ucm/transport/kv/asu/trans/src/sqe.cpp
+++ b/ucm/transport/kv/asu/trans/src/sqe.cpp
@@ -1,0 +1,543 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2026 Huawei Technologies Co., Ltd. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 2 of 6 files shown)


ucm/transport/kv/asu/trans/src/test_store_sqe.cpp
```diff
@@ -0,0 +1,94 @@
+#include <cstdio>
+#include <cstring>
+#include "store_sqe.h"
+
```
ygwpz: The include directive uses `"store_sqe.h"` but the actual file is named `"sqe.h"`. This will fail to compile.

ucm/transport/kv/asu/trans/src/sqe.h
```diff
@@ -0,0 +1,807 @@
...
+
+private:
+    static std::unordered_map<std::uint32_t, Creator>& Creators()
+    {
+        static std::unordered_map<std::uint32_t, Creator> instance;
+        return instance;
+    }
+};
+
+#define REGISTER_SQE(ClassName, Opcode) \
```
ygwpz: The REGISTER_SQE macro defines a static variable `_reg_##ClassName` in a header file. When this header is included in multiple translation units, each will have its own copy, potentially causing registration to run multiple times. Consider using `inline` for the static variable or moving registration to a separate .cpp file.

ucm/transport/kv/asu/trans/src/sqe.h
```diff
@@ -0,0 +1,807 @@
...
+        // Dwords 12-15: key[15:0] - 16 bytes key, low-byte aligned
+        std::size_t key_len = std::min(r.key.size(), static_cast<std::size_t>(16));
+        if (key_len > 0) { std::memcpy(&dwords[12], r.key.data(), key_len); }
+    }
+
+    Status Validate() const override
+    {
+        if (dwords.size() < kSqeDwordCount) {
+            return Status::Error(StatusCode::NOT_INITIALIZED, "SQE not packed");
+        }
```
ygwpz: The `dtype` and `dspec` values are masked with `& 0xF` but not validated against expected ranges. Should validation be added in Validate() to check if these values are within acceptable bounds?

ucm/transport/kv/asu/trans/src/test_store_sqe.cpp
yumingyue624: Fixed the include path in test_store_sqe.cpp from "store_sqe.h" to "sqe.h" to match the actual file name.

ucm/transport/kv/asu/trans/src/sqe.h
yumingyue624: Changed the REGISTER_SQE macro to use inline static bool instead of static bool. This ensures a single definition across translation units and prevents the registration lambda from running multiple times.

ucm/transport/kv/asu/trans/src/sqe.h
```diff
@@ -0,0 +1,807 @@
...
+        if (key_empty) { return Status::Error(StatusCode::INVALID_ARGUMENT, "key is empty"); }
+        return Status::OK();
+    }
+
+private:
+    std::vector<std::uint32_t> dwords;
+};
+
+REGISTER_SQE(StoreSqe, kStoreOpcode)
```
nrj868: Put all REGISTER_SQE calls together

ucm/transport/kv/asu/trans/src/test_store_sqe.cpp
```diff
@@ -0,0 +1,98 @@
...
+        printf("Validation failed: %s\n", status.message.c_str());
+        return 1;
+    }
+
+    PrintSqe(sqe);
+
+    printf("\n=== Verification ===\n");
+    printf("Dword 0: CID=0x%04X, Fixed=0x%02X, Opcode=0x%02X (expect 0x12340101)\n",
+           req.cid, kFixedBits, kStoreOpcode);
+    printf("  Actual: 0x%08X\n", sqe.GetDword(0));
```
nrj868: Using GTest for DT, put them under the test folder

ucm/transport/kv/asu/trans/src/sqe.h
```diff
@@ -0,0 +1,807 @@
...
+        // Entry Dword 1: key[15:0]
+        std::size_t key_len = std::min(entry.key.size(), static_cast<std::size_t>(16));
+        if (key_len > 0) { std::memcpy(&dwords[base + 1], entry.key.data(), key_len); }
+
+        // Entry Dword 5-6: Buffer Address[63:0]
+        dwords[base + 5] = entry.buffer_addr & 0xFFFFFFFFULL;
+        dwords[base + 6] = (entry.buffer_addr >> 32) & 0xFFFFFFFFULL;
+
+        // Entry Dword 8: MR_KEY[0][31:24] = MR_KEY low 8 bits | Length[23:0]
+        dwords[base + 8] = ((entry.mr_key & 0xFF) << 24) | (entry.length & 0xFFFFFF);
```
nrj868: Should last two dwords be dwords[base + 7]  and dwords[base + 8] ？Please check other opcode too

ucm/transport/kv/asu/trans/src/sqe.h
```diff
@@ -0,0 +1,807 @@
...
+        dwords[10] = r.batch_number;
+
+        // Dword 11: LR[31]
+        if (r.lr) { dwords[11] |= (1U << 31); }
+
+        // Dwords 12-15: reserved (zero)
+
+        // Pack batch entries
+        for (std::size_t i = 0; i < r.batch_number; ++i) {
+            PackEntry(r.entries[i], i);
```
nrj868: Validate the entries size against batch_number before using it

ucm/transport/kv/asu/trans/src/sqe.h
```diff
@@ -0,0 +1,787 @@
...
+        // Entry Dword 5-6: Buffer Address[63:0]
+        dwords[base + 5] = entry.buffer_addr & 0xFFFFFFFFULL;
+        dwords[base + 6] = (entry.buffer_addr >> 32) & 0xFFFFFFFFULL;
+
+        // Entry Dword 8: MR_KEY[0][31:24] = MR_KEY low 8 bits | Length[23:0]
+        dwords[base + 8] = ((entry.mr_key & 0xFF) << 24) | (entry.length & 0xFFFFFF);
+
+        // Entry Dword 9: DPTR.Type = 0x40 | MR_KEY[3:1][23:0] = MR_KEY high 24 bits
+        dwords[base + 9] = (static_cast<std::uint32_t>(DptrType::Standard) << 24) | ((entry.mr_key >> 8) & 0xFFFFFF);
+    }
```
ygwpz: kBatchEntryDwordCount is 9, meaning each entry should use indices 0-8. Writing to base + 9 overflows into the next entry. Should this be base + 7?

ucm/transport/kv/asu/trans/src/sqe.h
```diff
@@ -0,0 +1,787 @@
...
```
ygwpz: Same issue as BatchStoreSqe: kBatchEntryDwordCount is 9, so base + 9 overflows. Should this be base + 7?

ucm/transport/kv/asu/trans/src/sqe.h
```diff
@@ -0,0 +1,787 @@
...
+
+        // Dword 11: LR[31]
+        if (r.lr) { dwords[11] |= (1U << 31); }
+
+        // Dwords 12-15: reserved (zero)
+
+        // Pack batch entries
+        for (std::size_t i = 0; i < r.batch_number; ++i) {
+            PackEntry(r.entries[i], i);
+        }
```
ygwpz: Should validate r.entries.size() >= r.batch_number before this loop to prevent out-of-bounds access.

ucm/transport/kv/asu/trans/src/sqe.h
yumingyue624: Fixed in both BatchStoreSqe::PackEntry and BatchRetrieveSqe::PackEntry.

ucm/transport/kv/asu/trans/src/sqe.h
yumingyue624: Fixed in both BatchStoreSqe::PackEntry and BatchRetrieveSqe::PackEntry.

ucm/transport/kv/asu/trans/src/sqe.h
yumingyue624: Fixed in both BatchStoreSqe::PackEntry and BatchRetrieveSqe::PackEntry.

ucm/transport/kv/asu/trans/src/sqe.h
yumingyue624: Done. Added a pre-condition check with [[unlikely]] before the loop in all four batch SQE classes (BatchStore, BatchRetrieve, Delete, Exist):
if (r.batch_number > r.entries.size()) [[unlikely]] {
    throw std::runtime_error{"batch_number exceeds entries.size()"};
}

ucm/transport/kv/asu/trans/src/sqe.h
yumingyue624: Done. Added a pre-condition check with [[unlikely]] before the loop in all four batch SQE classes (BatchStore, BatchRetrieve, Delete, Exist):
if (r.batch_number > r.entries.size()) [[unlikely]] {
    throw std::runtime_error{"batch_number exceeds entries.size()"};
}

ucm/transport/kv/asu/trans/src/sqe.h
yumingyue624: Done. All 7 REGISTER_SQE calls are now consolidated at the end of the file, right before the closing namespace brace.

ucm/transport/kv/asu/trans/src/test_store_sqe.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+using namespace UC::ASU;
+
+void PrintSqe(const StoreSqe& sqe) {
+    for (int i = 0; i < 16; ++i) {
+        printf("Dword %2d: 0x%08X\n", i, sqe.GetDword(i));
+    }
+}
+
+int main() {
```
Infinite666: Add more unit test with gtest for each cases, use expect equal to determine if the mem value meet expectations.

ucm/transport/kv/asu/trans/src/sqe.h
```diff
@@ -0,0 +1,793 @@
...
+    virtual void Pack(const SqeRequest& req) = 0;
+    virtual Status Validate() const = 0;
+
+protected:
+    // TODO: Optimize memory allocation. Currently using vector for dynamic size,
+    // but final implementation should write directly into SendBuffer to avoid copying.
+    std::vector<std::uint32_t> dwords;
+};
+
+class SqeRegistry {
```
Infinite666, line 182: Every op will create a Seq object, here can be refind.

ucm/transport/kv/asu/trans/src/test_store_sqe.cpp
yumingyue624: Done. GTest-based unit tests:
- sqe_pack_test.cc: 7 test cases covering Store, Retrieve, BatchStore, BatchRetrieve, Delete, Exist, KeepAlive
- link_proto_pack_test.cc: 4 test cases covering Negotiate, Handshake, HandshakeDone, Disconnect
Each test builds the expected memory layout from the protocol spec and uses EXPECT_EQ to verify byte-by-byte against the Pack() output. All 11 tests passed.

ucm/transport/kv/asu/trans/src/test_store_sqe.cpp
yumingyue624: Done. GTest-based unit tests:
- sqe_pack_test.cc: 7 test cases covering Store, Retrieve, BatchStore, BatchRetrieve, Delete, Exist, KeepAlive
- link_proto_pack_test.cc: 4 test cases covering Negotiate, Handshake, HandshakeDone, Disconnect
Each test builds the expected memory layout from the protocol spec and uses EXPECT_EQ to verify byte-by-byte against the Pack() output. All 11 tests passed.

ucm/transport/kv/asu/trans/src/sqe.h
yumingyue624: Will be refactored together with the SendBuffer PR, where SQE packing will write directly into pre-allocated SendBuffer memory instead of creating a new Sqe object per operation.

ucm/transport/kv/asu/trans/src/sqe.h
yumingyue624: Will be refactored together with the SendBuffer PR, where SQE packing will write directly into pre-allocated SendBuffer memory instead of creating a new Sqe object per operation.

nrj868 (approved):

ucm/transport/kv/asu/trans/src/link_proto.h
```diff
@@ -0,0 +1,163 @@
...
+    void Pack(std::vector<std::uint8_t>& buffer) const
+    {
+        std::memcpy(&buffer[0], &crc, 4);
+        buffer[4] = ver;
+        buffer[5] = static_cast<std::uint8_t>(cmd);
+        std::memcpy(&buffer[12], &len, 4);
+    }
+
+    static MsgHeader Unpack(const std::uint8_t* data)
+    {
```
ygwpz, line 66: ⚠️ **Warning:** `MsgHeader::Unpack` reads from `data` without validating that `data` is non-null and has at least `kMsgHeaderSize` (16) bytes. This could cause undefined behavior if called with invalid input. Consider adding a null check and size validation, or document that the caller must ensure valid input.

ucm/transport/kv/asu/trans/src/sqe.cpp
```diff
@@ -0,0 +1,543 @@
...
+#include "sqe.h"
+#include <algorithm>
+#include <cstring>
+
+namespace UC::ASU {
+
+Status KvStoreSqe::Pack(const SqeRequest& req)
+{
+    auto& r = static_cast<const KvStoreRequest&>(req);
+    dwords.assign(kSqeDwordCount, 0);
```
ygwpz, line 33: ⚠️ **Warning:** `static_cast<const KvStoreRequest&>(req)` has no type validation. If the caller passes a wrong request type (e.g., `KvRetrieveRequest`), this will cause undefined behavior. Consider using a virtual method approach or adding a runtime type check (e.g., `req.GetType() == RequestType::Store`).

ucm/transport/kv/asu/trans/src/sqe.cpp
```diff
@@ -0,0 +1,543 @@
...
+        }
+    }
+    if (key_empty) { return Status::Error(StatusCode::INVALID_ARGUMENT, "key is empty"); }
+    return Status::OK();
+}
+
+Status KvBatchStoreSqe::Pack(const SqeRequest& req)
+{
+    auto& r = static_cast<const KvBatchStoreRequest&>(req);
+    if (r.batch_number > r.entries.size()) [[unlikely]] {
```
ygwpz, line 176: ⚠️ **Warning:** Same as above - `static_cast<const KvBatchStoreRequest&>(req)` has no type validation. The `Pack` method accepts a base class reference but assumes the derived type. This design is fragile and could lead to crashes if the wrong request type is passed.

PR #980, opened 2026-05-27
[Feat] Optimize ASU task manager with lock-free slots

AooooooA-C:
## Purpose
This PR replaces the mutex-protected `unordered_map` in `TaskManagerBase` with a fixed-size slot table based on open addressing and atomic state transitions. The goal is to reduce global lock contention.
## Modifications 

The original implementation:

```cpp
std::mutex mu_;
std::unordered_map<TaskId, std::shared_ptr<Context>> tasks_;
```

is replaced by a fixed-size slot array:

```cpp
std::vector<Slot> slots_;
```

Each slot stores:

```text
state
task_id
ctx
```
#### 1. Create and enter the build directory
```bash
cd /YOURPATH/unified-cache-management
mkdir -p build && cd build
```
#### 2. Configure CMake
```bash
cmake .. \
  -DBUILD_UCM_ASU=ON \
  -DBUILD_UNIT_TESTS=ON 
  -DBUILD_UCM_STORE=OFF \
  -DDOWNLOAD_DEPENDENCE=ON \
  -DRUNTIME_ENVIRONMENT=ascend
```
#### 3. Build the asu.test target
```bash
cmake --build . --target asu.test -j$(nproc))
```
#### 4. Run the tests
```bash
./ucm/transport/kv/asu/asu.test --gtest_also_run_disabled_tests --gtest_filter='TransportTaskManagerTest*:ClientTaskManagerTest*:TaskManagerBaseBench*:SlotTaskManagerBaseBench*'
```

#### 5. Test Results
Unit tests for SlotTaskManagerBase verifying submit/get/remove/getAll correctness:
2026-06-05-task_manager_slot_test

Benchmark: TaskManagerBase (mutex + unordered_map) vs SlotTaskManagerBase (lock-free slots + IndexPool)
2026-06-05-task_manager_base_bench


ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -24,77 +24,245 @@
...
+            kMinSlotCount,
+            max_inflight_tasks * 2);
+        return NormalizeSlotCount(required);
     }
 
-    Status Submit(std::unique_ptr<Context> ctx, TaskId& taskId)
+    explicit TaskManagerBase(
+        State initial_state,
+        std::string task_name,
+        std::size_t slot_count = kDefaultSlotCount)
```
nrj868: Make sure using `RecommendSlotCount` to calculate slot count

ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -24,77 +24,245 @@
...
+        const auto start = Hash(new_task_id) & slot_mask_;
+        const auto capacity = slots_.size();
+
+        for (std::size_t probe = 0; probe < capacity; ++probe) {
+            auto& slot = slots_[(start + probe) & slot_mask_];
+
+            // CAS: Try to transition EMPTY → WRITING
+            std::uint8_t expected = SlotState::EMPTY;
+            if (!slot.state.compare_exchange_strong(
+                    expected,
```
nrj868: Please clang-formating this pr

ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -24,77 +24,245 @@
...
+        }
+
+        task_id = kInvalidTaskId;
 
-        sharedCtx->taskId = taskId;
-        tasks_.emplace(taskId, std::move(sharedCtx));
-        return Status::OK();
+        // Consider adding RESOURCE_EXHAUSTED / NO_SPACE error codes to StatusCode
+        return Status::Error(
+            StatusCode::INVALID_ARGUMENT,
```
nrj868: Using `INTERNAL_ERROR` in here

ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -24,77 +24,245 @@
...
-        if (iter == tasks_.end()) { return nullptr; }
-        return iter->second;
+        if (task_id == kInvalidTaskId) {
+            return nullptr;
+        }
+
+        const auto start = Hash(task_id) & slot_mask_;
+        const auto capacity = slots_.size();
+
+        for (std::size_t probe = 0; probe < capacity; ++probe) {
```
nrj868: In the worst case， both `Submit` and `Get` run in O(n)

ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -24,77 +24,245 @@
...
+        if (task_id == kInvalidTaskId) {
+            return Status::Error(
+                StatusCode::TASK_NOT_FOUND,
+                task_name_ + " task not found");
+        }
+
+        const auto start = Hash(task_id) & slot_mask_;
+        const auto capacity = slots_.size();
+
+        for (std::size_t probe = 0; probe < capacity; ++probe) {
```
nrj868: Same as Submit & Get,  it should be in O(1)

ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -51,27 +74,89 @@ class TaskManagerBase {
...
+            }
+        }
+
+        return nullptr;
     }
 
     std::vector<std::shared_ptr<Context>> GetAll()
     {
-        std::lock_guard<std::mutex> lock(mutex_);
+        std::lock_guard<std::mutex> lock(mu_);
```
ygwpz: `GetAll()` still uses mutex-based locking while `Add`, `Get`, and `Remove` are lock-free. This inconsistency could lead to race conditions if `GetAll()` is called concurrently with other operations. Consider implementing a lock-free version of `GetAll()` or documenting the thread-safety limitations.

ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -80,21 +165,114 @@ class TaskManagerBase {
...
     }
 
 private:
     State initialState_;
     std::string taskName_;
     std::atomic<TaskId> nextTaskId_{1};
     // TODO: consider using a lock-free structure !
-    std::mutex mutex_;
+    std::mutex mu_;
     std::unordered_map<TaskId, std::shared_ptr<Context>> tasks_;
```
ygwpz, line 97: The `tasks_` unordered_map member appears to be unused after the lock-free refactor. `Add`, `Get`, and `Remove` now use the slot-based approach. Should this member be removed to avoid confusion, or is it kept for backward compatibility?

ucm/transport/kv/asu/common/task_manager_base.h
AooooooA-C: Calculated slot count using RecommendSlotCount based on maxInflightTasks.

ucm/transport/kv/asu/common/task_manager_base.h
AooooooA-C: Updated as suggested

ucm/transport/kv/asu/common/task_manager_base.h
AooooooA-C: Revised the implementation: free slots are now managed via a free list instead of linear probing.

ucm/transport/kv/asu/common/task_manager_base.h
AooooooA-C: Updated as suggested

ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -48,53 +76,249 @@ class TaskManagerBase {
...
+            }
+        }
+    }
+
+    void FreeListPush(std::size_t slotIndex)
+    {
+        auto oldHead = freeListHead_.load(std::memory_order_acquire);
+        while (true) {
+            slots_[slotIndex].freeNext.store(static_cast<std::size_t>(oldHead & freeListMask_),
+                                             std::memory_order_release);
```
ygwpz: ⚠️ **Warning:** In `FreeListPush`, the `freeNext.store(..., memory_order_release)` must be visible before the `compare_exchange_weak` on `freeListHead_`. However, the current ordering (release for store, then release/acquire for CAS) may not guarantee this on all architectures. Consider using `memory_order_seq_cst` for the CAS success ordering, or add a fence to ensure the store is visible before the head update.

ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -48,53 +76,249 @@ class TaskManagerBase {
...
+    {
+        const auto raw =
+            (generation << slotIndexBits_) | static_cast<TaskIdUInt>(slotIndex & slotMask_);
+        return static_cast<TaskId>(raw);
+    }
+
+    static std::shared_ptr<Context> AtomicLoadCtx(const Slot& slot, std::memory_order order)
+    {
+        return std::atomic_load_explicit(&slot.ctx, order);
+    }
```
ygwpz: 💡 **Suggestion:** `std::atomic_load_explicit` on `std::shared_ptr<Context>` requires C++20 or a library that supports atomic operations on shared_ptr. If using C++17, this may not work on all platforms. Consider using `std::atomic_load(&slot.ctx)` (C++20) or ensure the compiler/library supports this feature.

ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -48,53 +76,249 @@ class TaskManagerBase {
...
+        if (!ptr) { return nullptr; }
+
+        const auto id2 = slot.taskId.load(std::memory_order_acquire);
+        const auto state2 = slot.state.load(std::memory_order_acquire);
+        if (state2 == SlotState::READY && id2 == taskId && ptr->taskId == taskId) { return ptr; }
+
+        return nullptr;
     }
 
     std::vector<std::shared_ptr<Context>> GetAll()
```
ygwpz, line 72: ⚠️ **Warning:** `GetAll()` iterates through all slots and does double-checking, but the iteration is not atomic with respect to concurrent Submit/Remove operations. If a slot is removed and immediately reused during GetAll, the same context could be returned twice (from the old slot iteration and the new slot). Consider documenting this behavior or adding a comment about the trade-off between lock-free iteration and snapshot consistency.

ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -48,53 +76,249 @@ class TaskManagerBase {
...
+        std::atomic_store_explicit(&slot.ctx, std::move(ptr), order);
+    }
+
+    std::uint64_t PackFreeListHead(std::uint64_t generation, std::size_t index) const
+    {
+        return (generation << freeListShift_) | static_cast<std::uint64_t>(index);
+    }
+
+    std::size_t FreeListPop()
+    {
```
nrj868: Use `ucm/shared/infra/thread/index_pool.h` for slot alloc can decrease code complexity

ucm/transport/kv/asu/test/case/task_manager_base_slot_test.cc
```diff
@@ -0,0 +1,673 @@
...
+#include <vector>
+#include "client_task_manager.h"
+#include "transport_task_manager.h"
+
+namespace UC::ASU {
+namespace {
+
+// ==================== Transport TaskManager Tests ====================
+
+class TransportTaskManagerTest : public ::testing::Test {
```
nrj868, line 40: Add some bench test for `Submit` `Get` `Remove` to measure actual performance of different TaskMgr impl

ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -48,53 +76,249 @@ class TaskManagerBase {
...
+        auto& slot = slots_[slotIndex];
+
+        std::uint8_t expected = SlotState::READY;
+        if (!slot.state.compare_exchange_strong(expected, SlotState::REMOVING,
+                                                std::memory_order_acq_rel,
+                                                std::memory_order_acquire)) {
+            return Status::Error(StatusCode::TASK_NOT_FOUND, taskName_ + " task not found");
+        }
+
+        if (slot.taskId.load(std::memory_order_acquire) != taskId) {
```
nrj868: Should check `taskId` before CAS `state`?

e -i HEAD~2
```diff
@@ -0,0 +1,577 @@
+[33mca15956[m[33m ([m[1;36mHEAD[m[33m -> [m[1;32mtask_manager_w_indexpool[m[33m, [m[1;31mcaz/task_manager_w_indexpool[m[33m)[m reuse IndexPool for task slot free list
```
Infinite666: Please remove the wrong file.

ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -48,53 +70,197 @@ class TaskManagerBase {
...
+        std::uint8_t expected = SlotState::EMPTY;
+        if (!slot.state.compare_exchange_strong(expected, SlotState::WRITING,
+                                                std::memory_order_acq_rel,
+                                                std::memory_order_acquire)) {
+            freeList_.Release(acquiredIndex);
+            taskId = kInvalidTaskId;
+            return Status::Error(StatusCode::INTERNAL_ERROR, taskName_ + " task slot not empty");
+        }
+
+        const auto generation = slot.generation.fetch_add(1, std::memory_order_relaxed) + 1;
```
Infinite666: What's the usage of generation? It will be more simple to just use slot index as task id. If we want to avoid ABA problem in CAS, index_pool.h already have version to prevent.

AooooooA-C: In concurrency, can other threads still retain the old taskID after one thread remove it? 
For example:  multi-threaded concurrency -> after Thread A removes Task 1, Slot 1 becomes free. Concurrently, Thread B acquires Slot 1 to insert a new task and gets task ID 1. If Thread C holding the stale task ID 1 calls Get , it will retrieve the newly created task by mistake.

e -i HEAD~2
AooooooA-C: Removed.

ucm/transport/kv/asu/common/task_manager_base.h
AooooooA-C: YES, Remove proceeds only if the slot currently belongs to the target taskId, so should check `taskId` . If the slot has already been reused for a new task and the taskId mismatches, returns TASK_NOT_FOUND directly without modifying the new task's state.

ucm/transport/kv/asu/common/task_manager_base.h
ygwpz: 💡 **Suggestion:** Using `memory_order_relaxed` for generation increment may cause visibility issues on weak memory ordering architectures (e.g., ARM). Consider using `memory_order_acq_rel` to ensure the generation increment is visible to other threads before they read the slot state. Alternatively, document why relaxed ordering is safe here (e.g., state CAS provides necessary synchronization).

ucm/transport/kv/asu/common/task_manager_base.h
```diff
@@ -48,53 +70,197 @@ class TaskManagerBase {
...
+        }
+
+        auto expectedTaskId = taskId;
+        if (!slot.taskId.compare_exchange_strong(expectedTaskId, kInvalidTaskId,
+                                                 std::memory_order_acq_rel,
+                                                 std::memory_order_acquire)) {
+            return Status::Error(StatusCode::TASK_NOT_FOUND, taskName_ + " task not found");
+        }
+
+        slot.state.store(SlotState::REMOVING, std::memory_order_release);
```
ygwpz: 💡 **Suggestion:** The REMOVING → EMPTY state transition happens in two steps. Consider documenting whether concurrent Get() operations might see REMOVING state and correctly return nullptr, or if this could cause issues with slot reuse timing.

nrj868 (approved):

nrj868 (approved):
Nice

PR #982, opened 2026-05-28
[opt]Add delay wait tasks for direct connector

qyh111:

### DIFF
--- a/ucm/integration/vllm/device.py
+++ b/ucm/integration/vllm/device.py
@@ -41,6 +41,10 @@ class Device(ABC):
         pass
 
     @abstractmethod
+    def destroy_event_handle(self, event_handle: int):
+        pass
+
+    @abstractmethod
     def get_cpu_affinity(self, local_rank: int) -> Optional[str]:
         """
         Return CPU affinity as a cpulist string, e.g. "0-43,88-131".
@@ -117,6 +121,13 @@ class CudaDevice(Device):
 
     def destroy_event_handles(self):
         self.events.clear()
+
+    def destroy_event_handle(self, event_handle: int):
+        self.events = [
+            event
+            for event in self.events
+            if int(getattr(event, "cuda_event", 0)) != event_handle
+        ]
 
     def get_cpu_affinity(self, local_rank: int) -> Optional[str]:
         """
@@ -235,6 +246,8 @@ class NpuDevice(Device):
             ret = acl.rt.record_event(event, stream)
             if ret != 0:
                 logger.error(f"acl record_event failed: {ret}")
+                acl.rt.destroy_event(event)
+                self.events.remove(event)
                 return 0
             handle = int(event)
             if not handle:
@@ -256,6 +269,19 @@ class NpuDevice(Device):
             except Exception as e:
                 logger.error(f"destroy npu event failed. {e}")
         self.events.clear()
+
+    def destroy_event_handle(self, event_handle: int):
+        import acl
+
+        for event in list(self.events):
+            if int(event) != event_handle:
+                continue
+            try:
+                acl.rt.destroy_event(event)
+            except Exception as e:
+                logger.error(f"destroy npu event failed. {e}")
+            self.events.remove(event)
+            break
 
     def _execute_command(self, cmd_list: List[str]) -> str:
         try:
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -1145,13 +1145,15 @@ class UCMLayerWiseConnector(UCMDirectConnector):
         metadata = self._get_connector_metadata()
 
         total_ucm_block_ids, total_vllm_block_ids = [], []
+        dump_request_ids: set[str] = set()
         layer_id = self.layer_name_to_id[layer_name]
         local_layer_id = layer_id - self.first_layer_id
-        for _, request in metadata.request_meta.items():
+        for request_id, request in metadata.request_meta.items():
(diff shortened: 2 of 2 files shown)


ucm/integration/vllm/device.py
```diff
@@ -118,6 +122,13 @@ def synchronize(self):
     def destroy_event_handles(self):
         self.events.clear()
 
+    def destroy_event_handle(self, event_handle: int):
```
ygwpz, line 125: **Resource Leak**: The `destroy_event_handle` method for CudaDevice filters out the event from the list but never actually destroys the underlying CUDA event resource. Compare this to `NpuDevice.destroy_event_handle` which calls `acl.rt.destroy_event(event)`. For proper CUDA resource management, the event should be destroyed before being removed from the list.

ucm/integration/vllm/device.py
```diff
@@ -235,6 +246,8 @@ def get_event_handle(self) -> int:
             ret = acl.rt.record_event(event, stream)
             if ret != 0:
                 logger.error(f"acl record_event failed: {ret}")
+                acl.rt.destroy_event(event)
+                self.events.remove(event)
                 return 0
```
ygwpz, line 251: **Error Handling Issue**: After `acl.rt.record_event` fails, the code calls `self.events.remove(event)`. If this raises a `ValueError` (event not in list), the exception will propagate and the error logging will be incomplete. Consider wrapping the remove in a try-except or checking membership first.

ucm/integration/vllm/device.py
```diff
@@ -257,6 +270,19 @@ def destroy_event_handles(self):
...
 
+    def destroy_event_handle(self, event_handle: int):
+        import acl
+
+        for event in list(self.events):
+            if int(event) != event_handle:
+                continue
+            try:
+                acl.rt.destroy_event(event)
+            except Exception as e:
```
ygwpz, line 281: **State Consistency Issue**: The `self.events.remove(event)` call is outside the try-except block. If it raises a `ValueError`, the event has already been destroyed, leaving the system in an inconsistent state. Move the remove inside the try block.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -804,19 +823,83 @@ def save_kv_layer(
     ) -> None:
         pass
 
+    def _wait_pending_dump_task(self, pending_dump_task: PendingDumpTask) -> None:
+        try:
+            self.store.wait(pending_dump_task.task)
+        finally:
+            self._release_dump_event_handle(pending_dump_task)
```
ygwpz, line 828: **Potential State Inconsistency**: After the `finally` block releases the event handle, `_mark_pending_dump_task_done` is called outside the protected block. If this method raises an exception, the pending counts could become inconsistent. Consider wrapping the entire operation in a try-finally.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -804,19 +823,83 @@ def save_kv_layer(
...
+                self._mark_pending_dump_task_done(pending_dump_task)
+        self._pending_dump_tasks = remaining_tasks
+
+    def _flush_pending_dump_tasks(self, request_ids: Optional[set[str]] = None) -> None:
+        pending_dump_tasks = self._pending_dump_tasks
+        self._pending_dump_tasks = []
+        remaining_tasks: list[PendingDumpTask] = []
+        for pending_dump_task in pending_dump_tasks:
+            if request_ids is not None and not (
+                pending_dump_task.request_ids & request_ids
```
ygwpz, line 845: **CRITICAL: Resource Leak**: When `store.check()` throws an exception, only `_mark_pending_dump_task_done` is called but `_release_dump_event_handle` is NEVER called. This leaks the event handle (GPU/NPU resource). Fix: add `self._release_dump_event_handle(pending_dump_task)` before `_mark_pending_dump_task_done`.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -896,12 +965,64 @@ def get_block_ids_with_load_errors(self) -> set[int]:
...
+                        )
+                        self._mark_pending_dump_task_done(pending_dump_task)
+                else:
+                    remaining_tasks.append(pending_dump_task)
+            self._pending_dump_tasks = remaining_tasks
+
+        finished_sending = {
+            request_id
+            for request_id in self._delayed_finished_req_ids
+            if request_id not in self._pending_dump_count_by_req
```
ygwpz: **CRITICAL: Resource Leak**: Same issue as above. When `store.wait()` throws an exception in `get_finished`, the event handle is not released. Add `_release_dump_event_handle(pending_dump_task)` before `_mark_pending_dump_task_done`.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -804,19 +823,83 @@ def save_kv_layer(
...
+    def _poll_pending_dump_tasks(self) -> None:
+        remaining_tasks: list[PendingDumpTask] = []
+        for pending_dump_task in self._pending_dump_tasks:
+            try:
+                if self.store.check(pending_dump_task.task):
+                    self._wait_pending_dump_task(pending_dump_task)
+                else:
+                    remaining_tasks.append(pending_dump_task)
+            except Exception as e:
+                logger.error(f"check dump kv cache failed. {type(e).__name__}: {e}")
```
ygwpz: **CRITICAL: Resource Leak**: When `store.wait()` throws an exception, the event handle is not released. Add `self._release_dump_event_handle(pending_dump_task)` before `_mark_pending_dump_task_done`.

PR #989, opened 2026-06-01
[opt]Defer UCM KV Dump Waiting Until Request Completion

qyh111:
## Purpose

Move UCM KV dump completion waiting out of `wait_for_save()` so model execution is no longer blocked by async dump tasks, while still preserving delayed block release correctness for finished requests.

This also fixes the non-HMA `MultiConnector` path by forwarding `request_finished()` from the outer `UCMConnector` to the actual inner connector, ensuring requests that UCM saves asynchronously are properly marked for delayed free before they are returned from `get_finished()`.

## Modifications

- Added async dump task tracking for UCM Direct and LayerWise connectors.
- Changed `wait_for_save()` to submit dump tasks without blocking on task completion.
- Moved dump completion waiting to `get_finished()` for requests that have finished and require delayed block release.
- Added preemption handling to wait only for pending dump tasks related to preempted requests.
- Added per-task event handle release after dump task completion or submission failure.
- Removed save-duration metrics from the delayed `wait_for_save()` path because the timing is no longer accurate after deferring completion.
- Updated LayerWise save flow to use the same delayed dump completion mechanism.
- Added `UCMConnector.request_finished()` forwarding for the non-HMA path so `MultiConnector` can correctly count UCM async saves.
- Rebased the branch on latest `origin/develop` and resolved conflicts with the new layerwise / pipeline store metrics changes.
## Test
Tested on 8*NPU on A3 for Minimax2.7
direct connector
16K concurrancy 8 TTFT decrease(3.6605->3.2271)
64K concurrancy 4 TTFT decrease(13.0390->11.4577)
layerwise connector
slight improve almost the same

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -259,6 +259,14 @@ class KVCacheLayout:
 @dataclass
 class UCMConnectorMetadata(KVConnectorMetadata):
     request_meta: dict[str, RequestDispatchMeta] = field(default_factory=dict)
+    preempted_req_ids: set[str] = field(default_factory=set)
+
+
+@dataclass
+class PendingDumpTask:
+    task: Task
+    request_ids: set[str]
+    event_handle: int = 0
 
 
 class RequestHasher:
@@ -407,6 +415,10 @@ class UCMDirectConnector(KVConnectorBase_V1):
 
         # invalid block ids due to load errors
         self._invalid_block_ids: set[int] = set()
+        self._async_dump_req_ids: set[str] = set()
+        self._pending_dump_tasks: list[PendingDumpTask] = []
+        self._pending_dump_count_by_req: dict[str, int] = defaultdict(int)
+        self._delayed_finished_req_ids: set[str] = set()
         self.cp_world_size = 1
         self.hash_block_size = self.block_size
         self.block_size *= self.cp_world_size
@@ -743,7 +755,14 @@ class UCMDirectConnector(KVConnectorBase_V1):
         for request_id in scheduler_output.finished_req_ids:
             self.requests_meta.pop(request_id, None)
 
-        return UCMConnectorMetadata(requests_dispatch_meta)
+        for request_id, dispatch_meta in requests_dispatch_meta.items():
+            if len(dispatch_meta.dump_block_ids[0]) > 0:
+                self._async_dump_req_ids.add(request_id)
+
+        return UCMConnectorMetadata(
+            requests_dispatch_meta,
+            scheduler_output.preempted_req_ids or set(),
+        )
 
     def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
         metadata = self._get_connector_metadata()
@@ -845,19 +864,83 @@ class UCMDirectConnector(KVConnectorBase_V1):
     ) -> None:
         pass
 
+    def _wait_pending_dump_task(self, pending_dump_task: PendingDumpTask) -> None:
+        try:
+            self.store.wait(pending_dump_task.task)
+        finally:
+            self._release_dump_event_handle(pending_dump_task)
+        self._mark_pending_dump_task_done(pending_dump_task)
+
+    def _release_dump_event_handle(self, pending_dump_task: PendingDumpTask) -> None:
+        if (
+            self.enable_event_sync
+            and pending_dump_task.event_handle
+            and self.device is not None
+        ):
+            self.device.destroy_event_handle(pending_dump_task.event_handle)
+            pending_dump_task.event_handle = 0
+
+    def _mark_pending_dump_task_done(self, pending_dump_task: PendingDumpTask) -> None:
+        for request_id in pending_dump_task.request_ids:
+            pending_count = self._pending_dump_count_by_req.get(request_id, 0)
+            if pending_count <= 1:
+                self._pending_dump_count_by_req.pop(request_id, None)
+            else:
+                self._pending_dump_count_by_req[request_id] = pending_count - 1
+
+    def _poll_pending_dump_tasks(self) -> None:
+        remaining_tasks: list[PendingDumpTask] = []
+        for pending_dump_task in self._pending_dump_tasks:
+            try:
+                if self.store.check(pending_dump_task.task):
+                    self._wait_pending_dump_task(pending_dump_task)
+                else:
+                    remaining_tasks.append(pending_dump_task)
+            except Exception as e:
+                logger.error(f"check dump kv cache failed. {type(e).__name__}: {e}")
+                self._mark_pending_dump_task_done(pending_dump_task)
+        self._pending_dump_tasks = remaining_tasks
+
+    def _flush_pending_dump_tasks(self, request_ids: Optional[set[str]] = None) -> None:
+        pending_dump_tasks = self._pending_dump_tasks
+        self._pending_dump_tasks = []
+        remaining_tasks: list[PendingDumpTask] = []
+        for pending_dump_task in pending_dump_tasks:
+            if request_ids is not None and not (
+                pending_dump_task.request_ids & request_ids
+            ):
+                remaining_tasks.append(pending_dump_task)
+                continue
+            try:
+                self._wait_pending_dump_task(pending_dump_task)
+            except Exception as e:
+                logger.error(f"wait for dump kv cache failed. {type(e).__name__}: {e}")
+                self._mark_pending_dump_task_done(pending_dump_task)
+        self._pending_dump_tasks = remaining_tasks
+
+    def handle_preemptions(self, kv_connector_metadata: KVConnectorMetadata):
+        preempted_req_ids = getattr(kv_connector_metadata, "preempted_req_ids", None)
+        if preempted_req_ids:
+            self._flush_pending_dump_tasks(preempted_req_ids)
+
     def wait_for_save(self) -> None:
         # TODO support PP
+        metadata = self._get_connector_metadata()
+        assert isinstance(metadata, UCMConnectorMetadata)
+        metadata_dump_request_ids = {
+            request_id
+            for request_id, request in metadata.request_meta.items()
+            if len(request.dump_block_ids[0]) > 0
+        }
+        self._async_dump_req_ids.update(metadata_dump_request_ids)
+
         if self.is_mla and self.tp_rank != 0:
             return
 
-        metadata = self._get_connector_metadata()
-        assert isinstance(metadata, UCMConnectorMetadata)
-
         dump_tasks: List[Task] = []
         is_save = False
-        num_saved_block = 0
-        num_saved_request = 0
         total_ucm_block_ids, total_vllm_block_ids = [], []
+        dump_request_ids: set[str] = set()
         for request_id, request in metadata.request_meta.items():
             if len(request.dump_block_ids[0]) == 0:
                 continue
@@ -872,8 +955,7 @@ class UCMDirectConnector(KVConnectorBase_V1):
                 if len(ucm_block_ids) == 0:
                     continue
             is_save = True
-            num_saved_block += len(ucm_block_ids)
-            num_saved_request += 1
+            dump_request_ids.add(request_id)
             if self.tp_rank != 0:
                 for i, ucm_block_id in enumerate(ucm_block_ids):
                     ucm_block_ids[i] = self.request_hasher(ucm_block_id)
@@ -881,6 +963,7 @@ class UCMDirectConnector(KVConnectorBase_V1):
             total_vllm_block_ids.extend(vllm_block_ids)
 
         if is_save:
+            event_handle = 0
             try:
                 total_ptrs = self.kv_cache_layout.extract_block_addrs(
                     total_vllm_block_ids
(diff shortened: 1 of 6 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -801,19 +818,60 @@ def save_kv_layer(
...
-            return
-
         metadata = self._get_connector_metadata()
         assert isinstance(metadata, UCMConnectorMetadata)
+        metadata_dump_request_ids = {
+            request_id
+            for request_id, request in metadata.request_meta.items()
+            if len(request.dump_block_ids[0]) > 0
+        }
+        self._async_dump_req_ids.update(metadata_dump_request_ids)
```
ygwpz, line 948: 🔴 **Logic Issue:** `_async_dump_req_ids.update()` happens before the MLA rank check at line 868. For MLA non-rank-0 processes, the function returns early without actually performing the dump, but the request IDs are still tracked in `_async_dump_req_ids`. This means `request_finished()` will return `(True, None)` for requests that never had their dump task created, potentially causing incorrect delayed-free behavior.

Consider moving the `update()` call after the MLA check, or clearing the tracked IDs for MLA non-rank-0 cases.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -801,19 +818,60 @@ def save_kv_layer(
...
+            self._release_dump_event_handle(pending_dump_task)
+
+    def _release_dump_event_handle(self, pending_dump_task: PendingDumpTask) -> None:
+        if (
+            self.enable_event_sync
+            and pending_dump_task.event_handle
+            and self.device is not None
+        ):
+            self.device.destroy_event_handle(pending_dump_task.event_handle)
+            pending_dump_task.event_handle = 0
```
ygwpz, line 886: 💡 **Side Effect:** Modifying `pending_dump_task.event_handle = 0` mutates the passed `PendingDumpTask` instance. While this works in Python (dataclasses are mutable by default), this side effect should be documented in the method signature or consider using a frozen dataclass to make the mutation explicit through a return value.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -890,12 +935,48 @@ def get_block_ids_with_load_errors(self) -> set[int]:
         self._invalid_block_ids = set()
         return res
 
+    def request_finished(
+        self,
+        request: "Request",
+        block_ids: list[int],
+    ) -> tuple[bool, dict[str, Any] | None]:
+        if request.request_id in self._async_dump_req_ids:
```
ygwpz, line 1043: ⚠️ **Potential Issue:** `request_finished()` returns `(True, None)` when a request is in `_async_dump_req_ids`, indicating delayed free is needed. However, the actual dump task completion is handled separately in `get_finished()`. If there's a timing gap between these calls, the block release might happen before the dump is truly complete.

Consider whether the semantics should be: (1) mark for delayed free immediately, or (2) only mark after dump completes. The current implementation assumes callers will call `get_finished()` before releasing blocks.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -890,12 +935,48 @@ def get_block_ids_with_load_errors(self) -> set[int]:
...
+    def get_finished(
+        self,
+        finished_req_ids: set[str],
+    ) -> tuple[Optional[set[str]], Optional[set[str]]]:
+        async_finished_req_ids = finished_req_ids & self._async_dump_req_ids
+
+        if async_finished_req_ids:
+            remaining_tasks: list[PendingDumpTask] = []
+            for pending_dump_task in self._pending_dump_tasks:
+                if pending_dump_task.request_ids & async_finished_req_ids:
```
ygwpz, line 1066: ⚠️ **Partial Task Handling:** When a `PendingDumpTask` has `request_ids` that partially intersect with `async_finished_req_ids` (i.e., some but not all requests in the task are finished), the entire task is waited on and then NOT added to `remaining_tasks`. This means other unfinished requests in the same task lose their pending dump tracking.

Consider whether the intended behavior is: (1) wait for the entire task if any request finishes, or (2) track unfinished requests separately. If option (1), this is correct. If option (2), the logic needs adjustment.

ucm/integration/vllm/ucm_connector.py
qyh111: MLA 非 rank0 也要上报 completion，因为 KVOutputAggregator 默认等 world_size 个 worker 完成信号

ucm/integration/vllm/ucm_connector.py
qyh111: event_handle = 0 是防止重复 destroy 的本地清理副作用

ucm/integration/vllm/ucm_connector.py
qyh111: request_finished=True 本来就是 vLLM delayed-free 合约，不是等 dump 完才返回 True

ucm/integration/vllm/ucm_connector.py
qyh111: 一个 PendingDumpTask 是批量 dump task，wait 一次代表 task 内所有 request 的 dump 都完成；移出 pending 是合理的

dante159753:
I found one issue in the CP path: `UCMCPConnector.build_connector_meta()` does not mirror `UCMDirectConnector.build_connector_meta()` by adding requests with `dump_block_ids` to the scheduler-side `_async_dump_req_ids` set. As a result, `request_finished()` returns `False` in the CP path, so vLLM may free the blocks immediately while the worker-side dump task has been deferred until `get_finished()`. That leaves a window where the blocks can be reused or overwritten before the dump wait completes. Please mirror the DirectConnector tracking logic in this override before returning the metadata.

qyh111:
> I found one issue in the CP path: `UCMCPConnector.build_connector_meta()` does not mirror `UCMDirectConnector.build_connector_meta()` by adding requests with `dump_block_ids` to the scheduler-side `_async_dump_req_ids` set. As a result, `request_finished()` returns `False` in the CP path, so vLLM may free the blocks immediately while the worker-side dump task has been deferred until `get_finished()`. That leaves a window where the blocks can be reused or overwritten before the dump wait completes. Please mirror the DirectConnector tracking logic in this override before returning the metadata.

I think this does not apply to the CP path. UCMCPConnector does not override build_connector_meta(); its MRO is UCMCPConnector -> UCMLayerWiseConnector -> UCMDirectConnector, and UCMLayerWiseConnector does not override build_connector_meta either. Therefore CP uses UCMDirectConnector.build_connector_meta(), which already adds dump requests to scheduler-side _async_dump_req_ids before returning metadata.

ucm/integration/vllm/ucm_connector.py
ygwpz, line 948: 💡 **Suggestion:** Consider updating `_async_dump_req_ids` before the MLA early return check. If MLA non-rank0 returns early, it might miss tracking request IDs that need dump completion tracking.

qyh111: This is already the current behavior. `_async_dump_req_ids.update(metadata_dump_request_ids)` is intentionally placed before the MLA non-rank-0 early return.

In MLA, only rank 0 submits the actual dump task, but non-rank-0 workers still need to report local completion so `KVOutputAggregator` can collect `world_size` completion signals before forwarding `finished_sending` to the scheduler. If we move the update after the early return, non-rank-0 workers would not report completion and delayed block release could wait indefinitely.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -801,19 +828,69 @@ def save_kv_layer(
...
+        for pending_dump_task in pending_dump_tasks:
+            if request_ids is not None and not (
+                pending_dump_task.request_ids & request_ids
+            ):
+                remaining_tasks.append(pending_dump_task)
+                continue
+            try:
+                self._wait_pending_dump_task(pending_dump_task)
+            except Exception as e:
+                logger.error(f"wait for dump kv cache failed. {type(e).__name__}: {e}")
```
ygwpz, line 922: ⚠️ **Resource Leak:** When an exception occurs in `_wait_pending_dump_task`, the event handle is not released. The `_release_dump_event_handle` is only called inside `_wait_pending_dump_task` after `store.wait()` succeeds. If `store.wait()` raises an exception, the event handle remains in `pending_dump_task.event_handle` and is never cleaned up. Consider adding cleanup in the except block: `self._release_dump_event_handle(pending_dump_task)`

ucm/integration/vllm/ucm_connector.py
```diff
@@ -801,19 +828,69 @@ def save_kv_layer(
...
+    def wait_for_save(self) -> None:
+        # TODO support PP
         metadata = self._get_connector_metadata()
         assert isinstance(metadata, UCMConnectorMetadata)
+        metadata_dump_request_ids = {
+            request_id
+            for request_id, request in metadata.request_meta.items()
+            if len(request.dump_block_ids[0]) > 0
+        }
+        self._async_dump_req_ids.update(metadata_dump_request_ids)
```
ygwpz, line 948: ⚠️ **State Accumulation:** `_async_dump_req_ids` is updated but never cleared. Over time, this set could accumulate stale request IDs from completed or preempted requests. Consider clearing this set when requests are finished, or in `_flush_pending_dump_tasks` when flushing matching request IDs.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -801,19 +828,69 @@ def save_kv_layer(
...
+                logger.error(f"wait for dump kv cache failed. {type(e).__name__}: {e}")
+        self._pending_dump_tasks = remaining_tasks
+
+    def handle_preemptions(
+        self,
+        kv_connector_metadata: KVConnectorMetadata | set[str],
+    ) -> None:
+        if self.use_layerwise:
             return
+        preempted_req_ids = (
```
ygwpz, line 929: 💡 **Type Safety:** Using `getattr(kv_connector_metadata, 'preempted_req_ids', None)` assumes the metadata object has this attribute. If the metadata type changes, this could silently return `None` instead of raising an error. Consider explicit type checking or adding a `hasattr` check with logging if the attribute is missing.

ucm/integration/vllm/ucm_connector.py
qyh111: _wait_pending_dump_task() 已使用 finally：

try:
    self.store.wait(task)
finally:
    self._release_dump_event_handle(...)
因此 wait() 成功、task 失败或抛异常都会释放 event。

ucm/integration/vllm/ucm_connector.py
qyh111: 当前写法是为了兼容两种输入：

KVConnectorMetadata | set[str]
MultiConnector 和不同 vLLM 版本也可能采用不同包装方式。这里静默返回 None 只是跳过无 preemption 的处理，不构成实际 bug。

ucm/integration/vllm/ucm_connector.py
qyh111: 它会在两处清理：

Scheduler：request_finished() 中 discard()
Worker：get_finished() 中 difference_update()
Preemption 时不清理是合理的，因为请求没有结束，之后可能恢复执行。最终完成或取消进入 finished 生命周期时会清理。

ucm/integration/vllm/ucm_connector.py
```diff
@@ -872,52 +978,39 @@ def wait_for_save(self) -> None:
...
                 dump_tasks.append(task)
             except Exception as e:
                 logger.error(f"dump kv cache failed. {type(e).__name__}: {e}")
+                if self.enable_event_sync and event_handle and self.device is not None:
+                    self.device.destroy_event_handle(event_handle)
                 return
 
-            try:
-                for task in dump_tasks:
-                    self.store.wait(task)
```
ygwpz, line 902: 🔴 **Resource Leak on Check Failure:** When `store.check()` raises an exception, the `pending_dump_task` is added to `remaining_tasks` without releasing its `event_handle`. This causes:
1. Event handle leak (GPU/NPU resources not freed)
2. Broken task remains in `_pending_dump_tasks` indefinitely
3. Subsequent `_poll_pending_dump_tasks` calls keep encountering the same broken task

**Recommendation:** Either release the event handle in the except block, or remove the task entirely instead of keeping it in `remaining_tasks`. Example fix:
```python
except Exception as e:
    logger.error(f"check dump kv cache failed. {type(e).__name__}: {e}")
    # Release event handle to prevent resource leak
    self._release_dump_event_handle(pending_dump_task)
    # Don't add to remaining_tasks - remove broken task entirely
    continue
```

ucm/integration/vllm/ucm_connector.py
```diff
@@ -2652,14 +2820,12 @@ def build_connector_meta(
         for request_id in scheduler_output.finished_req_ids:
             self.requests_meta.pop(request_id, None)
 
-        return UCMConnectorMetadata(requests_dispatch_meta)
+        self._track_async_dump_requests(requests_dispatch_meta)
 
-    def request_finished_all_groups(
```
ygwpz, line 2657: ⚠️ **Partial Task Handling Logic:** When `pending_dump_task.request_ids` has partial intersection with `async_finished_req_ids` (some requests finished, some not), the entire task is waited on. If `_wait_pending_dump_task` fails with an exception here:
1. The event handle IS correctly released (in finally block of `_wait_pending_dump_task`)
2. BUT the unfinished requests in the same task lose their dump completion tracking
3. These requests won't have their dump waited for when they finish later

**Question:** Is this intentional design (wait entire task if any request finishes)? If yes, document this behavior. If no, consider tracking unfinished requests separately after partial wait failure.

ygwpz (commented):
## Follow-up Review (2026-06-13)

I've reviewed the new commits pushed today. The previous concerns from 2026-06-07 have been addressed:

✅ **Event handle release** (line 930): The `finally` block in `_wait_pending_dump_task` ensures cleanup on all paths.

✅ **State accumulation** (line 958): `_async_dump_req_ids` is properly cleared in `request_finished()` (discard) and `get_finished()` (difference_update).

✅ **Type safety** (line 939): The `isinstance` check before `getattr` properly handles both `KVConnectorMetadata | set[str]` types.

However, I found **2 NEW issues** that need attention:

See inline comments below for details on:
1. Resource leak when `store.check()` fails in `_poll_pending_dump_tasks`
2. Partial task handling logic in `get_finished`

Overall, the deferred wait mechanism is well-designed, but these edge cases need handling to prevent resource leaks and ensure correctness.

ucm/integration/vllm/ucm_connector.py
qyh111: Reviewer 的修复建议不合适。check() 只是查询，异常不代表异步任务已经结束，此时释放 event 可能过早。当前保留 task，之后在 get_finished()/handle_preemptions() 中调用 wait()，并通过 finally 释放 event，是正确的。
唯一风险是异常期间可能重复打印日志，但不是资源泄漏。

ucm/integration/vllm/ucm_connector.py
qyh111: 批量 task 部分 request 完成，wait() 异常后丢失跟踪
也不是实际问题。一个 PendingDumpTask 对应一个批量 dump task；PipelineStore 的 Wait() 会先消费 task handle，并等待/排空到终态后才返回成功或异常。因此无论成功还是失败，该批次都已终止，不能也不需要为剩余 request 再次等待。

flesher0813 (approved):

harrisonyhq (approved):

PR #990, opened 2026-06-01
[WIP] [Feat] Ffts D2D transport dev

pyxyzc:
## Purpose
Introduce an Ascend FFTS Plus based D2D transport module for P2P cache data movement, providing an asynchronous copy interface backed by ACL stream execution and FFTS SDMA task submission.

## Modifications
- Add `ucm_transport_ffts` static library under `ucm/transport/p2p/ffts`.
- Add CMake discovery and linkage for Ascend CANN headers/libraries, including `ascendcl`, `runtime`, FFTS runtime headers, profiling headers, and toolchain headers.
- Add public `FftsTransport` API with `Setup`, `WaitEvent`, `CopyAsync`, batched `Submit`, and `Synchronize`.
- Implement FFTS engine logic to:
  - create and manage an ACL fast-launch/fast-sync stream;
  - convert copy descriptors into FFTS Plus SDMA contexts;
  - split large batches by the FFTS ready-context limit;
  - filter no-op copies and validate invalid descriptors;
  - keep submitted context buffers alive until stream synchronization.

## Test
 FFTS transport tests covering:

  - **Copy before setup**: verifies copy requests fail before `Setup()` is called.
  - **Single device copy**: verifies one device-to-device copy preserves the source data.
  - **Batched copies**: verifies multiple chunked copy descriptors are submitted and completed correctly.
  - **Large batch split**: verifies large submissions are split internally and all chunks are copied.
  - **Two-stage pipeline**: verifies data can be copied through an intermediate device buffer.
  - **Invalid/no-op inputs**: verifies zero-size operations succeed as no-ops and invalid non-zero operations fail.

### DIFF
--- a/ucm/transport/p2p/ffts/bench/ffts_bench_options.cpp
+++ b/ucm/transport/p2p/ffts/bench/ffts_bench_options.cpp
@@ -1,0 +1,144 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2026 Huawei Technologies Co., Ltd. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ * */
+#include "ffts_bench_options.h"
+
+#include <cctype>
+#include <cstdlib>
+#include <iostream>
+#include <limits>
+#include <stdexcept>
+#include <string>
+
+namespace UC::Transport::Ffts::Bench {
+namespace {
+size_t ParseBytes(const std::string& value)
+{
+    if (value.empty()) { throw std::invalid_argument("empty size value"); }
+
+    std::string number = value;
+    size_t multiplier = 1;
+    const auto suffix = static_cast<char>(std::tolower(static_cast<unsigned char>(value.back())));
+    if (suffix == 'k' || suffix == 'm' || suffix == 'g') {
+        number = value.substr(0, value.size() - 1);
+        multiplier = suffix == 'k' ? kKiB : (suffix == 'm' ? kMiB : kGiB);
+    }
+
+    size_t consumed = 0;
+    const auto parsed = std::stoull(number, &consumed, 0);
+    if (consumed != number.size()) { throw std::invalid_argument("invalid size value: " + value); }
+    if (parsed > std::numeric_limits<size_t>::max() / multiplier) {
+        throw std::overflow_error("size value overflows: " + value);
+    }
+    return static_cast<size_t>(parsed) * multiplier;
+}
+
+size_t ParseCount(const std::string& value)
+{
+    size_t consumed = 0;
+    const auto parsed = std::stoull(value, &consumed, 0);
+    if (consumed != value.size()) { throw std::invalid_argument("invalid count value: " + value); }
+    return static_cast<size_t>(parsed);
+}
+
+std::vector<size_t> ParseList(const std::string& value, bool parseBytes)
+{
+    std::vector<size_t> result;
+    size_t begin = 0;
+    while (begin <= value.size()) {
+        const auto end = value.find(',', begin);
+        const auto token = value.substr(begin, end == std::string::npos ? std::string::npos : end - begin);
+        if (token.empty()) { throw std::invalid_argument("empty list item: " + value); }
+        result.push_back(parseBytes ? ParseBytes(token) : ParseCount(token));
+        if (end == std::string::npos) { break; }
+        begin = end + 1;
+    }
+    return result;
+}
+
+Scenario ParseScenario(const std::string& value)
+{
+    if (value == "all") { return Scenario::All; }
+    if (value == "single") { return Scenario::Single; }
+    if (value == "batch") { return Scenario::Batch; }
+    throw std::invalid_argument("invalid scenario: " + value);
+}
+
+void PrintUsage(const char* program)
+{
+    std::cout << "Usage: " << program << " [options]\n"
+              << "Options:\n"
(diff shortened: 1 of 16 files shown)


ucm/transport/p2p/ffts/src/ffts_engine.cpp
```diff
@@ -0,0 +1,263 @@
...
+           kSourceNonSecure | kDestinationNonSecure;
+}
+
+void FillSdmaContext(rtFftsPlusComCtx_t& storage, const CopyDesc& copy)
+{
+    static_assert(sizeof(rtFftsPlusSdmaCtx_t) <= sizeof(rtFftsPlusComCtx_t),
+                  "FFTS SDMA context must fit in common context storage");
+    std::memset(&storage, 0, sizeof(storage));
+
+    auto& ctx = *reinterpret_cast<rtFftsPlusSdmaCtx_t*>(&storage);
```
ygwpz, line 86: ⚠️ **Type Safety:** The `static_assert` only verifies size compatibility between `rtFftsPlusComCtx_t` and `rtFftsPlusSdmaCtx_t`, but doesn't check alignment or actual field layout compatibility. If the structs have different padding or field ordering, this `reinterpret_cast` could corrupt data.

Consider adding a compile-time check that the first field of `rtFftsPlusComCtx_t` is compatible with `contextType`, or document why this reinterpret_cast is safe for the specific Ascend SDK version.

ucm/transport/p2p/ffts/src/ffts_engine.cpp
```diff
@@ -0,0 +1,263 @@
...
+ * @return Status::OK() on success, otherwise an error status.
+ */
+Status FftsEngine::Setup(int32_t deviceId)
+{
+    if (ready_) {
+        if (deviceId_ == deviceId) { return Status::OK(); }
+        return Status::InvalidParam("FFTS transport is already setup with a different device");
+    }
+
+    auto status = AclStatus(aclrtSetDevice(deviceId), "aclrtSetDevice");
```
ygwpz, line 142: 💡 **Suggestion:** Consider validating `deviceId` is non-negative before calling `aclrtSetDevice`. While ACL will reject invalid IDs, an upfront check provides clearer error messages and prevents potential confusion with negative device IDs.

ucm/transport/p2p/ffts/src/ffts_engine.cpp
```diff
@@ -0,0 +1,263 @@
...
+    if (activeCopies.empty()) { return Status::OK(); }
+
+    // Build one SDMA context for each valid copy descriptor.
+    auto contexts = std::make_shared<ContextBuffer>(activeCopies.size());
+    for (size_t i = 0; i < activeCopies.size(); ++i) {
+        FillSdmaContext((*contexts)[i], activeCopies[i]);
+    }
+
+    // Build the FFTS Plus SQE using the number of generated contexts.
+    const auto contextCount = static_cast<uint16_t>(contexts->size());
```
ygwpz, line 247: ⚠️ **Potential Overflow:** `static_cast<uint16_t>(contexts->size())` could truncate if `activeCopies.size()` exceeds UINT16_MAX (65535). Given `kMaxReadyContexts = 128`, this shouldn't happen in practice, but the code should either validate this explicitly or document the assumption.

ucm/transport/p2p/ffts/src/ffts_engine.cpp
```diff
@@ -0,0 +1,263 @@
...
+    if (status.Failure()) { return status; }
+
+    // Keep the context buffer alive until the stream is synchronized.
+    KeepAlive(std::move(contexts));
+    return Status::OK();
+}
+
+void FftsEngine::KeepAlive(std::shared_ptr<ContextBuffer> contexts)
+{
+    pendingContexts_.emplace_back(std::move(contexts));
```
ygwpz, line 269: 💡 **Memory Consideration:** `pendingContexts_` grows without bound if `Synchronize()` is not called frequently. Consider adding a size limit or a cleanup mechanism to prevent unbounded memory growth in long-running operations.

ucm/transport/p2p/ffts/src/ffts_engine.cpp
pyxyzc: Done. Added compile-time checks for SDMA/common context alignment and contextType ABI compatibility, including both offset and field type. Also documented that Ascend FFTS Plus treats rtFftsPlusComCtx_t as fixed-size storage for tagged concrete contexts before the reinterpret_cast.

ucm/transport/p2p/ffts/detail/ffts_engine.h
```diff
@@ -0,0 +1,71 @@
...
+    FftsEngine();
+    ~FftsEngine();
+
+    // Non-copyable and non-movable
+    FftsEngine(const FftsEngine&) = delete;
+    FftsEngine& operator=(const FftsEngine&) = delete;
+    FftsEngine(FftsEngine&&) = delete;
+    FftsEngine& operator=(FftsEngine&&) = delete;
+
+    Status Setup(int32_t deviceId);
```
Wwwzff, line 50: is it necessary to also have a teardown/shutdown for this class?

ucm/transport/p2p/ffts/src/ffts_engine.cpp
```diff
@@ -0,0 +1,274 @@
...
+    std::ostringstream oss;
+    oss << call << " failed: " << code;
+    return Status{static_cast<int32_t>(code), oss.str()};
+}
+
+uint32_t Low32(uint64_t value) { return static_cast<uint32_t>(value & 0xFFFFFFFFULL); }
+
+uint32_t High32(uint64_t value) { return static_cast<uint32_t>((value >> 32U) & 0xFFFFFFFFULL); }
+
+uint32_t BuildSdmaMoveHeader()
```
Wwwzff, line 60: this func can be a constexpr too

ucm/transport/p2p/ffts/detail/ffts_engine.h
pyxyzc: RAII cleanup is sufficient for the current FftsEngine. The destructor already releases pending context buffers and destroys the ACL stream. Global ACL runtime and device lifecycle are intentionally not managed by FftsEngine.

ucm/transport/p2p/ffts/src/ffts_engine.cpp
pyxyzc: Done.

ucm/transport/p2p/ffts/src/ffts_engine.cpp
pyxyzc: `deviceId` is provided by the upper Transport layer, where validation is expected. So FFTS does not duplicate the check. Mark as done.

ucm/transport/p2p/ffts/src/ffts_engine.cpp
pyxyzc: SubmitChunk() receives at most kMaxReadyContexts descriptors. After filtering no-op copies, activeCopies.size() and therefore contexts->size() can only be <= kMaxReadyContexts. And now we have kMaxReadyContexts equal to 128. Mark as done.

ucm/transport/p2p/ffts/bench/ffts_bench_options.cpp
```diff
@@ -0,0 +1,144 @@
...
+    const auto parsed = std::stoull(value, &consumed, 0);
+    if (consumed != value.size()) { throw std::invalid_argument("invalid count value: " + value); }
+    return static_cast<size_t>(parsed);
+}
+
+std::vector<size_t> ParseList(const std::string& value, bool parseBytes)
+{
+    std::vector<size_t> result;
+    size_t begin = 0;
+    while (begin <= value.size()) {
```
ygwpz, line 68: ⚠️ **Boundary Condition:** The loop condition `while (begin <= value.size())` could potentially cause issues. When `begin == value.size()`, the subsequent `value.substr(begin, ...)` would return an empty string, which throws an exception. While this works due to the break statement, the condition should ideally be `while (begin < value.size())` for clarity and to avoid the edge case entirely. The current logic relies on the exception to handle the boundary.

PR #991, opened 2026-06-02
[Feat] Implementation of BufferManager and KvProtocol.

yumingyue624:
## Purpose
Implement BufferManager for slot-based memory management with IndexPool, and unify SQE packing/CQE unpacking into a single KvProtocol abstraction with comprehensive request validation and full integration testing.

## Modifications
1. Add BufferManager with IndexPool-based slot allocation, supporting HOST, HOST_PINNED, and ASCEND_DEVICE memory types via AscendBuffer.
2. Unify Sqe/Cqe into KvProtocol base class with PackSqe/UnpackCqe, replacing separate SqeManager and CqeManager.
3. Add ValidateRequest to check Request fields before packing, including bit-field overflow, alignment, key length, and batch number validation with actual values in error messages and [[unlikely]] hints.
4. Add ProtocolManager to manage send_buffer and flag_buffer instances, with PackRequest, UnpackResponse, and PollResponseCid interfaces.
5. Use protocol-specific max batch numbers per SSU spec: BatchStore/ BatchRetrieve=110, Delete=254, Exist=256.

## Test
BufferManagerTest (19 cases), KvProtocolPackTest (38 cases), and ProtocolManagerTest (13 integration cases) all passed. Total: 70 tests.


ucm/transport/kv/asu/trans/src/buffer_manager.cpp
```diff
@@ -0,0 +1,205 @@
...
+    }
+    if (size == 0) {
+        return Status::Error(StatusCode::INVALID_ARGUMENT, name_ + ": size must be non-zero");
+    }
+    if (size > slot_size_) {
+        return Status::Error(StatusCode::INVALID_ARGUMENT, name_ + ": size exceeds slot_size");
+    }
+
+    int attempt = 0;
+    for (;;) {
```
ygwpz: 🔴 **Infinite loop risk:** The `Allocate` function contains an infinite loop without a timeout or maximum retry limit. If all slots are exhausted and never released, threads calling this function will hang indefinitely.

```cpp
for (;;) {
    auto idx = index_pool_.Acquire();
    if (idx != IndexPool::npos) {
        // ... return success
    }
    Backoff(attempt);  // Just backs off, never exits
}
```

**Fix options:**
1. Add a configurable timeout or maximum retry count
2. Return an error after exhausting retries
3. Add a blocking wait mechanism with proper signaling

```cpp
// Example fix with timeout:
constexpr int kMaxAttempts = 100;
if (attempt >= kMaxAttempts) {
    return Status::Error(StatusCode::RESOURCE_BUSY, name_ + " pool exhausted after retries");
}
```

ucm/transport/kv/asu/trans/src/kv_protocol.cpp
```diff
@@ -0,0 +1,790 @@
...
+    for (std::uint16_t i = 0; i < batch_number; ++i) {
+        std::uint32_t dword_idx = i / 32;
+        std::uint32_t bit = i % 32;
+        result_buffer[i] = (result_data[dword_idx] >> bit) & 0x1;
+    }
+}
+
+Status KvStoreProtocol::PackSqe(const SqeRequest& req, std::uint32_t* target)
+{
+    auto& r = static_cast<const KvStoreRequest&>(req);
```
ygwpz, line 60: ⚠️ **Unsafe type casting:** The `static_cast<const KvStoreRequest&>(req)` assumes the input is always the correct type without validation. If an incorrect request type is passed, this will result in undefined behavior.

```cpp
auto& r = static_cast<const KvStoreRequest&>(req);  // No type check!
```

**Consider adding a type check:**

1. Add a type identifier field to `SqeRequest` base class
2. Use `dynamic_cast` with null check (requires RTTI)
3. Add explicit documentation that passing wrong types is UB
4. Use template-based dispatch instead of polymorphism

```cpp
// Option 1: Add type check
if (req.GetType() != RequestType::Store) {
    return Status::Error(StatusCode::INVALID_ARGUMENT, "wrong request type for Store protocol");
}
```

yumingyue624: The upper-layer ProtocolManager dispatches requests by KvOpcode, ensuring each protocol's PackSqe always receives the correct derived type. A runtime type check would be redundant given this invariant. We'll revisit if the dispatch logic changes.

ucm/transport/kv/asu/CMakeLists.txt
nrj868: Make sure asu will be compiled only when runtime == ascend

yumingyue624: Done. Wrapped the entire CMakeLists.txt in if(RUNTIME_ENVIRONMENT STREQUAL "ascend") ... endif().

nrj868 (approved):

PR #992, opened 2026-06-03
[Feat]Wa dump blockwise & support for vllm-ascend:0.20.2rc1

wuhuxiao:
## Purpose

Add configurable block-wise WA KV cache dumping for the FAWA connector.

Previously, WA cache dumping only persisted the final WA tail block for each chunk prefill. This made WA external-cache reuse depend on chunk boundaries and could miss intermediate canonical block boundaries. This PR adds an option to dump WA cache block-wise so each canonical hash block can persist its corresponding WA tail state.

## Modifications

- Add `wa_dump_block_wise` config option, defaulting to `true`.
- Update FAWA dispatch metadata generation to support two WA dump modes:
  - block-wise WA dump: dump WA tail rows for every canonical block boundary.
  - chunk-wise WA dump: keep the old behavior and dump only the final WA tail per chunk.
- Keep WA load behavior loading only the final matched WA boundary.
- Update TP dump slicing so block-wise WA dump follows the same TP key partitioning as FA dump.
- Split configured `posix_capacity_gb` between FA and WA stores.
- Document `wa_dump_block_wise` in `examples/ucm_config_example.yaml`.

## Test

### DIFF
--- a/ucm/integration/vllm/hma_connector.py
+++ b/ucm/integration/vllm/hma_connector.py
@@ -573,9 +573,7 @@ class UCMFAWAConnector(UCMDirectConnector, SupportsHMA):
             else (None, None)
         )
 
-        for group_id, group_spec in enumerate(
-            self._kv_cache_config.kv_cache_groups
-        ):
+        for group_id, group_spec in enumerate(self._kv_cache_config.kv_cache_groups):
             group_caches: dict[str, torch.Tensor] = {}
             for layer_name in group_spec.layer_names:
                 if isinstance(kv_caches[layer_name], torch.Tensor):
--- a/ucm/integration/vllm/hma_connector.py
+++ b/ucm/integration/vllm/hma_connector.py
@@ -748,10 +748,16 @@ class UCMFAWAConnector(UCMDirectConnector, SupportsHMA):
                 return []
             if fetch_wa_block_wise:
                 # Block-wise WA stores one tail row for each canonical boundary.
-                boundary_block_indices = window_boundary_token_idx // group_meta.token_block_size
+                boundary_block_indices = (
+                    window_boundary_token_idx // group_meta.token_block_size
+                )
                 offsets = np.arange(group_meta.tail_blocks - 1, -1, -1, dtype=np.int64)
-                boundary_block_indices = boundary_block_indices[:, None] - offsets[None, :]
-                return np.array(group_block_ids)[boundary_block_indices.flatten()].tolist()
+                boundary_block_indices = (
+                    boundary_block_indices[:, None] - offsets[None, :]
+                )
+                return np.array(group_block_ids)[
+                    boundary_block_indices.flatten()
+                ].tolist()
             else:
                 # Chunk-wise WA stores only the tail for the final boundary.
                 boundary_block_idx = (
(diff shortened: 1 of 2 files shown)


ucm/integration/vllm/hma_connector.py
```diff
@@ -504,6 +505,8 @@ def _base_store_config(
         config["posix_gc_enable"] = (
             self._role != KVConnectorRole.WORKER and dp_rank == 0
         )
+        if config.get("posix_capacity_gb", None) is not None:
```
ygwpz, line 502: ⚠️ **Unrelated change:** The division of `posix_capacity_gb` by 2 seems unrelated to the WA dump block-wise feature. Is this intentional?

If this is fixing a capacity calculation bug, it should be:
1. Documented in the PR description
2. Separated into a dedicated commit with clear explanation

If accidental, this change should be reverted.

```python
if config.get("posix_capacity_gb", None) is not None:
    config["posix_capacity_gb"] = int(config["posix_capacity_gb"]) // 2  # Why divide by 2?
```

ucm/integration/vllm/hma_connector.py
```diff
@@ -742,13 +746,20 @@ def _slice_group_block_ids(
...
-            ) + 1
-            return group_block_ids[
-                boundary_block_idx - group_meta.tail_blocks : boundary_block_idx
-            ]
+            if fetch_wa_block_wise:
+                # Block-wise WA stores one tail row for each canonical boundary.
+                boundary_block_indices = window_boundary_token_idx // group_meta.token_block_size
+                offsets = np.arange(group_meta.tail_blocks - 1, -1, -1, dtype=np.int64)
+                boundary_block_indices = boundary_block_indices[:, None] - offsets[None, :]
+                return np.array(group_block_ids)[boundary_block_indices.flatten()].tolist()
```
ygwpz: ⚠️ **Potential index out of bounds:** When computing block indices for block-wise WA:

```python
boundary_block_indices = boundary_block_indices[:, None] - offsets[None, :]
return np.array(group_block_ids)[boundary_block_indices.flatten()].tolist()
```

If `boundary_block_indices` contains negative values (when `offsets > boundary_block_indices`), numpy indexing will wrap around or produce unexpected results.

**Consider adding validation:**
```python
indices = boundary_block_indices.flatten()
if np.any(indices < 0) or np.any(indices >= len(group_block_ids)):
    # Handle edge case: return empty or clipped result
    valid_mask = (indices >= 0) & (indices < len(group_block_ids))
    indices = indices[valid_mask]
return np.array(group_block_ids)[indices].tolist()
```

ucm/integration/vllm/hma_connector.py
wuhuxiao: fa/ wa store share posix_capacity

ucm/integration/vllm/hma_connector.py
```diff
@@ -504,6 +505,8 @@ def _base_store_config(
         config["posix_gc_enable"] = (
             self._role != KVConnectorRole.WORKER and dp_rank == 0
         )
+        if config.get("posix_capacity_gb", None) is not None:
+            config["posix_capacity_gb"] = int(config["posix_capacity_gb"]) // 2
```
ygwpz, line 503: 💡 **Suggestion:** This division of `posix_capacity_gb` by 2 needs documentation. Why is the capacity halved? Is it to split capacity between FA and WA stores? Consider adding a comment explaining the rationale to avoid confusion.

qyh111 (approved):

harrisonyhq (approved):

PR #993, opened 2026-06-03
add connection manager

Frankielx:


ucm/transport/kv/asu/trans/src/connection_manager.h
```diff
@@ -0,0 +1,101 @@
...
+        std::function<std::vector<ConnectionHandle>(const AsuEndpoint&, std::uint32_t)>;
+
+    explicit ConnectionManager(CreateConnectionFunc create_fn);
+    ~ConnectionManager();
+
+    Status AddGroup(const AsuEndpoint& endpoint, std::uint32_t qp_num);
+    Status Shutdown();
+
+    std::shared_ptr<ConnectionChannel>
+    SelectConnection();  // the current implementation does not support concurrent calls
```
ygwpz, line 61: ⚠️ **Concurrency limitation:** The comment states `SelectConnection()` does not support concurrent calls. This is a significant architectural constraint. Consider:

1. Making this limitation more prominent in the API documentation
2. Adding assertions or debug checks to detect concurrent access
3. If this is intentional for performance, document the design rationale (single worker thread pattern)

Current callers may not be aware of this restriction, leading to potential race conditions.

ucm/transport/kv/asu/trans/src/connection_manager.cpp
```diff
@@ -0,0 +1,276 @@
...
+        return;
+    }
+
+    UC_DEBUG("ConnectionManager::ReportFailure MarkForDrain OK, ch_id={} state=DRAINING",
+             channel->GetChannelId());
+    cacheDirty_.store(true, std::memory_order_release);
+
+    {
+        std::unique_lock<std::shared_mutex> lock(drainMu_);
+        for (const auto& existing : drainList_) {
```
ygwpz, line 137: 💡 **Performance optimization:** Linear search through `drainList_` to check for duplicates could be inefficient when many channels are draining. Consider:

1. Using a `std::unordered_set` or `std::set` to track draining channels by pointer/ID
2. Or maintaining a separate "in-drain-list" flag in `ConnectionChannel`

```cpp
// Current: O(n) lookup for each ReportFailure
for (const auto& existing : drainList_) {
    if (existing.get() == channel.get()) return;
}
```

This becomes O(n²) if many failures occur.

ucm/transport/kv/asu/trans/src/connection_manager.cpp
```diff
@@ -0,0 +1,276 @@
...
+        std::this_thread::sleep_for(std::chrono::milliseconds(kRecoverIntervalMs));
+        if (stopRecover_.load(std::memory_order_acquire)) { break; }
+
+        std::vector<std::shared_ptr<ConnectionChannel>> to_recover;
+        {
+            std::unique_lock<std::shared_mutex> lock(drainMu_);
+            to_recover.swap(drainList_);
+        }
+
+        for (auto& channel : to_recover) {
```
ygwpz, line 172: ⚠️ **Recovery loop infinite retry risk:** When `createFn_()` fails, the channel is pushed back to `drainList_`. If recovery keeps failing (e.g., endpoint permanently down), channels will accumulate indefinitely.

Consider:
1. Adding a retry count limit per channel
2. Adding a "failed permanently" state and removing from recovery
3. Logging when channels stay in drain list for extended periods

```cpp
if (new_handles.empty()) {
    // This keeps retrying forever
    drainList_.push_back(std::move(channel));
    continue;
}
```

ucm/transport/kv/asu/trans/src/completion_poller.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+        if (batch.empty()) {
+            std::this_thread::sleep_for(std::chrono::milliseconds(1));
+            continue;
+        }
+
+        for (auto& r : batch) {
+            // 直接读取 flagBuffer 检查任务是否完成
+            if (r.flagBuffer != nullptr) {
+                volatile uint32_t* flagPtr = static_cast<volatile uint32_t*>(r.flagBuffer);
```
ygwpz, line 92: 🔴 **Incorrect atomic usage:** Using `volatile uint32_t*` for `flagBuffer` is not sufficient for concurrent access in C++. The `volatile` qualifier only prevents compiler optimizations, not CPU reordering.

```cpp
volatile uint32_t* flagPtr = static_cast<volatile uint32_t*>(r.flagBuffer);
if (*flagPtr != 0) {  // This read is NOT atomic!
```

**Fix:** Use `std::atomic<uint32_t>*` or ensure proper atomic operations:
```cpp
std::atomic<uint32_t>* flagPtr = static_cast<std::atomic<uint32_t>*>(r.flagBuffer);
if (flagPtr->load(std::memory_order_acquire) != 0) {
```

Alternatively, if the hardware guarantees atomic writes to aligned 32-bit values, document this assumption clearly.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -68,10 +166,33 @@ Status AsuTransportImpl::Shutdown()
...
     for (const auto& ctx : taskManager_.GetAll()) {
         if (ctx != nullptr) { (void)taskManager_.Remove(ctx->taskId); }
     }
+    if (connManager_) {
+        connManager_->Shutdown();
+        connManager_.reset();
+    }
+
+    // 释放 flagBuffer 内存
+    if (flagBufferMemHandle_ && transProvider_ && !config_.endpoints.empty()) {
```
ygwpz, line 181: ⚠️ **Shutdown order issue:** The logic at lines 181-191 attempts to unregister memory using `connManager_->SelectConnection()`, but `connManager_` may already be reset at line 177.

```cpp
if (connManager_) {
    connManager_->Shutdown();
    connManager_.reset();  // connManager_ is now nullptr
}
// Below: connManager_ might be nullptr here
if (flagBufferMemHandle_ && transProvider_ && !config_.endpoints.empty()) {
    auto firstConn = connManager_ ? connManager_->SelectConnection() : nullptr;  // Always nullptr!
```

**Fix:** Either:
1. Unregister memory BEFORE `connManager_.reset()`
2. Or store a connection handle separately before shutdown

ucm/transport/kv/asu/trans/src/connection_internal.h
```diff
@@ -0,0 +1,96 @@
...
+
+    std::uint32_t GetChannelId() const { return channelId; }
+    ChannelState GetState() const { return state.load(std::memory_order_acquire); }
+    std::uint32_t GetInflightCount() const { return inflightCount.load(std::memory_order_acquire); }
+    ConnectionGroup* GetGroup() const { return group; }
+    TransProvider::ConnectionHandle GetLink() const { return handle_; }
+
+    std::uint32_t FetchAddErrorCount(std::uint32_t val);
+
+    // Test helper to set state directly
```
ygwpz, line 56: 💡 **Test helper methods visibility:** The `SetState()` and `SetInflightCount()` methods are public but marked as "Test helper". This allows production code to accidentally modify internal state.

**Options:**
1. Make them private with `friend class` declarations for test classes
2. Move to a separate `ConnectionChannelTestAccess` helper in test files
3. Use `#ifdef ASU_BUILD_TESTS` to conditionally expose them

```cpp
#ifdef ASU_BUILD_TESTS
public:
    void SetState(ChannelState s) { ... }
#else
private:
#endif
```

PR #994, opened 2026-06-03
[Feat] Restore Ascend PC e2e gate

dante159753:
## What

Restores the pull-request PC online-inference e2e gate on the Ascend A2 self-hosted runner.

The new job runs inside the self-hosted runner container, verifies that the host Docker daemon is reachable through the mounted Docker socket, builds the Ascend vLLM test image with `docker build`, then runs the e2e test inside that image with `docker run`.

## Why

The previous PR gate no longer runs the prefix-cache online inference e2e coverage. This restores a focused PC smoke path for Ascend while keeping the runner model compatible with Docker-based self-hosted runners.

## Details

- Adds `test-e2e-pc-a2` to `.github/workflows/pull-request.yml` with a 20-minute timeout.
- Uses the Ascend A2 vLLM Dockerfile and host Docker daemon.
- Mounts Ascend runtime/device files into the test container.
- Defaults to `Qwen3-0.6B`, `UcmPipelineStore`, layerwise mode, 2GB cache buffer, and eager execution so it fits a 16GB vNPU.
- Lets the runner provide `A2_E2E_HOST_DEVICE`; falls back to `/dev/vdavinci212`.
- Adds Ascend-aware test setup so GPU lock allocation is skipped for Ascend jobs.

## Verification

Local:

- `python -m py_compile test/common/online_inference_utils.py test/suites/E2E/test_online_inference.py test/conftest.py`
- YAML parse for `.github/workflows/pull-request.yml` and `.github/actionlint.yaml`
- `git diff --check`

wyz validation:

- https://github.com/dante159753/unified-cache-management/actions/runs/26860793883
- Pull Request Gate #99 passed, including `test-e2e-pc-a2` on the A2 runner.
- During the e2e run, the runner successfully used the mounted Docker socket, built `ucm-e2etest-pc-a2`, started the test container, loaded `Qwen3-0.6B` on Ascend, completed the selected PC e2e test, uploaded pytest results, and cleaned up.
- A separate 4-vNPU runner smoke run passed with four parallel GitHub Actions jobs: https://github.com/dante159753/unified-cache-management/actions/runs/26859619609

Notes:

- Local `actionlint`, `black`, `isort`, and `codespell` were not installed in this sandbox; the PR gate runs the full pre-commit checks.


.github/workflows/pull-request.yml
```diff
@@ -196,3 +196,167 @@ jobs:
...
+
+  test-e2e-pc-a2:
+    timeout-minutes: 20
+    runs-on: [self-hosted, npu, a2]
+    needs: lint-and-unit-tests
+    permissions:
+      checks: write
+      pull-requests: write
+    env:
+      http_proxy: http://127.0.0.1:7890
```
ygwpz: ⚠️ **Hardcoded proxy configuration:** The proxy settings are hardcoded to `127.0.0.1:7890`. This works for the specific runner but may cause issues if:
1. The proxy port changes
2. Different runners have different proxy configurations
3. A runner doesn't have a proxy running at all

Consider using repository secrets or environment variables: `http_proxy: ${{ secrets.HTTP_PROXY || 'http://127.0.0.1:7890' }}`

.github/workflows/pull-request.yml
```diff
@@ -196,3 +196,167 @@ jobs:
...
+      - name: Test E2E Online Inference in Docker
+        run: |
+          RESULT_DIR="${{ github.workspace }}/test_results"
+          RESULT_FILE="${{ steps.version.outputs.version }}-online-inference.xml"
+          CONTAINER_NAME="ucm-e2etest-pc-a2-${{ steps.version.outputs.version }}"
+          SUDO=""
+          if [ "$(id -u)" -ne 0 ] && command -v sudo &> /dev/null; then
+            SUDO="sudo"
+          fi
+          A2_E2E_HOST_DEVICE="${A2_E2E_HOST_DEVICE:-/dev/vdavinci212}"
```
ygwpz, line 279: ⚠️ **Hardcoded device path:** The default device path `/dev/vdavinci212` appears to be specific to a particular machine. This could cause confusion for other contributors or if the runner hardware changes.

Consider:
1. Using a repository secret or variable for the device path
2. Adding a comment explaining this is the default A2 device
3. Making the fallback more robust (e.g., checking if the device exists)

.github/workflows/pull-request.yml
```diff
@@ -196,3 +196,167 @@ jobs:
...
+            --device "${A2_E2E_HOST_DEVICE}:/dev/davinci0" \
+            --device /dev/davinci_manager \
+            --device /dev/devmm_svm \
+            --device /dev/hisi_hdc \
+            -v /usr/local/dcmi:/usr/local/dcmi \
+            -v /usr/local/bin/npu-smi:/usr/local/bin/npu-smi \
+            -v /usr/local/Ascend/driver/lib64/:/usr/local/Ascend/driver/lib64/ \
+            -v /usr/local/Ascend/driver/version.info:/usr/local/Ascend/driver/version.info \
+            -v /etc/ascend_install.info:/etc/ascend_install.info \
+            -v /home/huangmy/models:/home/models:ro \
```
ygwpz, line 296: 🔴 **Hardcoded user-specific path:** The model path `/home/huangmy/models` is specific to one user's home directory. This will fail on any runner where the user is not `huangmy` or the models are stored elsewhere.

Consider:
1. Using a repository secret/variable: `${{ secrets.A2_MODEL_PATH || '/home/models' }}`
2. Mounting from a well-known shared location like `/opt/models` or `/data/models`
3. Adding a step to download the model if not present

test/conftest.py
```diff
@@ -242,12 +242,19 @@ def setup_gpu_resource(request):
     gpu_count_marker = request.node.get_closest_marker("gpu_count")
 
     if not gpu_mem_marker and not gpu_count_marker:
-        # No GPU markers, skip GPU resource setup
+        yield
+        return
+
+    if os.getenv("UCM_E2E_ACCELERATOR", "").lower() == "ascend":
+        if not os.getenv("ASCEND_RT_VISIBLE_DEVICES"):
+            device_count = int(gpu_count_marker.args[0]) if gpu_count_marker else 1
```
ygwpz: 💡 **Logic concern for Ascend device allocation:** When `gpu_count_marker` is None (only `gpu_mem_marker` present), `device_count` defaults to 1. However, for Ascend accelerators, the memory-based allocation logic from the GPU path is not replicated here. Consider:

1. Should Ascend also respect `gpu_mem_marker` to determine device count?
2. If a test has `@pytest.mark.gpu_mem(6000)` but no `gpu_count`, it gets only device 0 - is this intentional?

```python
# Current: always defaults to 1 device when gpu_count_marker is None
device_count = int(gpu_count_marker.args[0]) if gpu_count_marker else 1
```

If Ascend has different memory allocation semantics than CUDA GPUs, a comment explaining the difference would help maintainers understand the intent.

.github/workflows/pull-request.yml
```diff
@@ -195,4 +195,127 @@ jobs:
...
```
ygwpz, line 296: ⚠️ **Warning:** Hardcoded path `/home/huangmy/models` is specific to one user's machine. Consider using an environment variable like `${{ env.MODEL_PATH }}` or a repository secret/variable to make this portable across different runner machines.

ygwpz (approved):

PR #995, opened 2026-06-05
[Feat] Integrate ucm metrics into vLLM connector metrics

dante159753:
## Summary

- Add Python-side UCM metrics dispatching so existing UCM metrics can feed both multiproc and vLLM connector consumers.
- Expose UCM connector stats through vLLM KV connector metrics with `vllm:`-prefixed Prometheus names and worker/engine labels.
- Add and refine Grafana dashboards for UCM vLLM metrics, including direct connector task breakdowns, layerwise batch breakdowns, engine/worker filtering, and aggregated/per-worker views.
- Keep the dashboard identity isolated from the existing origin dashboards so both sets can coexist before merge.

## Why

The previous multiproc metrics path only covered the local multiproc exporter cleanly and did not fit vLLM's KV connector metrics snapshot path. This change lets UCM metrics be consumed through vLLM-native metric collection while preserving the existing multiproc path and making multi-worker/multi-engine deployments easier to inspect.

## Impact

- Users can view UCM connector, layerwise, cache/store, and vLLM overview metrics through the vLLM-prefixed dashboards.
- Aggregated dashboard mode now aggregates across engines, while Per Worker mode keeps `engine` and `worker_rank` dimensions for disambiguation.
- Existing `ucm:` multiproc metric output remains available through the existing configuration path.

## Verification
test using jenkins pipeline, no performance impact, metrics works fine.


ucm/metrics_config.py
```diff
@@ -0,0 +1,239 @@
...
+    vllm_connector_buckets: tuple[float, ...] = ()
+    vllm_connector_value_scale: float = 1.0
+    vllm_connector_enabled: bool = True
+
+
+def load_metrics_config(config_path: str) -> dict[str, Any]:
+    if not config_path:
+        return {}
+    try:
+        import yaml
```
ygwpz, line 55: 💡 **Suggestion:** The `yaml` import is inside a try-except block in `load_metrics_config`, but the ImportError is logged and returns empty dict. This means the function silently fails if PyYAML is not installed. Consider raising an exception when PyYAML is required (if a YAML config path is provided) to make the failure more explicit, or document this dependency clearly.

ucm/metrics_dispatcher.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def _merge_histogram(self, histograms: dict[str, Any], metric_name: str, value):
+        bucket_counts, sum_delta = self._histogram_tuple(value)
+        current_counts, current_sum = histograms.get(
+            metric_name, ([0] * len(bucket_counts), 0.0)
+        )
+        if len(current_counts) < len(bucket_counts):
+            current_counts.extend([0] * (len(bucket_counts) - len(current_counts)))
+        for index, count in enumerate(bucket_counts):
+            current_counts[index] += int(count)
```
ygwpz, line 128: 💡 **Suggestion:** When `count` is very large, `int(count)` could be problematic if `count` is a float that represents a very large number. Consider validating that `count` is within reasonable bounds before accumulating, or using `int(round(count))` to avoid floating-point precision issues when converting large floats to integers.

ucm/metrics_dispatcher.py
```diff
@@ -0,0 +1,150 @@
...
+        if isinstance(value, dict):
+            return list(value.get("bucket_counts", [])), float(value.get("sum", 0.0))
+        if isinstance(value, (tuple, list)) and len(value) == 2:
+            bucket_counts, sum_delta = value
+            return list(bucket_counts), float(sum_delta)
+        return list(getattr(value, "bucketCounts", [])), float(
+            getattr(value, "sum", 0.0)
+        )
+
+    def _empty_stats(self):
```
ygwpz, line 141: 💡 **Suggestion:** The `_histogram_tuple` function returns empty list for bucket_counts if the value doesn't match expected formats. This could lead to silently dropped histogram data. Consider logging a warning when histogram data format is unrecognized:
```python
if not bucket_counts:
    logger.warning(f"Unrecognized histogram format: {type(value)}")
```

ucm/integration/vllm/metrics.py
```diff
@@ -0,0 +1,450 @@
...
+        for rank, rank_data in other_histograms.items():
+            accumulator = histograms_by_rank.setdefault(str(rank), {})
+            for metric_name, value in rank_data.items():
+                _merge_histogram_value(accumulator, metric_name, value)
+        return self
+
+    def reduce(self) -> dict[str, int | float]:
+        return {
+            "ucm_load_requests_total": self._histogram_sum("load_requests_num"),
+            "ucm_load_blocks_total": self._histogram_sum("load_blocks_num"),
```
ygwpz: ⚠️ **Warning:** The `reduce()` method returns a hardcoded dict of specific metric names (`ucm_load_requests_total`, etc.). This means if new metrics are added to the config, they won be included in the reduced output. Consider making this method configurable or iterating over all histogram metrics in `self.data` to provide a complete summary.

ucm/metrics_dispatcher.py
```diff
@@ -0,0 +1,150 @@
...
+    VLLM_CONNECTOR_CONSUMER,
+    MetricDefinition,
+    consumer_enabled,
+    get_metric_definitions,
+)
+from ucm.shared.metrics import ucmmetrics
+
+CONSUMERS = (MULTIPROC_CONSUMER, VLLM_CONNECTOR_CONSUMER)
+_DISPATCHER: "MetricsDispatcher | None" = None
+_DISPATCHER_LOCK = threading.Lock()
```
ygwpz, line 40: ⚠️ **Warning:** The singleton pattern in `get_metrics_dispatcher` does not provide a way to reset the dispatcher. If the config changes (e.g., during hot-reload or testing), the old dispatcher will continue to be used. Consider adding a `reset_metrics_dispatcher()` function or passing a `force_new` parameter for scenarios where config needs to be reloaded:
```python
def reset_metrics_dispatcher():
    global _DISPATCHER
    with _DISPATCHER_LOCK:
        _DISPATCHER = None
```

ucm/metrics_config.py
dante159753: it should not block vllm from starting, error log should be enough to capture this

ucm/metrics_dispatcher.py
dante159753: it won't change

ucm/integration/vllm/metrics.py
dante159753: it's old implementation and no longer used, removed

ucm/shared/infra/logger/cc/spdlog_logger.cc
```diff
@@ -126,38 +145,84 @@ bool Logger::FilterCallSite(const char* file, int line)
...
+    if (value.empty()) { return defaultValue; }
+    std::transform(value.begin(), value.end(), value.begin(), ::tolower);
+    return value != "false" && value != "0" && value != "off";
+}
+
 std::shared_ptr<spdlog::logger> Logger::Make()
 {
     if (this->logger_) { return this->logger_; }
     std::lock_guard<std::mutex> lg(this->mutex_);
     if (this->logger_) { return this->logger_; }
```
ygwpz, line 160: 💡 **L4 Performance Consideration:** The `InternSourceString` function uses a static mutex for protecting the string pool. In high-frequency logging scenarios (many log statements per millisecond), this mutex could become a contention bottleneck since every log call with a new source location needs to intern the string.

Consider alternatives:
1. Use a concurrent hash map (e.g., `folly::ConcurrentHashMap` or similar)
2. Pre-intern common source locations at initialization
3. Use lock-free string interning with atomic pointers

For typical UCM usage this may not be critical, but worth documenting if high-frequency logging is expected.

ucm/metrics_dispatcher.py
ygwpz, line 128: 💡 **Previous Concern Still Unaddressed:** The `int(count)` conversion on very large float values could have precision issues. For example, if `count` is a float like `1e15`, `int(count)` would work, but intermediate floating-point operations could introduce rounding errors.

Consider using `int(round(count))` or validating that `count` is within reasonable bounds (e.g., `abs(count) < 1e12`) before conversion. This is especially important for histogram bucket counts which should always be integers.

ucm/metrics_dispatcher.py
```diff
@@ -0,0 +1,150 @@
...
+            current_counts[index] += int(count)
+        histograms[metric_name] = (current_counts, current_sum + float(sum_delta))
+
+    def _histogram_tuple(self, value):
+        if isinstance(value, dict):
+            return list(value.get("bucket_counts", [])), float(value.get("sum", 0.0))
+        if isinstance(value, (tuple, list)) and len(value) == 2:
+            bucket_counts, sum_delta = value
+            return list(bucket_counts), float(sum_delta)
+        return list(getattr(value, "bucketCounts", [])), float(
```
ygwpz, line 137: 💡 **Previous Concern Still Unaddressed:** The `_histogram_tuple` function returns empty list `[]` when the histogram value format is unrecognized (line 137-139). This silently drops histogram data without any warning.

Consider adding a warning log:
```python
if not bucket_counts:
    logger.warning(f"Unrecognized histogram format: {type(value)}, data dropped")
```

This helps diagnose configuration issues or unexpected metric formats.

ucm/store/cache/cc/dump_queue.cc
```diff
@@ -100,13 +105,23 @@ Status DumpQueue::DumpOneTask(CopyStream& stream, TaskPtr task)
...
             UC_ERROR("Failed({}) to wait prerequisite event for dump task({}).", s, task->id);
             return s;
         }
+        // Timestamp the moment the prerequisite event is actually satisfied, so the
+        // compute-wait phase can be split from the pure D2H copy phase. The callback
+        // is enqueued behind WaitEvent and ahead of the copies, adding no serialization.
+        eventReadyTp = std::make_shared<std::atomic<double>>(0.0);
+        auto cbStatus = stream.AppendCallback([eventReadyTp](bool) {
+            eventReadyTp->store(NowTime::Now(), std::memory_order_relaxed);
+        });
```
ygwpz, line 120: ⚠️ **L3 Issue - Potential Race Condition:** The callback stores the timestamp using `memory_order_relaxed`, which may not provide sufficient synchronization guarantees. The callback could be executed on a different thread (e.g., driver/stream callback thread) while the main thread reads the value. Consider using `memory_order_release` in the store and `memory_order_acquire` in the load:

```cpp
eventReadyTp->store(NowTime::Now(), std::memory_order_release);
// ...
auto ready = eventReadyTp->load(std::memory_order_acquire);
```

While `relaxed` is technically correct for a single atomic value, the timing semantics (splitting compute-wait vs D2H) rely on precise ordering which relaxed may not guarantee.

ucm/store/cache/cc/load_queue.cc
```diff
@@ -158,13 +160,21 @@ void LoadQueue::TransferOneTask(CopyStream& stream, ShardTask&& task)
...
         }
+        const auto copiedShards = holder_.size() + 1;
         s = stream.Synchronize();
+        auto h2dSyncMs = (NowTime::Now() - tpH2dSubmitted) * 1e3;
+        UC::Metrics::UpdateStats(NAME_TO_METRIC_ID("cache_h2d_sync_ms"), h2dSyncMs);
+        if (copiedShards > 0 && h2dSyncMs > 0.0) {
+            auto copiedBytes = static_cast<double>(copiedShards) * static_cast<double>(shardBytes_);
+            UC::Metrics::UpdateStats(NAME_TO_METRIC_ID("cache_h2d_bandwidth_gbps"),
+                                     copiedBytes / (h2dSyncMs * 1e-3) / 1e9);
+        }
```
ygwpz, line 182: 💡 **L4 Observation:** The bandwidth calculation `copiedShards = holder_.size() + 1` assumes all tasks in `holder_` are from the same batch and have identical shard sizes. However, `holder_` accumulates tasks across multiple `TransferOneTask` calls before synchronization. If different tasks have different sizes, the total bytes calculation could be inaccurate.

Consider:
1. Tracking actual bytes per task
2. Adding a comment explaining that `shardBytes_` is assumed constant for all tasks in `holder_`

ygwpz:
## Follow-up Review Summary (2026-06-13)

This follow-up review focused on the most recent changes since the last review on 2026-06-05, particularly:
- Logger redesign (commit d5f7185)
- Cache H2D/D2H bandwidth metrics (commits ab03186, edc4680)

### Previous Concerns Status
- **ucm/metrics_config.py:55** (PyYAML silent failure) - Author addressed with error logging. Accepted.
- **ucm/integration/vllm/metrics.py** (hardcoded reduce()) - Method now returns empty dict. Accepted.
- **ucm/metrics_dispatcher.py:40** (singleton reset) - Author declined. Accepted given use case.
- **ucm/metrics_dispatcher.py:128** (int conversion) - **Still unaddressed**, posted follow-up comment.
- **ucm/metrics_dispatcher.py:137** (silent histogram drop) - **Still unaddressed**, posted follow-up comment.

### New Issues Found
- **L3**: Potential race condition in dump_queue.cc callback timing mechanism (memory_order_relaxed)
- **L4**: Bandwidth calculation assumption in load_queue.cc may be inaccurate for variable-sized batches
- **L4**: String interning mutex contention in spdlog_logger.cc could impact high-frequency logging

Inline comments have been posted for these issues.

### Overall Assessment
The logger redesign addresses the self-reference and memory issues mentioned in the commit. The new bandwidth metrics provide useful visibility into H2D/D2H performance. The main concerns are around memory ordering in the callback-based timing and the assumptions in bandwidth calculations.

ygwpz (dismissed):
New changes look good - comprehensive metrics integration with proper consumer architecture and updated Grafana dashboards. The metrics dispatcher pattern allows proper fanout to multiple consumers without clearing data.

ucm/store/cache/cc/dump_queue.cc
dante159753: Fixed in the current branch. The callback now stores the timestamp with `std::memory_order_release`, and the reader side loads it with `std::memory_order_acquire` before calculating `cache_dump_prereq_wait_ms`.

ucm/store/cache/cc/load_queue.cc
dante159753: Keeping this calculation as-is. `shardBytes_` is fixed for the `LoadQueue`, and `holder_` batches shards from the same queue before the synchronization point, so `copiedShards * shardBytes_` matches the byte count this metric is intended to report.

ucm/metrics_dispatcher.py
dante159753: Keeping this as-is. The dispatcher consumes bucket counts from the UCM C++ metrics snapshot, where bucket deltas are integer counts. The `int(count)` conversion normalizes pybind/Python numeric values into the accumulator; it is not rounding arbitrary measured values.

ucm/metrics_dispatcher.py
dante159753: Keeping this as-is for now. `_histogram_tuple` only handles the formats emitted by `ucmmetrics`; unknown formats are ignored to avoid repeated warning logs in the metrics drain path. If another producer is added later, a `warning_once` style diagnostic would make sense there.

ucm/shared/infra/logger/cc/spdlog_logger.cc
dante159753: This thread is outdated now. The logger redesign is no longer part of the current PR diff, so I am leaving this out of the metrics PR.

ucm/metrics_config.py
```diff
@@ -0,0 +1,237 @@
...
+    if isinstance(inline_config, dict):
+        return inline_config
+    return load_metrics_config(launch_config.get("metrics_config_path", ""))
+
+
+def consumer_enabled(
+    config: dict[str, Any] | None, consumer: str, default: bool = True
+) -> bool:
+    consumers = (config or {}).get("consumers")
+    if not isinstance(consumers, dict):
```
qyh111, line 89: [P1] [旧配置会默认同时开启两个 consumer](https://github.com/ModelEngine-Group/unified-cache-management/blob/3a1195e217dbced1c39ff3ce01d51686c56dff60/ucm/metrics_config.py#L85-L91)：旧版配置没有 consumers 字段，此处会让 multiproc 和 vllm_connector 均返回 True；两者又默认使用相同的 ucm: 前缀，可能产生重复 metric family，并破坏声称的向后兼容。缺少 consumers 时应仅默认启用原有 multiproc 路径。

ucm/store/cache/cc/load_queue.cc
```diff
@@ -159,17 +161,27 @@ void LoadQueue::TransferOneTask(CopyStream& stream, ShardTask&& task)
...
+        UC::Metrics::UpdateStats(NAME_TO_METRIC_ID("cache_h2d_submit_ms"),
                                  (tpH2dSubmitted - tpBackendReady) * 1e3);
         if (!task.waiter) {
             holder_.push_back(std::move(task));
             return;
         }
+        const auto copiedShards = holder_.size() + 1;
         s = stream.Synchronize();
+        auto h2dSyncMs = (NowTime::Now() - h2dBatchStartTp_) * 1e3;
+        UC::Metrics::UpdateStats(NAME_TO_METRIC_ID("cache_h2d_sync_ms"), h2dSyncMs);
```
qyh111, line 177: [P1] [H2D 指标混入 backend wait](https://github.com/ModelEngine-Group/unified-cache-management/blob/3a1195e217dbced1c39ff3ce01d51686c56dff60/ucm/store/cache/cc/load_queue.cc#L153-L181)：计时从首个 shard 开始，包含后续 shard 的 WaitBackendTaskReady()。因此 cache_h2d_sync_ms 和 bandwidth 并非“纯 H2D”，存储慢时会被误判为 DMA 慢。应在最后一次 submit 后单独计量 residual sync，或使用设备事件计算拷贝时间。

qyh111 (approved):

ucm/metrics_config.py
dante159753: 构建都会包含metrics config，应该不会存在新代码使用旧metrics config的场景

ucm/store/cache/cc/load_queue.cc
dante159753: 准确计算纯h2d需要引入aclrtEventElapsedTime api计算设备端拷贝时间，要创建很多event来算event之间的设备时间，要新加很多逻辑，感觉不值得，本身现在event就容易不够。这里有通过已有callback逻辑统计sync过程中的计算时间，但这个有回调延迟，跟设备端误差比较大(~1ms)，用来算bandwidth不靠谱，用来对计算时间有个直观的感觉还是够得

ygwpz (approved):

PR #997, opened 2026-06-05
[Feat] Asu: add transport sub-batch foundation

yuanzhg078:
## Purpose
Add the ASU transport sub-batch foundation for splitting large requests, tracking sub-batch state, mapping ASU response/result codes, and finalizing transport task completion.

## Modifications
Add TransportSubBatchContext and TransportSubBatchState to track per-sub-batch CID, operation type, connection, buffers, status, and entry results.
Add ASU batch limits by operation type
Add IoScheduler to split entry and key batches into SSU-compliant sub-batches.
Add ASU response status mapping from CQE status and result-buffer entries to transport Status and QueryResult.
Add task completion utilities to release send/flag buffers, release connection inflight counts, count terminal sub-batches, and finalize task state.
Update BufferManager SGE lkey assignment and related test expectation.
Wire ASU transport unit tests into the CMake test target.

## Test
Added AsuResponseStatusTest, IoSchedulerTest, TransportTaskCompletionTest, and updated BufferManagerTest coverage


ucm/transport/kv/asu/CMakeLists.txt
```diff
@@ -1,9 +1,28 @@
+if(DEFINED ASCEND_ROOT AND NOT DEFINED Ascend_ROOT)
+    set(Ascend_ROOT "${ASCEND_ROOT}" CACHE PATH "Path to Ascend root directory")
+endif()
+if(NOT DEFINED Ascend_ROOT)
+    set(Ascend_ROOT "/usr/local/Ascend/ascend-toolkit/latest" CACHE PATH "Path to Ascend root directory")
+endif()
+
+find_path(ASU_ASCEND_INCLUDE_DIR
```
pyxyzc, line 8: Require acl/acl.h before checking the runtime. Move the Ascend header discovery inside the Ascend runtime block, or make the option explicitly reject non-Ascend runtimes.

ucm/transport/kv/asu/trans/src/transport_task_completion.cpp
```diff
@@ -0,0 +1,147 @@
...
+}  // namespace
+
+void InitializeTerminalSubBatchCount(TransportTaskContext& ctx)
+{
+    // At submit completion time, terminal sub-batches are usually submit/send failures.
+    ctx.completedSubBatchCount = 0;
+    for (const auto& subBatchContext : ctx.subBatchContexts) {
+        if (!IsSubBatchTerminal(subBatchContext.state)) { continue; }
+
+        ++ctx.completedSubBatchCount;
```
Infinite666: If two threads call InitializeTerminalSubBatchCount at the same time, completedSubBatchCount will be added twice.

ucm/transport/kv/asu/trans/src/asu_response_status.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    if (subBatchContext.status.code != StatusCode::ASU_CQE_CHECK_RESULT_BUFFER ||
+        response.result_buffer.empty()) {
+        std::fill(entryStatus.begin(), entryStatus.end(), subBatchContext.status);
+        return;
+    }
+
+    for (std::size_t index = 0; index < entryStatus.size(); ++index) {
+        entryStatus[index] =
+            ResultBufferEntryToStatus(subBatchContext.opType, response.result_buffer[index]);
```
pyxyzc: Is it necessary to validate the buffer size here :)

ucm/transport/kv/asu/trans/src/asu_response_status.cpp
```diff
@@ -0,0 +1,193 @@
...
+        case StatusCode::ASU_CQE_OUT_OF_CREATE_SIZE:
+            return Status::Error(StatusCode::ASU_CQE_OUT_OF_CREATE_SIZE, "Out of Create Size");
+        case StatusCode::ASU_CQE_IO_TIMEOUT:
+            return Status::Error(StatusCode::ASU_CQE_IO_TIMEOUT, "IO TimeOut");
+        case StatusCode::ASU_CQE_KEY_ALREADY_EXISTED:
+            return Status::Error(StatusCode::ASU_CQE_KEY_ALREADY_EXISTED, "Key already Existed");
+        case StatusCode::ASU_CQE_RESOURCE_BUSY:
+            return Status::Error(StatusCode::ASU_CQE_RESOURCE_BUSY, "Resource Busy");
+        case StatusCode::ASU_CQE_CHECK_RESULT_BUFFER:
+            return Status::Error(StatusCode::ASU_CQE_CHECK_RESULT_BUFFER,
```
pyxyzc: Ensure valid mixed query result is treated as an error or not :)

ucm/transport/kv/asu/trans/src/transport_task_completion.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+Status ReleaseSubBatchResources(TransportSubBatchContext& subBatchContext,
+                                BufferManager& sendBufferManager, BufferManager& flagBufferManager)
+{
+    Status finalStatus = Status::OK();
+
+    if (subBatchContext.sendSge.slot_index != UINT32_MAX) {
+        auto status = sendBufferManager.Free(subBatchContext.sendSge.slot_index);
+        if (!status.ok()) {
+            if (finalStatus.ok()) { finalStatus = status; }
```
Infinite666: No need to add judgement for finalStatus, it's always ok. And add some error logs.

ucm/transport/kv/asu/trans/src/transport_task_completion.cpp
```diff
@@ -0,0 +1,147 @@
...
+        }
+        subBatchContext.sendSge = {};
+    }
+
+    if (subBatchContext.flagBuffer.slot_index != UINT32_MAX) {
+        auto status = flagBufferManager.Free(subBatchContext.flagBuffer.slot_index);
+        if (!status.ok()) {
+            if (finalStatus.ok()) { finalStatus = status; }
```
Infinite666: Just set finalStatus and add some error logs.

ucm/transport/kv/asu/trans/src/transport_task_completion.cpp
```diff
@@ -0,0 +1,147 @@
...
+        ReleaseSubBatchResources(subBatchContext, sendBufferManager, flagBufferManager);
+    const auto completionStatus = status.ok() ? releaseStatus : status;
+    subBatchContext.state = (!completionStatus.ok() && state == TransportSubBatchState::COMPLETED)
+                                ? TransportSubBatchState::FAILED
+                                : state;
+    subBatchContext.status = completionStatus;
+    ++ctx.completedSubBatchCount;
+}
+
+void TryFinalizeTaskFromSubBatches(TransportTaskContext& ctx)
```
Infinite666: This function is dealing with TransportTaskContext inner variables and logics, so it should be inside TransportSubBatchContext for code structure and readability.

ucm/transport/kv/asu/trans/src/transport_task_completion.cpp
```diff
@@ -0,0 +1,147 @@
...
+    }
+}
+
+Status ReleaseSubBatchResources(TransportSubBatchContext& subBatchContext,
+                                BufferManager& sendBufferManager, BufferManager& flagBufferManager)
+{
+    Status finalStatus = Status::OK();
+
+    if (subBatchContext.sendSge.slot_index != UINT32_MAX) {
+        auto status = sendBufferManager.Free(subBatchContext.sendSge.slot_index);
```
Infinite666: Maybe should consider add lock to avoid double free.

ucm/transport/kv/asu/trans/src/io_scheduler.cpp
```diff
@@ -0,0 +1,85 @@
...
+std::size_t GetSubBatchCount(std::size_t total, std::size_t maxIoNum)
+{
+    if (total == 0 || maxIoNum == 0) { return 0; }
+    return 1 + (total - 1) / maxIoNum;
+}
+
+}  // namespace
+
+std::vector<IoScheduler::ScheduledIoBatch> IoScheduler::SplitForAsu(
+    const BatchView<KVBuffer>& entries, std::size_t maxIoNum) const
```
Infinite666: Can we set maxIoNum when IoScheduler init like config instead of passing it every time?

ucm/transport/kv/asu/trans/src/io_scheduler.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+    std::vector<ScheduledIoBatch> result;
+    if (entries.empty() || maxIoNum == 0) { return result; }
+
+    const std::size_t subBatchCount = GetSubBatchCount(entries.size, maxIoNum);
```
Infinite666: Looks like maxIoSize is a more suitable name compared to maxIoNum?

ucm/transport/kv/asu/trans/src/asu_response_status.cpp
yuanzhg078: No need here. In the normal completion path, these two sizes are expected to be equal: batchNumber passed to UnpackResponse is derived from entryStatus.size(), and the unpacking logic resizes result_buffer to that batchNumber.

ucm/transport/kv/asu/trans/src/asu_response_status.cpp
yuanzhg078: 0x732 means the batch operation failed and the caller should continue checking the result buffer for per-entry status. We can optimize/refine the status code definition later.

ucm/transport/kv/asu/trans/src/transport_task_completion.cpp
Fengli5355: There would be only one call to InitializeTerminalSubBatchCount (in CompleteTask) for each Transport task

ucm/transport/kv/asu/trans/src/transport_task_manager.h
```diff
@@ -51,6 +55,38 @@ enum class TransportTaskState {
     CANCELED = 4,
 };
 
+enum class TransportSubBatchState {
+    PENDING = 0,
+    COMPLETED = 1,
+    FAILED = 2,
```
Infinite666, line 61: Consider if we still need FAILED since we have it in status.

ucm/transport/kv/asu/trans/src/transport_task_manager.h
Infinite666, line 127: Consider if we still need this mutext.

yuanzhg078: Still need

ucm/transport/kv/asu/trans/src/transport_task_manager.h
yuanzhg078: Delet Failed state.

ucm/transport/kv/asu/trans/src/io_scheduler.cpp
yuanzhg078: Change to maxIoSize , include
kAsu...MaxIoNum -> kAsu...MaxIoSize
GetAsuMaxIoNum(...) -> GetAsuMaxIoSize(...)
GetSqeBatchLimit(...) -> GetSqeMaxIoSize(...)
SplitForAsu(..., maxIoNum)

ucm/transport/kv/asu/trans/src/transport_task_completion.cpp
yuanzhg078: Done. Changed this function to return void and added error logs for buffer release failures.

ucm/transport/kv/asu/trans/src/transport_task_completion.cpp
yuanzhg078: Done. Changed this function to return void and added error logs for buffer release failures.

ucm/transport/kv/asu/trans/src/transport_task_completion.cpp
yuanzhg078: Should move into TransportTaskContext

ucm/transport/kv/asu/CMakeLists.txt
yuanzhg078: Done. Add platform check at the begin of cmake file and reject non-Ascend runtime

ucm/transport/kv/asu/trans/src/io_scheduler.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    const std::size_t subBatchCount = GetSubBatchCount(keys.size, maxIoNum);
+    result.reserve(subBatchCount);
+
+    for (std::size_t offset = 0; offset < keys.size; offset += maxIoNum) {
+        const std::size_t end = std::min(offset + maxIoNum, keys.size);
+
+        ScheduledKeyBatch batch;
+        batch.keys = BatchView<CacheKey>{keys.data + offset, end - offset};
+        result.push_back(batch);
```
Wwwzff: use emplace_back instead

yuanzhg078: Thanks, but push_back is preferred here. ScheduledKeyBatch has no constructor accepting BatchView , so emplace_back would still require default-construct + assign, which is less readable with no performance gain for this lightweight type.

ucm/transport/kv/asu/trans/src/transport_task_completion.cpp
```diff
@@ -0,0 +1,147 @@
...
+    if (subBatchContext.state != TransportSubBatchState::PENDING) { return; }
+
+    const auto releaseStatus =
+        ReleaseSubBatchResources(subBatchContext, sendBufferManager, flagBufferManager);
+    const auto completionStatus = status.ok() ? releaseStatus : status;
+    subBatchContext.state = (!completionStatus.ok() && state == TransportSubBatchState::COMPLETED)
+                                ? TransportSubBatchState::FAILED
+                                : state;
+    subBatchContext.status = completionStatus;
+    ++ctx.completedSubBatchCount;
```
ygwpz: 🔴 **Race Condition:** `++ctx.completedSubBatchCount` is not atomic but appears to be accessed from multiple threads (completion handling). The `TransportTaskContext::state` is atomic, suggesting concurrent access. Consider using `std::atomic<std::uint32_t>` for `completedSubBatchCount` or protecting with a mutex.

ucm/transport/kv/asu/trans/src/transport_task_completion.cpp
```diff
@@ -0,0 +1,147 @@
...
+        if (finalStatus.ok() && !status.ok()) { finalStatus = status; }
+    }
+    return finalStatus;
+}
+
+void CompleteSubBatch(TransportTaskContext& ctx, TransportSubBatchContext& subBatchContext,
+                      TransportSubBatchState state, const Status& status,
+                      BufferManager& sendBufferManager, BufferManager& flagBufferManager)
+{
+    if (subBatchContext.state != TransportSubBatchState::PENDING) { return; }
```
ygwpz: ⚠️ **Thread Safety:** The state check `if (subBatchContext.state != TransportSubBatchState::PENDING) { return; }` is not atomic. If `CompleteSubBatch` can be called concurrently, this check-then-act pattern could lead to double completion or missed completions. Consider using atomic operations or a mutex to protect the state transition.

ucm/transport/kv/asu/trans/src/asu_response_status.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    if (subBatchContext.status.ok()) {
+        std::fill(entryStatus.begin(), entryStatus.end(), isQuery ? keyExist : Status::OK());
+        return;
+    }
+
+    if (isQuery && !subBatchContext.useSeekControl &&
+        subBatchContext.status.code == StatusCode::ASU_CQE_CHECK_RESULT_BUFFER) {
+        const auto existingKeyCount =
+            std::min(entryStatus.size(), static_cast<std::size_t>(response.existing_key_number));
```
ygwpz: ⚠️ **Potential Bug:** `static_cast<std::size_t>(response.existing_key_number)` does not check if `existing_key_number` could be negative. If the field type is signed and contains a negative value due to error or corruption, this cast would produce a very large positive number, causing `existingKeyCount` to exceed `entryStatus.size()`. Consider adding a bounds check or ensuring the field type is unsigned.

ucm/transport/kv/asu/trans/src/transport_task_manager.h
```diff
@@ -68,6 +116,8 @@ struct TransportTaskContext {
     QueryOptions queryOptions;
     QueryResult queryResult;
     std::vector<Status> entryStatus;
+    std::vector<TransportSubBatchContext> subBatchContexts;
+    std::uint32_t completedSubBatchCount{0};
```
ygwpz, line 120: 🔴 **Concurrency Issue:** `completedSubBatchCount` is declared as `std::uint32_t` but is modified in `CompleteSubBatch` and read in `TryFinalizeTaskFromSubBatches`. Given that `state` is atomic (suggesting concurrent access), this field should also be `std::atomic<std::uint32_t>` to prevent data races.

ucm/transport/kv/asu/trans/src/io_scheduler.cpp
yuanzhg078: Done. maxIoNum is now stored as a member variable in IoScheduler (set during initialization) and is no longer passed as a function parameter.

Infinite666 (dismissed):

Wwwzff (dismissed):

Infinite666 (approved):

PR #1000, opened 2026-06-06
[Feat] Add RouterConfig for AsuClient

Fengli5355:
## Purpose
AsuClient needs to parse config of router to initialize Router component.

## Modifications 
1. Support AsuClient Router config.
2. Supplement errer log.
3. Rename HashTableConfig and HashTableType to RouterConfig and RouterType

### DIFF
--- a/ucm/transport/kv/asu/client/src/client_router_config.cpp
+++ b/ucm/transport/kv/asu/client/src/client_router_config.cpp
@@ -1,0 +1,144 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2026 Huawei Technologies Co., Ltd. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ * */
+#include "client_router_config.h"
+#include <algorithm>
+#include <cctype>
+#include <string>
+#include "status_utils.h"
+
+namespace UC::ASU {
+namespace {
+
+std::string NormalizeAttrValue(std::string value)
+{
+    std::replace(value.begin(), value.end(), '-', '_');
+    std::transform(value.begin(), value.end(), value.begin(),
+                   [](unsigned char ch) { return static_cast<char>(std::toupper(ch)); });
+    return value;
+}
+
+bool GetAttr(const std::unordered_map<std::string, std::string>& attrs, const std::string& key,
+             std::string& value)
+{
+    auto iter = attrs.find(key);
+    if (iter == attrs.end()) { return false; }
+    value = iter->second;
+    return true;
+}
+
+Status GetUint64Attr(const std::unordered_map<std::string, std::string>& attrs,
+                     const std::string& key, std::uint64_t& value)
+{
+    std::string text;
+    if (!GetAttr(attrs, key, text)) { return Status::OK(); }
+
+    try {
+        std::size_t parsed{0};
+        const auto parsedValue = std::stoull(text, &parsed, 0);
+        if (parsed != text.size()) {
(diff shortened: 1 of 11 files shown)


ucm/transport/kv/asu/client/src/client_router_config.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include "client_router_config.h"
+#include <algorithm>
+#include <cctype>
+#include <string>
+#include "status_utils.h"
+
+namespace UC::ASU {
+namespace {
+
+std::string NormalizeAttrValue(std::string value)
```
ygwpz, line 33: 💡 **Performance:** `NormalizeAttrValue` takes `std::string` by value, causing an unnecessary copy. Consider changing to `const std::string& value` and return the modified copy only when needed, or use `std::string_view` for the parameter.

ucm/transport/kv/asu/client/src/client_router_config.cpp
```diff
@@ -0,0 +1,144 @@
...
+    try {
+        std::size_t parsed{0};
+        const auto parsedValue = std::stoull(text, &parsed, 0);
+        if (parsed != text.size()) {
+            return ASU_LOG_ERROR_STATUS(StatusCode::INVALID_ARGUMENT,
+                                        "invalid router config " + key + "=" + text);
+        }
+        value = parsedValue;
+        return Status::OK();
+    } catch (const std::exception&) {
```
ygwpz, line 65: 💡 **Suggestion:** Catching `std::exception` broadly hides whether the error was `std::invalid_argument` (invalid format) or `std::out_of_range` (value too large). Consider catching specific exceptions or logging the exception type for better debugging.

ucm/transport/kv/asu/client/src/client_router_config.cpp
```diff
@@ -0,0 +1,144 @@
...
+        return UC::KV::RouterType::RING_HASH_FULL_SPREAD;
+    }
+    if (type == "MAGLEV" || type == "MAGLEV_FULL_SPREAD") {
+        return UC::KV::RouterType::MAGLEV_FULL_SPREAD;
+    }
+    if (type == "CONTIGUOUS_BLOCK_AFFINITY") {
+        return UC::KV::RouterType::CONTIGUOUS_BLOCK_AFFINITY;
+    }
+    if (type == "BATCH_TOPK_AFFINITY") { return UC::KV::RouterType::BATCH_TOPK_AFFINITY; }
+    return fallback;
```
ygwpz, line 103: ⚠️ **Silent Misconfiguration:** When `ParseRouterType` receives an unrecognized type string, it silently returns the fallback value. This could lead to unexpected routing behavior if a user provides an invalid configuration. Consider logging a warning or returning an error status.

ucm/transport/kv/asu/client/src/view_server.cpp
```diff
@@ -161,6 +161,10 @@ bool ViewServer::ShouldRefreshView(const TaskResult& result) const
 
 std::shared_ptr<ViewServer> CreateDefaultViewServer(const AsuClientConfig& config)
 {
+    auto viewConfigPath = config.attrs.find("view.config_path");
+    if (viewConfigPath != config.attrs.end() && !viewConfigPath->second.empty()) {
+        return std::make_shared<ConfigFileViewServer>(viewConfigPath->second);
```
ygwpz, line 166: ⚠️ **Validation Missing:** The code creates a `ConfigFileViewServer` without validating that the config file path exists or is readable. If the path is invalid, the error will only surface later during `GetGlobalView()`. Consider adding a validation check here: `std::ifstream test(viewConfigPath->second); if (!test.is_open()) return Status::Error(...)`

ucm/transport/kv/asu/common/status_utils.h
```diff
@@ -0,0 +1,36 @@
...
+ * SOFTWARE.
+ * */
+#pragma once
+
+#include <string>
+#include <utility>
+#include "asu_transport/types.h"
+#include "logger/logger.h"
+
+#define ASU_LOG_ERROR_STATUS(code, message)                                   \
```
ygwpz, line 31: 💡 **Suggestion:** The macro uses a lambda to ensure message evaluation happens before `UC_ERROR`. This works correctly, but consider using a simpler inline function or template function instead of a macro for better type safety and debugging experience.

ucm/transport/kv/asu/client/src/client_router_config.cpp
```diff
@@ -0,0 +1,144 @@
...
+    }
+    if (normalized == "0" || normalized == "FALSE" || normalized == "OFF" || normalized == "NO") {
+        value = false;
+        return Status::OK();
+    }
+    return ASU_LOG_ERROR_STATUS(StatusCode::INVALID_ARGUMENT,
+                                "invalid router config " + key + "=" + text);
+}
+
+UC::KV::RouterType ParseRouterType(const std::string& value, UC::KV::RouterType fallback)
```
yuanzhg078, line 90: You should change ParseRouterType(...) so that it validates the input strictly instead of silently returning the fallback value for unknown strings. If the router type is invalid, the function should return an error status with a clear message, so configuration mistakes fail fast and are easier to diagnose.

ucm/transport/kv/asu/client/src/client_router_config.cpp
Fengli5355: This would be contained in further optimization.

ucm/transport/kv/asu/client/src/client_router_config.cpp
Fengli5355: This would be contained in further optimization.

ucm/transport/kv/asu/client/src/client_router_config.cpp
Fengli5355: The fallback value point to a default RouterType, which will not be written by users.

ucm/transport/kv/asu/client/src/view_server.cpp
Fengli5355: This would be contained in further fixes.

ucm/transport/kv/asu/client/src/client_router_config.cpp
Fengli5355: The fallback value point to a default RouterType, which will not be written by users.

Infinite666 (approved):

PR #1001, opened 2026-06-06
[FEAT] add trans provider

Frankielx:
## Purpose
Adapt the TransProvider abstraction layer. Refactor the original data sending path based on the free function Send() to use the virtual interface TransProvider::Send(), decoupling the underlying transport implementation (AICPU hcomm) from the upper-layer connection/channel management.
## Modifications 
1. TransProvider Abstraction Layer: Added trans_provider.h, defining the TransProvider virtual base class with interfaces including CreateConnection/DeleteConnections/Send/RegisterMemory/AllocThread/FreeThread/GetMemTokenId, as well as the SendIoBatch struct (replacing the same-named type previously in ConnectionManager).
2. AICPUTransProvider Implementation: Added aicpu_trans_provider.h/.cpp, implementing all TransProvider interfaces based on hcomm, including endpoint management, channel creation/destruction, AICPU kernel loading and sending, memory registration/deregistration, thread allocation/deallocation, and memory export/import (demo-specific).
3. ConnectionManager Refactoring:
- Constructor now accepts CreateConnectionFunc + DeleteConnectionFunc lambdas (no longer internally depends on TransProvider).
- SelectConnection() returns shared_ptr<ConnectionChannel> (replacing the original raw pointer).
- ConnectionChannel added GetLink() to return TransProvider::ConnectionHandle.
- Retained the channelCacheMu_ + shared_mutex structureMu_ dual-lock mechanism, introduced cacheDirty_ flag for deferred rebuild.
- RecoverLoop adopts incoming branch logic (no buggy old_qp reference).
4. AsuTransportImpl Adaptation:
- Init creates AICPUTransProvider + ConnectionManager (unique_ptr), bridging them via lambdas.
- BuildSubBatchSendBuffers constructs TransProvider::SendIoBatch{connectionHandle, sendBuffer, flagBuffer, len}.
- SendSubBatchBuffers calls transProvider_->Send() replacing the original free function Send().
5. Legacy Code Removal: Removed the old SendIoBatch struct and Send() free function from ConnectionManager, removed legacy connection concurrency/transport tests (connection_concurrent_test.cc, connection_transport_test.cc, test_helper.h), and removed the -include fmt_compat.h compiler option.
6. Test Adaptation: asu_submit_flow_test.cpp added StubTransProvider mock + ACL fixture (AsuSubmitFlowBufferTest); transport_task_completion_test.cpp adapted to unique_ptr<ConnectionManager> API; asu_smoke_test.cc uses StubTransport in place of the real AsuTransportImpl.                          
## Test
- 180/180 unit tests all passed (including 3 new Send-related tests, 2 BufferManager ACL fixture tests, and 16 ConnectionManager new API tests).
-  Provider demo dual-card (NPU 0↔2) E2E verification passed: rank0 sends "Hello World!" → rank1 successfully receives and validates.


ucm/transport/kv/asu/trans/src/aicpu_trans_provider.cpp
```diff
@@ -0,0 +1,635 @@
...
+            continue;
+        }
+
+        aclRet = aclrtKernelArgsAppend(argsHandle, &args, sizeof(args), &paramHandle);
+        if (aclRet != ACL_SUCCESS) {
+            UC_ERROR("AICPUTransProvider::Send: KernelArgsAppend failed, ret={}", aclRet);
+            aclrtKernelArgsFinalize(argsHandle);
+            results.push_back(Status::Error(StatusCode::INTERNAL_ERROR, "KernelArgsAppend failed"));
+            continue;
+        }
```
ygwpz, line 303: ⚠️ **Logic Issue:** `aclrtKernelArgsFinalize` is called before checking if `aclrtKernelArgsAppend` succeeded. If `aclrtKernelArgsAppend` fails, the argsHandle should be finalized before returning, but the current code calls finalize first, then checks its result - this means a failed Append would still proceed to Finalize. Consider restructuring: call Finalize only after Append succeeds.

ucm/transport/kv/asu/trans/src/aicpu_trans_provider.cpp
```diff
@@ -0,0 +1,635 @@
...
+                mapFile);
+        ipToDeviceMapLoaded_ = true;
+        return;
+    }
+
+    char line[256];
+    while (fgets(line, sizeof(line), fp)) {
+        char ip[64] = {};
+        uint32_t devId = 0;
+        if (sscanf(line, "%63s %u", ip, &devId) == 2) { ipToDeviceMap_[ip] = devId; }
```
ygwpz, line 71: 💡 **Suggestion:** The sscanf format specifier `%63s` correctly limits the input to 63 characters to prevent buffer overflow. However, consider using `std::string` and `std::istringstream` for safer parsing instead of C-style sscanf with fixed buffers.

ucm/transport/kv/asu/trans/src/aicpu_trans_provider.cpp
```diff
@@ -0,0 +1,635 @@
...
+        HcommEndpointDestroy(endpoint_);
+        endpoint_ = nullptr;
+        localIp_.clear();
+    }
+}
+
+Status AICPUTransProvider::CreateConnection(const std::string& localIp, const std::string& remoteIp,
+                                            uint32_t port, uint32_t qpNum, uint32_t timeout,
+                                            std::vector<ConnectionHandle>& connectionHandles)
+{
```
ygwpz, line 146: ⚠️ **Edge Case:** If `qpNum` is 0, the function returns `Status::OK()` with an empty `connectionHandles` vector. Should this be considered an error case? Consider adding a validation check at the beginning: `if (qpNum == 0) return Status::Error(...)`

ucm/transport/kv/asu/trans/src/aicpu_trans_provider.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+    void* memDesc = nullptr;
+    uint32_t memDescLen = 0;
+    HcommResult ret = HcommMemExport(endpoint_, memHandle, &memDesc, &memDescLen);
+    if (ret != 0) {
+        UC_ERROR("AICPUTransProvider::GetMemTokenId: HcommMemExport failed, ret={}", ret);
+        return Status::Error(StatusCode::INTERNAL_ERROR, "HcommMemExport failed");
+    }
+
+    // 解析序列化的 ExchangeUbBufferDto 提取 tokenId
```
ygwpz, line 473: ⚠️ **Maintainability:** The parsing logic assumes a specific serialization format (`addr(8) + size(8) + memType(4) + memTag(4+len) + tokenValue(4) + tokenId(4)`). If the upstream serialization format changes, this code will silently produce incorrect results or crash. Consider: (1) adding a format version check, (2) documenting the expected format in a comment, or (3) using a more structured parsing approach.

ucm/transport/kv/asu/trans/src/aicpu_trans_provider.h
```diff
@@ -0,0 +1,106 @@
...
+                        uint32_t importLen, MemHandle* importedHandle);
+
+    Status GetImportedMemoryInfo(ConnectionHandle connectionHandle, MemHandle importedHandle,
+                                 uint64_t* addr, uint64_t* size);
+
+    Status UnimportMemory(ConnectionHandle connectionHandle, MemHandle importedHandle);
+
+    Status InitEndpoint(const std::string& localIp);
+
+    const EndpointDesc& GetLocalEndpointDesc() const { return localEndpointDesc_; }
```
ygwpz, line 59: ⚠️ **Thread Safety:** `GetLocalEndpointDesc()` returns a const reference to `localEndpointDesc_` without acquiring the mutex. If `GetOrCreateEndpoint()` is modifying this field concurrently, this could return a partially updated or inconsistent descriptor. Consider either: (1) returning a copy instead of a reference, or (2) adding a mutex lock.

ucm/transport/kv/asu/trans/src/connection_manager.cpp
```diff
@@ -28,57 +28,26 @@
...
-void ConnectionManager::PrepareForInit()
-{
-    shuttingDown_.store(false, std::memory_order_release);
-    stopRecover_.store(false, std::memory_order_release);
-}
-
 Status ConnectionManager::AddGroup(const AsuEndpoint& endpoint, std::uint32_t qp_num)
 {
     if (shuttingDown_.load(std::memory_order_acquire)) {
         return Status::Error(StatusCode::NOT_INITIALIZED, "connection manager shutting down");
```
yuanzhg078, line 42: CreateConnection() should reject qpNum == 0 with INVALID_ARGUMENT instead of returning OK with no connections.

ucm/transport/kv/asu/trans/src/aicpu_trans_provider.cpp
Frankielx: The ACL API requires that argsHandle must release resources through Finalize, and Finalize cleaning is still necessary even if Append fails

ucm/transport/kv/asu/trans/src/aicpu_trans_provider.cpp
Frankielx: The IP mapping file is in a simple format generated by hccn_tool (" IP device_id "), and sscanf is fully sufficient and more concise. In this scenario, istringstream has no substantial security advantages; instead, it is even more verbose. %63s has correctly restricted the buffer length.

ucm/transport/kv/asu/trans/src/aicpu_trans_provider.h
Frankielx: This function has been deleted

ucm/transport/kv/asu/trans/include/acl/acl_base.h
```diff
@@ -0,0 +1,19 @@
+/**
+ * Copyright (c) 2025 Huawei Technologies Co., Ltd.
+ * This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+ * CANN Open Software License Agreement Version 2.0 (the "License").
+ * Please refer to the License for details. You may not use this file except in compliance with the License.
```
yuanzhg078, line 5: No need to add the ACL header here.

ucm/transport/kv/asu/trans/src/connection_manager.cpp
Frankielx: This should be modified

ucm/transport/kv/asu/trans/src/aicpu_trans_provider.cpp
Frankielx: This should be modified

PR #1002, opened 2026-06-08
[Feat] Add VerifyPackedBuffer for all KV protocols

yumingyue624:
## Purpose
Add pre-send verification for packed SQE buffers to detect memory
corruption or malformed packets before submission.

## Modifications
1. Add VerifyPackedBuffer(const uint32_t* data, size_t length) pure
virtual method to KvProtocol base class.
2. Implement VerifyPackedBuffer for all 7 protocols (Store, Retrieve,
BatchStore, BatchRetrieve, Delete, Exist, KeepAlive), verifying:
   - Total length matches expected size (fixed or batch_number-derived)
   - Opcode, kFixedBits, DptrType fixed values
   - Valid fields: buffer_addr non-zero, buffer_length/offset 512B
     aligned, key non-all-zero, batch_number in range, rflag/response
     consistency
   - All reserved bits are zero
3. Add ProtocolManager::VerifyPackedBuffer(const void*, size_t) that
extracts opcode from data[0] and dispatches to the correct protocol.

## Test
 All 123 new VerifyPackedBuffer tests added.

### DIFF
--- a/ucm/transport/kv/asu/trans/src/kv_protocol.cpp
+++ b/ucm/transport/kv/asu/trans/src/kv_protocol.cpp
@@ -55,6 +55,95 @@ static void UnpackResultBuffer1Bit(const std::uint32_t* result_data, std::uint16
     }
 }
 
+static Status VerifyFixedBits(const std::uint32_t* data, const char* log_prefix)
+{
+    if (((data[0] >> 14) & 0x3) != kFixedBits) {
+        return Status::Error(StatusCode::INVALID_ARGUMENT,
+                             std::string(log_prefix) + ": fixed bits mismatch");
+    }
+    return Status::OK();
+}
+
+static Status VerifyRflagConsistency(const std::uint32_t* data, const char* log_prefix)
+{
+    bool rflag = (data[0] >> 13) & 0x1;
+    std::uint64_t resp_addr =
+        static_cast<std::uint64_t>(data[3]) | (static_cast<std::uint64_t>(data[4]) << 32);
+    std::uint32_t resp_mr_key = data[5];
+    if (rflag) {
+        if (resp_addr == 0) {
+            return Status::Error(
+                StatusCode::INVALID_ARGUMENT,
+                std::string(log_prefix) + ": rflag set but response_buffer_addr is zero");
+        }
+        if (resp_mr_key == 0) {
+            return Status::Error(
+                StatusCode::INVALID_ARGUMENT,
+                std::string(log_prefix) + ": rflag set but response_mr_key is zero");
+        }
+    } else {
+        if (resp_addr != 0) {
+            return Status::Error(
+                StatusCode::INVALID_ARGUMENT,
+                std::string(log_prefix) + ": rflag false but response_buffer_addr non-zero");
+        }
+        if (resp_mr_key != 0) {
+            return Status::Error(
+                StatusCode::INVALID_ARGUMENT,
+                std::string(log_prefix) + ": rflag false but response_mr_key non-zero");
+        }
+    }
+    return Status::OK();
+}
+
+static Status VerifyKeyNotAllZero(const std::uint32_t* key_dwords, const std::string& log_prefix)
+{
+    if (key_dwords[0] == 0 && key_dwords[1] == 0 && key_dwords[2] == 0 && key_dwords[3] == 0) {
+        return Status::Error(StatusCode::INVALID_ARGUMENT, log_prefix + ": key is all zeros");
+    }
+    return Status::OK();
+}
+
+static Status VerifyBatchEntry(const std::uint32_t* base, const std::string& log_prefix,
+                               std::uint16_t i)
+{
+    if (base[0] % kAlignmentBytes != 0) {
+        return Status::Error(
+            StatusCode::INVALID_ARGUMENT,
+            log_prefix + ": entry[" + std::to_string(i) + "] offset not 512B aligned");
+    }
+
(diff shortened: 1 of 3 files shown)


ucm/transport/kv/asu/trans/src/kv_protocol.cpp
```diff
@@ -339,6 +488,136 @@ Status KvBatchStoreProtocol::UnpackCqe(const std::uint32_t* data, std::uint16_t
     return Status::OK();
 }
 
+Status KvBatchStoreProtocol::VerifyPackedBuffer(const std::uint32_t* data, std::size_t length) const
```
yuanzhg078, line 560: There is a lot of duplicated validation logic here; please consider moving the common checks (for example reserved bits, rflag consistency, alignment, and key validation) into shared helpers so each VerifyPackedBuffer() only keeps the protocol-specific rules.

ucm/transport/kv/asu/trans/src/kv_protocol.cpp
```diff
@@ -561,6 +972,105 @@ Status KvDeleteProtocol::UnpackCqe(const std::uint32_t* data, std::uint16_t batc
     return Status::OK();
 }
 
+Status KvDeleteProtocol::VerifyPackedBuffer(const std::uint32_t* data, std::size_t length) const
```
yuanzhg078, line 915: `KvDeleteProtocol::VerifyPackedBuffer `and `KvExistProtocol::VerifyPackedBuffer` have a very high degree of duplication. Their validation flow is almost identical, and the only real protocol-specific differences seem to be the batch_number upper bound and the data[10] bit layout/reserved-bit rule. Consider extracting the shared validation logic into a common helper.

nrj868 (dismissed):
nice

ucm/transport/kv/asu/trans/src/kv_protocol.cpp
yumingyue624: Extracted the common verification logic into shared helper functions.

ucm/transport/kv/asu/trans/src/kv_protocol.cpp
yumingyue624: Extracted the common verification logic into shared helper functions. The remaining differences between protocols such as Store/Retrieve are each protocol-specific, and under the current KvProtocol hierarchy where each protocol independently inherits from the base class, there is no suitable place to merge them.

ucm/transport/kv/asu/trans/src/kv_protocol.cpp
```diff
@@ -55,6 +55,95 @@ static void UnpackResultBuffer1Bit(const std::uint32_t* result_data, std::uint16
...
+    if (key_dwords[0] == 0 && key_dwords[1] == 0 && key_dwords[2] == 0 && key_dwords[3] == 0) {
+        return Status::Error(StatusCode::INVALID_ARGUMENT, log_prefix + ": key is all zeros");
+    }
+    return Status::OK();
+}
+
+static Status VerifyBatchEntry(const std::uint32_t* base, const std::string& log_prefix,
+                               std::uint16_t i)
+{
+    if (base[0] % kAlignmentBytes != 0) {
```
ygwpz, line 110: 💡 **Question:** The `VerifyBatchEntry` function validates that `base[0]` (offset) is 512B aligned. However, based on the diff context, `base` points to the start of a batch entry within the packed buffer, not the absolute offset in the SQE. Should this check validate that the entry's position in the buffer is aligned, or is `base[0]` actually a field representing an offset? Please clarify the semantic meaning of this check.

ucm/transport/kv/asu/trans/src/kv_protocol.cpp
```diff
@@ -739,6 +1255,25 @@ Status ProtocolManager::PollResponseCid(const void* data_ptr, std::uint16_t& cid
     return Status::OK();
 }
 
+Status ProtocolManager::VerifyPackedBuffer(const void* data_ptr, std::size_t length)
+{
+    if (!data_ptr || length < kSqeDwordCount * sizeof(std::uint32_t)) {
```
ygwpz, line 1260: 💡 **Suggestion:** The check `length < kSqeDwordCount * sizeof(std::uint32_t)` ensures minimum buffer size for reading the opcode. However, for batch operations (BatchStore, BatchRetrieve), the actual required length depends on `batch_number`. Consider adding a note that the caller should use the specific protocol's `VerifyPackedBuffer` for complete validation, or that this function validates only the header structure.

ucm/transport/kv/asu/trans/src/kv_protocol.cpp
yumingyue624: base[0] is a semantic field within the batch entry representing the KV offset value (not the entry's position in the packed buffer). The entry's position is computed as data + kSqeDwordCount + i * kBatchEntryDwordCount. This check validates that the offset field value is 512B aligned, consistent with the protocol specification.

ucm/transport/kv/asu/trans/src/kv_protocol.cpp
yumingyue624: The minimum length check here is intentionally for the header only, to safely read the opcode for dispatch. Full length validation (including batch_number-dependent size) is performed by each protocol's VerifyPackedBuffer. Will add a clarifying comment.

nrj868 (approved):

PR #1006, opened 2026-06-08
[Bugfix] Environment pre-check code adaptation

Menglths:
## Purpose
Implement the **Environment PreCheck Suite** to perform comprehensive health checks for Ascend NPU clusters before model training/deployment, ensuring device availability, network connectivity, and storage performance stability.

## Modifications
1. Added `run_hccn_device_status_check` function: Implements HCCN device status detection (link neighbor, physical link, network health, IP configuration) and adds overall status aggregation logic.
2. Added `remote_local_hccn_cards_ping_test` function: Supports local/remote HCCN card ping tests for NPU network connectivity verification with unified result return.
3. Added complete storage bandwidth check module: Implements multi-process concurrent dump/load performance testing, calculates average bandwidth, and outputs standardized performance reports.

## Test
All tests have passed.


test/common/envPreCheck/run_env_preCheck.py
```diff
@@ -690,33 +697,45 @@ def remote_local_hccn_cards_ping_test(
     :param dst_ip: destination IP
     :return: dict -> { "card_X_to_card_Y": {"status": True/False, "output": "..."}, "global_status": {...} }
     """
+    HCCN_TOOL = "/usr/local/Ascend/driver/tools/hccn_tool"
+
     if src_type == "local":
-        print(f"[TEST] LOCAL card {src_card} → REMOTE {dst_ip}")
+        print(f"[TEST] LOCAL card {src_card} ὆~R REMOTE {dst_ip}")
```
ygwpz: 🔴 **Encoding Issue:** This line contains a corrupted character sequence ("὆~R") that should be an arrow symbol ("→"). This appears to be an encoding corruption during editing. Please replace with the correct UTF-8 arrow character.

Menglths: The issues have been solved.

test/common/envPreCheck/run_env_preCheck.py
```diff
@@ -1119,282 +1138,127 @@ def run_check_model_weight():
...
-        ucmstore.CommitBatch(hashes, True)
+worker_number = 1
+shard_size = 1 * 1024 * 1024
+shard_number = 1
+block_number = 64
+dump_epoch_number = 32
+load_epoch_number = 32
+storage_backends = STORAGE_BACKENDS
+
+dump_bw_list = multiprocessing.Manager().list()
```
ygwpz: 🟠 **L3 Issue: Global multiprocessing.Manager() at module level**

These `multiprocessing.Manager().list()` objects are created at module import time, which has several issues:

1. **Server process spawn**: `multiprocessing.Manager()` spawns a server process when the module is imported, even if `run_bandwidth_check()` is never called.

2. **State accumulation**: If `run_bandwidth_check()` is called multiple times, the lists accumulate values from previous runs, leading to incorrect bandwidth calculations.

3. **Cleanup concerns**: The Manager server may not be properly cleaned up.

**Recommended fix**: Move these inside the `run_bandwidth_check()` function and reset at the start:
```python
def run_bandwidth_check():
    manager = multiprocessing.Manager()
    dump_bw_list = manager.list()
    load_bw_list = manager.list()
    # ... rest of the function
```
Or use `multiprocessing.Array` or pass the lists via function arguments to worker processes.

test/common/envPreCheck/run_env_preCheck.py
```diff
@@ -1119,282 +1138,127 @@ def run_check_model_weight():
...
-        ret = ucmstore.Wait(task_id)
-        if ret != 0:
-            raise RuntimeError(
-                f"ucmstore.Wait failed for embed task_id={task_id}, ret={ret}"
-            )
-
-        # Release host buffers and commit
-        ucmstore.ReleaseHostBuffers(buffers)
-        ucmstore.CommitBatch(hashes, True)
+worker_number = 1
```
ygwpz: 💡 **Suggestion: Hardcoded parameters should be configurable**

These bandwidth test parameters are hardcoded as global constants:
- `worker_number = 1`
- `shard_size = 1 * 1024 * 1024`
- `block_number = 64`
- `dump_epoch_number = 32`
- `load_epoch_number = 32`

Consider making them configurable via `config.yaml` similar to other parameters like `expected_embed_bandwidth` and `expected_fetch_bandwidth`. This would allow tuning the benchmark for different environments.

The `worker_number` should ideally match the number of available devices (from `ASCEND_RT_VISIBLE_DEVICES`) for realistic multi-device benchmarks.

test/common/envPreCheck/run_env_preCheck.py
```diff
@@ -1119,282 +1138,127 @@ def run_check_model_weight():
...
+    config["shard_size"] = shard_size
+    config["block_size"] = shard_size * shard_number
+    config["device_id"] = device_id
+    return UcmConnectorFactoryV1.create_connector(class_name, config, module_path)
+
+
+def make_array(size, alignment=262144, dtype=np.uint8) -> np.ndarray:
+    itemsize = np.dtype(dtype).itemsize
+    total_bytes = size * itemsize
+    mm = mmap.mmap(-1, total_bytes + alignment)
```
ygwpz, line 1164: ⚠️ **L4 Issue: Resource leak - mmap not closed**

The mmap object created here is never explicitly closed:
```python
mm = mmap.mmap(-1, total_bytes + alignment)
```

While Python's garbage collector will eventually clean it up, explicit cleanup is recommended for proper resource management. Consider:
1. Adding a cleanup mechanism (e.g., close the mmap when the test completes)
2. Or use a context manager wrapper
3. Or document that the memory is intentionally kept for the duration of the test

If the arrays need to persist throughout the test, at least add documentation explaining why the mmap isn't closed.

test/common/envPreCheck/run_env_preCheck.py
```diff
@@ -690,33 +697,45 @@ def remote_local_hccn_cards_ping_test(
...
-            print(f"[RESULT] REMOTE {src_ip} card {src_card} → LOCAL {dst_ip}: Failed")
+
+        if completed.returncode == 0:
+            print(f"[RESULT] {pair_key}: SUCCESS")
+        else:
+            print(f"[RESULT] {pair_key}: FAILED")
 
         result_dict[pair_key]["status"] = completed.returncode == 0
 
     except Exception as e:
```
ygwpz, line 742: ⚠️ **L4 Issue: Exception variable swallowed**

The exception `e` is caught but never used or logged:
```python
except Exception as e:
    result_dict[pair_key]["status"] = False
```

This makes debugging difficult when ping tests fail. Consider adding at least a debug-level log:
```python
except Exception as e:
    print(f"[ERROR] Ping test failed: {e}", file=sys.stderr)
    result_dict[pair_key]["status"] = False
```

ygwpz (commented):
## Follow-up Review Summary

### ✅ Previous Issue Addressed
The encoding issue with the corrupted arrow character ("὆~R" instead of "→") has been fixed by replacing it with the word "to". Good fix!

### 🟠 New Issues Found
I found a few issues in the newly introduced bandwidth check implementation:

1. **L3 - Global multiprocessing.Manager() at module level**: The `dump_bw_list` and `load_bw_list` are created at module level, causing unnecessary server process spawn and state accumulation across multiple runs.

2. **L4 - Resource leak**: mmap objects in `make_array()` are not explicitly closed.

3. **L4 - Exception swallowed**: In `remote_local_hccn_cards_ping_test`, exception is caught but not logged.

4. **Minor - Unused import**: The `threading` module is imported but no longer used.

### 💡 Suggestions
- Consider making bandwidth test parameters configurable via `config.yaml`
- Add cleanup handling for mmap objects

See inline comments for specific code locations.

test/common/envPreCheck/run_env_preCheck.py
Menglths: Moved multiprocessing.Manager and shared lists inside run_bandwidth_check() instead of defining them at module level.
Avoids spawning the manager process on module import.
Creates new empty lists for each run to prevent data accumulation.
Ensures proper cleanup of manager resources after execution.

test/common/envPreCheck/run_env_preCheck.py
Menglths: Explicitly store all mmap handles and close them after test execution to fix resource leak. The mmap is kept open during the whole test as the arrays are continuously used.

test/common/envPreCheck/run_env_preCheck.py
Menglths: Added error log for caught exception to avoid swallowing exception details, which helps troubleshoot test failures.

test/common/envPreCheck/run_env_preCheck.py
Menglths: All parameters have been configured via config.yaml.

test/common/envPreCheck/run_env_preCheck.py
```diff
@@ -1119,282 +1145,133 @@ def run_check_model_weight():
...
 
-    This is useful for capturing logs programmatically for filtering or analysis.
-    """
 
-    def __enter__(self):
-        # Save original stdout and stderr file descriptors
-        self.original_stdout = os.dup(1)
-        self.original_stderr = os.dup(2)
+def load(epoch, device_id, worker, block_ids, block_ptr, load_bw_list):
+    total_size = SHARD_SIZE * SHARD_NUMBER * block_number
```
ygwpz: 🔴 **Critical Bug**: `block_number` is undefined in this function. The global constant is defined as `BLOCK_NUMBER` (uppercase) at the top of the file. This will cause a `NameError` when the `load` function is called. Fix: change `block_number` to `BLOCK_NUMBER`.

test/common/envPreCheck/run_env_preCheck.py
```diff
@@ -1119,282 +1145,133 @@ def run_check_model_weight():
...
+    costs = []
+    for i in range(SHARD_NUMBER):
+        idxes = [i for _ in range(block_number)]
```
ygwpz: 🔴 **Critical Bug**: Same issue - `block_number` is undefined. Use `BLOCK_NUMBER` instead.

Menglths: The problem has been resolved.

ygwpz (approved):
The environment pre-check documentation and script updates look reasonable.

PR #1009, opened 2026-06-09
[Feat] Add Cache IO aggregation for Ascend

NaganooMei:
## Summary

- Add an Ascend Cache IO aggregation executor for shard-level H2D and D2H transfers.
- Add the Ascend shard IO aggregator and FFTS D2D dispatcher used by the aggregated executor.
- Wire CacheStore load and dump queues through the new executor abstraction while keeping the tensor executor as the default path.
- Add build and setup switches so upstream builds do not compile IO aggregation unless explicitly enabled.
- load path: 
image
## Validation

| Scenario | Baseline | With IO Aggregation | Result |
|---|---:|---:|---|
| GQA model, 32K IO load/dump | ~4 GB/s per device × 8 | ~22 GB/s per device × 8 | ~5.75× bandwidth improvement |
| DeepSeek-V4 CacheStore IO | ~3–4 GB/s per device × 8 | ~8–10 GB/s per device × 8 | ~2–3× bandwidth improvement |
| Default upstream build | IO aggregation disabled | Not compiled unless explicitly enabled | Default tensor executor path preserved |
| Code check | `git diff --check upstream/develop..HEAD` | Passed | No whitespace/check-format issues |


ucm/shared/trans/ascend/ascend_shard_io_aggregator.cc
```diff
@@ -0,0 +1,320 @@
...
+        s = AclStatus(aclrtSynchronizeStream(lane.fftsStream), "aclrtSynchronizeStream(ffts)");
+        if (s.Failure()) { return s; }
+        lane.inFlight.clear();
+    }
+    return Status::OK();
+}
+
+void AscendShardIOAggregator::Cleanup() noexcept
+{
+    if (deviceId_ >= 0) { (void)aclrtSetDevice(deviceId_); }
```
ygwpz, line 294: 💡 **Suggestion:** After `aclrtSetDevice`, the code should check for errors before proceeding. The current implementation checks this, which is good.

ucm/shared/trans/ascend/ffts_d2d_dispatcher.cc
```diff
@@ -0,0 +1,154 @@
...
+
+void FftsD2DDispatcher::Reset()
+{
+    contexts_.clear();
+    completed_ = false;
+}
+
+Status FftsD2DDispatcher::AddMemcpy(void* dst, const void* src, size_t size)
+{
+    if (completed_) { return Status::Error("FFTS dispatcher already launched"); }
```
ygwpz, line 35: 💡 **Suggestion:** The `completed_` flag prevents adding new copies after launch. This is a good safety check. Consider also resetting this flag in `Reset()` to allow reuse of the dispatcher.

ucm/store/cache/cc/cache_io_executor.cc
```diff
@@ -0,0 +1,144 @@
...
+        return aggregator_.SubmitDumpObject(devices, host, tensorSizes_);
+    }
+    Status Synchronize() override { return aggregator_.Synchronize(); }
+};
+#else
+class UnavailableAggregatedIOExecutor : public CacheIOExecutor {
+public:
+    Status Setup(const Config&) override
+    {
+        return Status::InvalidParam("Cache IO aggregation is not compiled");
```
ygwpz, line 113: 💡 **Suggestion:** The `UnavailableAggregatedIOExecutor` returns `Status::InvalidParam` for all methods when IO aggregation is not compiled. Since this is not a parameter validation error (the parameters may be perfectly valid), consider using `Status::Unsupported()` instead. This better communicates that the feature is simply not available in this build configuration.

The `Status::Unsupported()` method is available in `ucm/shared/infra/status/status.h` and maps to error code -50008, which the troubleshooting documentation describes as "Feature or operation not supported in current context" - exactly matching this scenario.

ygwpz (approved):
Cache IO aggregation for Ascend - follow-up review after minor update.

PR #1010, opened 2026-06-09
[Bugfix] Fix ASU SQE flag buffer calculation, send buffer default slots, and add debug/error logs.

yuanzhg078:
## Purpose
**Bugfix:** Fix ASU SQE response flag buffer sizing so each opcode allocates enough completion status space without over-allocating, and keep submit-flow logging focused on actionable errors.
## Modifications 
- Fix SQE flag buffer sizing by opcode: batch store/retrieve use 4-bit entry status, while delete/exist use 1-bit key status.
- Pass KvOpcode into flag-buffer allocation so each request type allocates the correct response buffer after the common header.
- Keep submit/SQE hot-path logs error-only and add diagnostics for allocation, pack, validation, and send failures.
- Update delete/exist SQE tests for compact multi-key flag buffer sizing.
- Change default ASU send buffer slot size/count for SQE submission.
## Test
Covered paths are validated through ASU SQE request unit tests for delete/exist flag buffer sizing and request submission behavior.

### DIFF
--- a/ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
+++ b/ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
@@ -335,9 +335,13 @@ Status AsuTransportImpl::AssignSubBatchConnections(
 
 void AsuTransportImpl::ProcessTask(const TransportTaskContextPtr& ctx)
 {
+    UC_DEBUG("AsuTransportImpl::ProcessTask start task_id={} op_type={} entries={} keys={}",
+             ctx->taskId, static_cast<int>(ctx->opType), ctx->entries.size, ctx->keys.size);
     TransportTaskState expected = TransportTaskState::PENDING;
     if (!ctx->state.compare_exchange_strong(expected, TransportTaskState::INFLIGHT,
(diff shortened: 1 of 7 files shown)


ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -346,19 +350,58 @@ void AsuTransportImpl::ProcessTask(const TransportTaskContextPtr& ctx)
 
     std::vector<TransportSubBatchContext> subBatchContexts;
     auto finalStatus = SubmitTaskRequests(*ctx, subBatchContexts);
+    UC_DEBUG(
+        "AsuTransportImpl::ProcessTask submit done task_id={} sub_batches={} code={} "
+        "message={}",
+        ctx->taskId, subBatchContexts.size(), static_cast<int>(finalStatus.code),
+        finalStatus.message);
+    if (!finalStatus.ok()) {
```
Infinite666, line 358: Do we still need to run the rest of codes if some error happens?

Infinite666 (approved):

PR #1013, opened 2026-06-10
[Fix] kv-test adaptations and fixes

Fengli5355:
## Purpose
Adapt to previous code changes of previous pull requests.

## Modifications 
1. Change signature of TransProvider and Send interface.
2. Use a hook to trigger fake backend transport in kv-test.
3. Modify CMakeLists to introduce dependency for kv-test.
4. AsuId now can be correctly parsed.
5. Use a placeholder in AICPUTransProvider::CreateConnection to avoid exitting.
6. Fix CompleteTask logic when query get non-existence results. Queries which have status ASU_CQE_CHECK_RESULT_BUFFER will call CompleteSubBatch with status OK to avoid PARTIAL_FAILED.
7. Modify kv-test configs.
8. Use effectiveOptions to merge cli command and config parameters.
9. Fix the logic of LocalAsuTransport::DeleteAsync, where status is OK when keys deleted do not exist.
10. Modify type cast for error code.
11. Modify domument of kv-test.
12. Remove useless struct and configs.

### DIFF
--- a/ucm/transport/kv/asu/trans/src/aicpu_trans_provider.h
+++ b/ucm/transport/kv/asu/trans/src/aicpu_trans_provider.h
@@ -4,11 +4,24 @@
 
 namespace UC::ASU {
 
+using AICPUTransProviderSendHook =
+    std::vector<Status> (*)(const std::vector<TransProvider::SendIoBatch>& ioBatches,
+                            uint32_t kernelCount, uint32_t quietCount);
+
+void SetAICPUTransProviderSendHook(AICPUTransProviderSendHook hook);
+AICPUTransProviderSendHook GetAICPUTransProviderSendHook();
+
 class AICPUTransProvider : public TransProvider {
 public:
-    Status CreateConnection(const std::string&, const std::string&, uint32_t, uint32_t, uint32_t,
-                            std::vector<ConnectionHandle>&) override
+    Status CreateConnection(const std::string&, const std::string&, uint32_t, uint32_t qpNum,
+                            uint32_t, std::vector<ConnectionHandle>& handles) override
     {
+        handles.clear();
+        handles.reserve(qpNum);
+        for (uint32_t index = 0; index < qpNum; ++index) {
+            handles.push_back(reinterpret_cast<ConnectionHandle>(
+                static_cast<std::uintptr_t>(index) + static_cast<std::uintptr_t>(1)));
+        }
         return Status::OK();
     }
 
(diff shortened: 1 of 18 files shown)


ucm/transport/kv/asu/trans/src/aicpu_trans_provider.h
```diff
@@ -4,11 +4,24 @@
...
 public:
-    Status CreateConnection(const std::string&, const std::string&, uint32_t, uint32_t, uint32_t,
-                            std::vector<ConnectionHandle>&) override
+    Status CreateConnection(const std::string&, const std::string&, uint32_t, uint32_t qpNum,
+                            uint32_t, std::vector<ConnectionHandle>& handles) override
     {
+        handles.clear();
+        handles.reserve(qpNum);
+        for (uint32_t index = 0; index < qpNum; ++index) {
+            handles.push_back(reinterpret_cast<ConnectionHandle>(
```
ygwpz, line 22: ⚠️ **Warning:** Creating placeholder connection handles by casting integers to pointer types (`reinterpret_cast<ConnectionHandle>(...)`) is unusual. If these handles are dereferenced elsewhere in the code, this could cause undefined behavior. Ensure that all code paths using these handles only treat them as opaque identifiers and never dereference them.

ucm/transport/kv/kv-test/src/fake_backend.cpp
```diff
@@ -365,16 +379,22 @@ FakeBackendConfig GetFakeBackendConfig(bool& enabled)
...
 {
-    std::lock_guard<std::mutex> lock(g_fakeBackendMu);
-    g_fakeBackendConfig = std::move(config);
-    g_fakeBackendEnabled = true;
+    {
+        std::lock_guard<std::mutex> lock(g_fakeBackendMu);
+        g_fakeBackendConfig = std::move(config);
+        g_fakeBackendEnabled = true;
+    }
+    UC::ASU::SetAICPUTransProviderSendHook(&UC::ASU::MockSend);
```
ygwpz, line 387: 💡 **Suggestion:** The hook is set after the lock is released. There's a potential race window where `g_fakeBackendEnabled` is true but the hook is still `nullptr`. If `Send()` is called during this window, it will use the default behavior instead of `MockSend`. Consider setting the hook before setting `g_fakeBackendEnabled = true`, or document this ordering assumption.

ucm/transport/kv/kv-test/src/asu_client_runner.cpp
```diff
@@ -109,10 +109,10 @@ Status SubmitEntriesOneByOne(UC::ASU::AsuClient& client,
         }
 
         if (!status.Ok() && firstFailure.ok()) {
-            firstFailure = singleResult.taskResult.status.ok()
-                               ? UC::ASU::Status::Error(
-                                     static_cast<UC::ASU::StatusCode>(status.code), status.message)
-                               : singleResult.taskResult.status;
+            firstFailure =
+                singleResult.taskResult.status.ok()
+                    ? UC::ASU::Status::Error(UC::ASU::StatusCode::INTERNAL_ERROR, status.message)
```
ygwpz, line 114: 💡 **Suggestion:** The original code preserved the status code from `status.code`, but this change always uses `INTERNAL_ERROR`. Was this intentional? If the original status code carried semantic meaning (e.g., timeout, invalid param), this change may lose that information in error reporting.

ucm/transport/kv/asu/trans/src/aicpu_trans_provider.h
Fengli5355: This is temprary implmentation with mock send interfaces in the current version, where handles will not be used.

Infinite666 (approved):

PR #1014, opened 2026-06-10
[Feat] Add RDMA memory registration to BufferManager

yumingyue624:
## Purpose
Register send/flag buffers to RDMA during BufferManager::Init, obtain
tokenId (rkey) via GetMemTokenId, and use it in SQE requests instead
of the previous placeholder lkey.

## Modifications
1. BufferManager: Init() accepts optional TransProvider* and
ConnectionHandle; after allocation, calls RegisterRdma() to register
the entire buffer as a single MR and obtain tokenId; destructor calls
UnregisterMemory for cleanup. ScatterGatherEntry::lkey renamed to
tokenId. Added GetTokenId() accessor.
2. AsuTransportImpl::Init: passes transProvider_ and a ConnectionHandle
from connManager_ to both buffer managers' Init.
3. sqe_request.cpp: flagBuffer.lkey -> flagBuffer.tokenId (5 places).
4. buffer_manager_test.cc: adapted lkey assertion to tokenId.

## Test
- Existing buffer_manager_test cases remain valid

### DIFF
--- a/ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
+++ b/ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
@@ -37,10 +37,6 @@
 
 namespace UC::ASU {
 
-namespace {
-
-}  // namespace
-
 AsuTransportImpl::~AsuTransportImpl() { Shutdown(); }
 
 Status AsuTransportImpl::Init(const std::string& configPath)
@@ -102,12 +98,14 @@ Status AsuTransportImpl::Init(const TransportConfig& config)
     auto status = ValidateSqeRequestAttrs();
     if (!status.ok()) { return status; }
 
-    status = sendBufferManager_.Init("asu send buffer", MemoryType::HOST,
-                                     config_.sendBufferSlotSize, config_.sendBufferSlotNum);
+    status =
+        sendBufferManager_.Init("asu send buffer", MemoryType::HOST, config_.sendBufferSlotSize,
+                                config_.sendBufferSlotNum, transProvider_.get());
     if (!status.ok()) { return status; }
 
-    status = flagBufferManager_.Init("asu flag buffer", MemoryType::HOST,
-                                     config_.flagBufferSlotSize, config_.flagBufferSlotNum);
+    status =
+        flagBufferManager_.Init("asu flag buffer", MemoryType::HOST, config_.flagBufferSlotSize,
+                                config_.flagBufferSlotNum, transProvider_.get());
     if (!status.ok()) { return status; }
     protocolManager_ = std::make_unique<ProtocolManager>();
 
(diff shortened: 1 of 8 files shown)


ucm/transport/kv/asu/trans/src/buffer_manager.cpp
```diff
@@ -79,6 +85,44 @@ Status BufferManager::Init(std::string name, MemoryType type, std::size_t slot_s
...
+        if (!regStatus.ok()) {
+            memory_.reset();
+            return regStatus;
+        }
+    }
+
+    return Status::OK();
+}
+
+Status BufferManager::RegisterRdma(TransProvider* provider,
```
yuanzhg078: Rename RegisterRdma to a protocol-agnostic name since the function supports multiple protocols, not just RDMA.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -102,12 +102,18 @@ Status AsuTransportImpl::Init(const TransportConfig& config)
...
     if (!status.ok()) { return status; }
 
+    TransProvider::ConnectionHandle connHandle = nullptr;
+    auto channel = connManager_->SelectConnection();
+    if (channel) { connHandle = channel->GetConnection(); }
+
     status = sendBufferManager_.Init("asu send buffer", MemoryType::HOST,
-                                     config_.sendBufferSlotSize, config_.sendBufferSlotNum);
+                                     config_.sendBufferSlotSize, config_.sendBufferSlotNum,
+                                     transProvider_.get(), connHandle);
```
yuanzhg078: Buffer manager initialize no need connHandle and should not be coupled with any specific channel.

ucm/transport/kv/asu/trans/src/buffer_manager.cpp
yumingyue624: Done. Renamed RegisterRdma to RegisterMemory to align with the underlying TransProvider::RegisterMemory interface, which is protocol-agnostic and will support both RDMA and UB.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
yumingyue624: Done. Removed connHandle from BufferManager::Init() — it now only takes TransProvider* and passes nullptr to RegisterMemory/UnregisterMemory internally. Buffer manager is no longer coupled with any specific channel.

ucm/transport/kv/asu/trans/src/buffer_manager.cpp
```diff
@@ -79,6 +85,44 @@ Status BufferManager::Init(std::string name, MemoryType type, std::size_t slot_s
 
     index_pool_.Setup(static_cast<IndexPool::Index>(slot_num));
 
+    if (provider) {
```
Infinite666, line 88: Why not setting provider_ here and RegisterMemory can use it directly without passing it as input.

Infinite666 (approved):

PR #1018, opened 2026-06-11
[Feat] Asu store adaptation

Fengli5355:
## Purpose
Adapt AsuStore into Connector in order to run an e2e inference.

## Modifications 
1. Adaptations for AsuStore:
(1) Add a .yaml config file;
(2) Modify build/compile scripts;
(3) Define "shard_size" and "tensor_size" the same as "block_size" for AsuStore;
(4) Introduce fakebackend in AsuTransport and AsuStore;

2. Fixes of FakeBackend:
(1) FakeBackend Adds deviceId config;
(2) FakeBackend Adds aclrtInit and aclrtSetDevice. (Reset and Finalize are not added yet)
(3) Correct batchNum value in FakeBackend::CompleteDelete and CompleteExist;
(4) Correct direct NPU I/O in FakeBackend, and use aclrtMemcpy instead;
(5) Correct FakeBackend::Exist which lacks existing_key_number.

3. Add some error logs for AsuClient.

4. Change the way of hashing to try to avoid hash collision.

5. Adapt to BufferManager changes PR#1014.

6. Fix of AsuStore:
Correct AsuStore::Dump by adding WaitPrerequisiteEvent before submitting the task.

7. Further fixes:
(1) Rename config file as ucm_config_asu.yaml;
(2) AsuStore makes key hash only by blockId now;
(3) Move PatchFakeBackendTransportConfig into asu_store.cpp:BuildTransportConfig where transProvider has type FAKE. 
(4) Remove fake_backend.h/cpp;
(5) Simplify fake_trans_provider.cpp when making FakeTransProviderConfig.

## Test
`examples/offline_inference.py` with asu store type shares the same output with cache|posix. (Basic correctness)
"Xiamen University" Example test passes. (Single concurrency)
multifieldqa-zh.jsonl produces similar outputs in 2 runs in a row. (32 concurrency）


ucm/transport/kv/asu/trans/include/asu_transport/fake_backend.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace UC::ASU {
+
+struct FakeBackendConfig {
+    std::string storePath;
+    std::uint64_t latencyMs{1};
+    std::int32_t deviceId{0};
+};
+
+Status EnableFakeBackend(FakeBackendConfig config);
```
Infinite666: In our future structure, we will at least have aicpu and aiv two kinds of trans provider, so can we code fake backend also as a fake trans provider and use config to determine which provider will we use?

Fengli5355: FakeTransProvider has been extracted as an individual class. And AICPUTransProvider is revised back to placeholder.

ucm/store/asu/cc/asu_store.cc
```diff
@@ -78,6 +101,35 @@ void LogAsuStatus(const char* operation, const AsuStatus& status)
...
+}
+
+UC::ASU::TransProviderType ParseTransProviderBackend(std::string backend)
+{
+    std::transform(backend.begin(), backend.end(), backend.begin(),
+                   [](unsigned char ch) { return static_cast<char>(std::toupper(ch)); });
+    if (backend == "FAKE") { return UC::ASU::TransProviderType::FAKE; }
+    if (backend == "AIV") { return UC::ASU::TransProviderType::AIV; }
+    if (backend == "AICPU") { return UC::ASU::TransProviderType::AICPU; }
+    throw std::invalid_argument("invalid ASU trans provider backend: " + backend);
```
yuanzhg078: Suggest returning Status::InvalidParam(...) here instead of throwing std::invalid_argument, to keep config parsing consistent with the rest of asu_store.cc and avoid exception-based control flow for invalid user config.

Fengli5355: Done. Add an UNSUPPORTED type and remove unused throw-catches.

ucm/transport/kv/asu/trans/src/fake_backend.cpp
```diff
@@ -0,0 +1,53 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ * */
+#include "asu_transport/fake_backend.h"
+#include <string>
+#include <utility>
+
+namespace UC::ASU {
+
+void PatchFakeBackendTransportConfig(TransportConfig& config, const FakeBackendConfig& fakeConfig)
```
Infinite666: Is this function neccessary? Almost all the normal configs won't be used in Fake trans provider. And it's weird to put it in this single file.

examples/ucm_config_asu.yaml
```diff
@@ -0,0 +1,33 @@
+# UCM ASU fake backend example for vLLM / vLLM-Ascend software integration tests.
```
Infinite666, line 1: The name of this config file should not be fakexxx

ucm/store/asu/cc/asu_store.cc
```diff
@@ -464,7 +586,7 @@ class AsuStore final : public StoreV1 {
         std::vector<UC::ASU::CacheKey> keys;
         keys.reserve(num);
         for (std::size_t blockIndex = 0; blockIndex < num; ++blockIndex) {
-            keys.emplace_back(ToHex(blocks[blockIndex]));
+            keys.emplace_back(MakeAsuKey(blocks[blockIndex], 0, 0));
```
Infinite666: Make sure the hash key can be always the same in here and BuildKvBuffers

ucm/store/asu/cc/asu_store.cc
```diff
@@ -501,7 +623,7 @@ class AsuStore final : public StoreV1 {
             }
             for (std::size_t tensorIndex = 0; tensorIndex < shard.addrs.size(); ++tensorIndex) {
                 UC::ASU::KVBuffer entry;
-                entry.key = ToHex(shard.owner);
+                entry.key = MakeAsuKey(shard.owner, shard.index, tensorIndex);
```
Infinite666: Try to find out if we need shard and tensor index.

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
```diff
@@ -0,0 +1,466 @@
...
+constexpr std::uint16_t kCqeSuccess = 0x000;
+constexpr std::uint16_t kCqeCheckResultBuffer = 0x732;
+constexpr std::uint8_t kBatchEntryOk = 0x0;
+constexpr std::uint8_t kBatchEntryKeyNotFound = 0x3;
+constexpr std::uint8_t kDeleteEntryOk = 0x0;
+constexpr std::uint8_t kDeleteEntryFailed = 0x1;
+constexpr std::uint8_t kExistEntryNotExist = 0x0;
+constexpr std::uint8_t kExistEntryExist = 0x1;
+
+std::filesystem::path StoreRoot(const FakeBackendConfig& config)
```
Infinite666: The config is being passing all the way down to here, try to code cleaner and simpler.

ucm/store/asu/cc/asu_store.cc
Fengli5355: Shard and tensor are unnecessary. It is already constrained that tensor_size, block_size and shard_size should be the same in `ucm_connector.py`.

ucm/transport/kv/asu/trans/src/fake_backend.cpp
Fengli5355: This part has been removed.

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
Fengli5355: Done

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
```diff
@@ -0,0 +1,461 @@
...
+        fakeConfig.deviceId = static_cast<std::int32_t>(std::stol(deviceIter->second));
+    }
+    return fakeConfig;
+}
+
+FakeTransProvider::FakeTransProvider(FakeTransProviderConfig config) : config_(std::move(config)) {}
+
+Status FakeTransProvider::SetUpAclRuntime()
+{
+    if (aclReady_) { return Status::OK(); }
```
ygwpz, line 373: 🔴 **Concurrency Issue: Thread Safety on aclReady_ flag**

The `aclReady_` member variable is checked and set without synchronization in `SetUpAclRuntime()`. If multiple threads call `CreateConnection()` concurrently (e.g., from different transport instances), they could race on this flag:

1. Thread A checks `aclReady_` (false), proceeds to call `aclInit`
2. Thread B checks `aclReady_` (false), also proceeds to call `aclInit`
3. Both threads call `aclrtSetDevice` concurrently
4. Thread A sets `aclReady_ = true`
5. Thread B sets `aclReady_ = true`

While `aclInit` handles `ACL_ERROR_REPEAT_INITIALIZE`, the `aclrtSetDevice` call is not idempotent and could cause issues when called concurrently. Additionally, the plain `bool aclReady_` is not atomic, leading to potential data races.

**Recommendation**: Use `std::mutex` or `std::atomic<bool>` with proper synchronization to protect the ACL initialization sequence.

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
```diff
@@ -0,0 +1,461 @@
...
+    PackCqeHeader(flagBuffer, cid, cqeStatus);
+    flagBuffer[0] = existingKeyNumber;
+    if (!allExist) { PackResultBuffer1Bit(flagBuffer + kCqeDwordCount, results); }
+    return Status::OK();
+}
+
+Status CompleteFakeBackendRequest(const FakeTransProviderConfig& config, const void* sendBuffer,
+                                  std::uint64_t len)
+{
+    if (config.latencyMs > 0) {
```
ygwpz, line 322: 💡 **Performance: Blocking sleep in Send path**

The `std::this_thread::sleep_for()` call blocks the calling thread for the simulated latency duration. In high-concurrency scenarios, this could significantly reduce throughput as the transport Send thread is blocked during each operation.

For a test/fake backend, this may be acceptable, but consider:
1. Using a thread pool to process requests asynchronously
2. Using a timer-based completion mechanism instead of blocking
3. Setting `latencyMs=0` for high-concurrency tests

This is especially relevant when `batchNumber` is large, as each batch entry processing also includes device I/O operations.

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
```diff
@@ -0,0 +1,461 @@
...
+{
+    FakeTransProviderConfig fakeConfig;
+    fakeConfig.deviceId = config.endpoints.empty() ? 0 : config.endpoints.front().deviceId;
+    auto pathIter = config.attrs.find("fake_backend.path");
+    if (pathIter != config.attrs.end() && !pathIter->second.empty()) {
+        fakeConfig.storePath = pathIter->second;
+    }
+    auto latencyIter = config.attrs.find("fake_backend.latency_ms");
+    if (latencyIter != config.attrs.end()) {
+        fakeConfig.latencyMs = static_cast<std::uint64_t>(std::stoull(latencyIter->second));
```
ygwpz, line 360: **Warning: Error Handling - std::stoull/std::stol without exception handling**

The calls to std::stoull() and std::stol() on lines 360 and 364 are not wrapped in try-catch blocks. If the configuration string contains invalid values (e.g., non-numeric characters, empty string, or out-of-range values), these functions will throw std::invalid_argument or std::out_of_range exceptions, which would crash the application.

**Recommendation**: Wrap these calls in try-catch blocks and return a proper error status if parsing fails.

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
```diff
@@ -0,0 +1,461 @@
...
+    }
+    return fakeConfig;
+}
+
+FakeTransProvider::FakeTransProvider(FakeTransProviderConfig config) : config_(std::move(config)) {}
+
+Status FakeTransProvider::SetUpAclRuntime()
+{
+    if (aclReady_) { return Status::OK(); }
+    auto ret = aclInit(nullptr);
```
ygwpz, line 374: **Warning: Architecture - Missing ACL runtime cleanup**

The FakeTransProvider calls aclInit() and aclrtSetDevice() in SetUpAclRuntime(), but there is no corresponding cleanup in the destructor or shutdown path:

- Missing aclFinalize() to match aclInit()
- Missing aclrtResetDevice() to match aclrtSetDevice()

This could lead to ACL runtime resource leaks, especially if multiple FakeTransProvider instances are created and destroyed.

**Recommendation**: Add cleanup in destructor or provide a Shutdown() method. Note that aclFinalize should typically be called once globally, so consider using a shared ACL runtime manager or static cleanup.

ucm/store/asu/cc/asu_store.cc
```diff
@@ -98,13 +168,37 @@ UC::ASU::TransportConfig BuildTransportConfig(const Config& config, std::size_t
...
     if (!config.asuIps.empty()) {
         UC::ASU::AsuEndpoint endpoint;
         endpoint.ip = config.asuIps[index];
         endpoint.port = config.asuPort;
         endpoint.deviceId = config.deviceId;
         transportConfig.endpoints.emplace_back(std::move(endpoint));
     }
+    if (config.transProviderType == UC::ASU::TransProviderType::FAKE) {
+        const auto fakeDeviceId = config.deviceId >= 0 ? config.deviceId : 0;
+        transportConfig.providerType = UC::ASU::TransProviderType::FAKE;
```
ygwpz: **Suggestion: Redundant providerType assignment**

The providerType is set twice in BuildTransportConfig:
1. Line 171: transportConfig.providerType = config.transProviderType;
2. Line 181: transportConfig.providerType = UC::ASU::TransProviderType::FAKE; (inside the FAKE block)

The second assignment (line 181) is redundant because line 171 already sets it correctly. Inside the FAKE block, the config.transProviderType should already be FAKE (as the condition checks), so this line could be removed.

Alternatively, if this is intentional for clarity, consider adding a comment explaining why the reassignment is needed.

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
Fengli5355: This varibale will only be visited in single thread for each Transport.

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
Fengli5355: This is just fake transport to mock the actual send process. Thus performance is not the first to consider.

ucm/store/asu/cc/asu_store.cc
Fengli5355: Done.

Infinite666 (approved):

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
```diff
@@ -0,0 +1,461 @@
...
+    const auto cid = static_cast<std::uint16_t>(RequestCid(request));
+    const auto responseBufferAddr = ReadU64(request[3], request[4]);
+    const auto batchNumber = static_cast<std::uint16_t>(request[10] & 0xFFFF);
+    auto* flagBuffer = reinterpret_cast<std::uint32_t*>(responseBufferAddr);
+    std::vector<std::uint8_t> results(batchNumber, kBatchEntryOk);
+
+    const auto entries = ReadBatchEntries(request, batchNumber);
+    for (std::size_t index = 0; index < entries.size(); ++index) {
+        const auto& entry = entries[index];
+        if (!StoreBytes(config, asuId, entry.key, entry.bufferAddr, entry.length)) {
```
ygwpz, line 233: ⚠️ **Warning:** When `StoreBytes` fails, the error is marked as `kBatchEntryKeyNotFound`. However, `StoreBytes` can fail for various reasons: device-to-host copy failure (aclrtMemcpy), file open failure, or write failure. Using `kBatchEntryKeyNotFound` is misleading for these error cases. Consider introducing a distinct error code (e.g., `kBatchEntryStoreFailed`) for non-"key not found" failures to help with debugging.

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
```diff
@@ -0,0 +1,461 @@
...
+    const auto cid = static_cast<std::uint16_t>(RequestCid(request));
+    const auto responseBufferAddr = ReadU64(request[3], request[4]);
+    const auto batchNumber = static_cast<std::uint16_t>(request[10] & 0xFFFF);
+    auto* flagBuffer = reinterpret_cast<std::uint32_t*>(responseBufferAddr);
+    std::vector<std::uint8_t> results(batchNumber, kBatchEntryOk);
+
+    const auto entries = ReadBatchEntries(request, batchNumber);
+    for (std::size_t index = 0; index < entries.size(); ++index) {
+        const auto& entry = entries[index];
+        if (!LoadBytes(config, asuId, entry.key, entry.bufferAddr, entry.length)) {
```
ygwpz, line 258: ⚠️ **Warning:** Similar to `CompleteBatchStore` above - `LoadBytes` failure is marked as `kBatchEntryKeyNotFound`, but the failure could be from: file not existing (actual key not found), file read failure, or host-to-device copy failure (aclrtMemcpy). Using different error codes would improve error diagnosis.

mag1c-h (approved):

PR #1021, opened 2026-06-11
Add a HIP/ROCm device backend for KV-transfer and Hamming kernels

jeffdaily:
This adds a new RUNTIME_ENVIRONMENT=rocm device backend (PLATFORM=rocm in
setup.py) alongside the existing cuda/ascend/maca/musa/simu backends, so the
KV block-transfer kernels, the H2D/D2H store path, and the sparse Hamming
scoring kernel build and run on AMD GPUs via HIP. The backend is purely
additive: the existing backends are not modified. We have made every effort
to leave the NVIDIA build unchanged -- every source change to a shared file is
behind a `__CUDA_ARCH__` or `USE_ROCM` guard that the CUDA build does not
compile, and the compat shim is only placed on the include path for the rocm
build, never the cuda one. Select it with `export PLATFORM=rocm` (or
`-DRUNTIME_ENVIRONMENT=rocm` when invoking CMake directly).

Review order: start with ucm/shared/vendor/hip_compat/ (the compat shim), then
the three new rocm/CMakeLists.txt arms (trans, store, sparse ham_dist), then
the two guarded kernel sources, then the docs and Windows host-build guards.

The compat shim resolves the project's `#include <cuda_runtime.h>`/`<cuda.h>`
to <hip/hip_runtime.h> and aliases the small cuda* runtime surface in use
(Malloc/Free/Memcpy[Async]/HostMalloc/HostRegister/Stream*/Event*) to hip*.
Each rocm CMake arm calls enable_language(HIP)/find_package(hip), reuses the
existing cuda .cc/.cu sources marked LANGUAGE HIP, and reads
CMAKE_HIP_ARCHITECTURES (defaulting to gfx90a only when unset) so other AMD
targets need no source edit.

Root cause of the one non-mechanical change: the two grid-stride copy kernels
used inline PTX (ld.global.cs / st.volatile.global vectorized loads/stores)
that does not exist on AMD. The PTX is now guarded by
`#if defined(__CUDA_ARCH__)` with a HIP branch doing a plain vectorized uint4
load/store (32-byte and 16-byte units). ROCm 7.2.1 does not provide the
__ldcs/__stcg/__stcs cache-streaming builtins, and those PTX qualifiers are
cache-policy hints rather than visibility semantics for this memcpy (each
thread writes a disjoint unit and the only consumer is the host after a stream
sync), so the plain copy is functionally equivalent. FlashInfer's cp_async.cuh
already selects its portable non-PTX fallback under hipcc, so it needed no
change.

The sparse Hamming module links libtorch. operator.h now includes
<ATen/hip/HIPContext.h> under USE_ROCM (the cuda-spelled context header pulls
in NVIDIA-only cuda_runtime_api.h/cusparse.h, while the hipified header exposes
the same c10::cuda::getCurrentCUDAStream backed by HIP). It builds at C++20
(torch 2.x headers use requires-clauses), without -ffast-math (the kernel uses
INFINITY as a masking sentinel that finite-math would drop), and with
pybind11_add_module(NO_EXTRAS) to avoid pybind's default LTO+strip dropping the
module init symbol under -fvisibility=hidden.

A set of WIN32-guarded host-build fixes let the backend also compile with the
clang-cl toolchain on Windows ROCm (Linux-only compiler/linker flags guarded,
three header-only infra sub-libraries changed from OBJECT to INTERFACE so the
linker language is determinable under Ninja+clang-cl, a getpid shim, metrics
symbol export, and excluding a POSIX-only thread test). The GPU device code is
unchanged by those guards.

Docs: the supported-platform matrix gains a ROCm/AMD row, and the vLLM and
SGLang quickstarts document `PLATFORM=rocm` beside `PLATFORM=cuda`.

This work was authored with the assistance of Claude, an AI assistant.

Test Plan:
Built and validated on real GPUs: Linux gfx90a (MI250X) and gfx1100 (Radeon
Pro W7800), and Windows gfx1201 (RX 9070 XT), gfx1101 (Radeon PRO V710), and
gfx1151 (Radeon 8060S), all on ROCm. All GPU tests run serially with one GPU
visible (HIP_VISIBLE_DEVICES=0).

Store/trans C++ surface plus unit tests:
```
cmake -S . -B build_rocm -DRUNTIME_ENVIRONMENT=rocm -DBUILD_UCM_STORE=ON \
  -DBUILD_UNIT_TESTS=ON -DCMAKE_HIP_ARCHITECTURES=gfx90a \
  -DCMAKE_HIP_COMPILER=/opt/rocm/llvm/bin/clang++ -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_CXX_FLAGS="-Wno-error=unused-result"
cmake --build build_rocm -j16
HIP_VISIBLE_DEVICES=0 ctest --test-dir build_rocm -j1
```
79/80 pass on Linux. The copy-kernel correctness gates all pass:
UCTransUnitTest.{CopyDataWithCE,CopyDataWithSM,CopyDataBatchWithSM} (byte-exact
host<->device round-trip) and the UCPosixTrans*/UCCacheTransBuffer cases
(store H2D/D2H batch copy with readback). The one failure,
UCMetricsUT.ConcurrentUpdateAndCollect, is a pre-existing CPU-only
multi-threaded metrics counter test (untouched ucm/shared/metrics), unrelated
to the GPU backend.

Hamming kernel (against a ROCm PyTorch):
```
cmake -S . -B build_sparse -DRUNTIME_ENVIRONMENT=rocm -DBUILD_UCM_STORE=OFF \
  -DBUILD_UCM_SPARSE=ON -DCMAKE_HIP_ARCHITECTURES=gfx90a \
  -DCMAKE_HIP_COMPILER=/opt/rocm/llvm/bin/clang++ -DCMAKE_BUILD_TYPE=Release \
  -DPython_EXECUTABLE=<rocm-torch-python> -DCMAKE_CXX_FLAGS="-Wno-error=unused-result"
cmake --build build_sparse -j16 --target hamming
HIP_VISIBLE_DEVICES=0 HAMMING_DIR=<dir with hamming*.so> \
  python ucm/sparse/test/gsa/test_hamming_rocm_ref.py
```
The new test computes an independent CPU popcount reference for the paged
block-mode score: mla (no kv reduction) matches within fp16 rounding, gqa
(min over kv heads) matches exactly, two-run output is bit-identical.

The NVIDIA path was checked to be unaffected: with PLATFORM=cuda the guards
select the original inline-PTX branch (compile-checked with nvcc 12.8 at
sm_80; the emitted PTX still contains the ld.global.cs / st.volatile.global
streaming ops), so the CUDA build compiles the same device code as before.


setup.py
```diff
@@ -174,6 +174,8 @@ def build_cmake(self, ext: CMakeExtension):
         match PLATFORM:
             case "cuda":
                 cmake_args += ["-DRUNTIME_ENVIRONMENT=cuda"]
+            case "rocm":
```
flesher0813, line 181: setup.py adds PLATFORM=rocm support, but the warning message still lists only cuda/ascend/ascend-a3/musa/maca. Please include rocm there as well.

ucm/shared/trans/cuda/cuda_sm_kernel.cu
```diff
@@ -47,6 +48,15 @@ inline __device__ void CudaCopyUnit(const uint8_t* __restrict__ src,
     asm volatile("st.volatile.global.v4.b32 [%0+16], {%1,%2,%3,%4};"
                  :
                  : "l"(dst), "r"(hi.x), "r"(hi.y), "r"(hi.z), "r"(hi.w));
+#else
+    // HIP/ROCm has no ld.global.cs/st.volatile.global PTX and no __ldcs/__stcg
+    // streaming builtins; those are NVIDIA cache-policy hints, not semantics.
+    // A plain vectorized 32-byte copy is the portable equivalent.
+    const uint4* src4 = reinterpret_cast<const uint4*>(src);
+    uint4* dst4 = reinterpret_cast<uint4*>(const_cast<uint8_t*>(dst));
```
ygwpz, line 64: ⚠️ Warning: The `const_cast<uint8_t*>(dst)` removes the `volatile` qualifier from the destination pointer. The original CUDA implementation uses `st.volatile.global` PTX which ensures proper memory visibility semantics for device-to-host transfers. While the comment correctly notes that cache-policy hints are NVIDIA-specific, the `volatile` qualifier itself has semantic meaning - it prevents compiler optimizations that could reorder or eliminate memory operations. On HIP, consider using `__builtin_nontemporal_store` or ensuring proper memory fence semantics to maintain equivalent visibility guarantees. The current plain copy may be functionally correct for this use case (disjoint writes with stream sync), but the removal of volatile should be explicitly justified in the comment.

ucm/shared/trans/cuda/cuda_sm_kernel.cu
```diff
@@ -47,6 +48,15 @@ inline __device__ void CudaCopyUnit(const uint8_t* __restrict__ src,
...
+    dst4[0] = src4[0];
```
ygwpz, line 65: 💡 Suggestion: The `reinterpret_cast<uint4*>` assumes the source pointer is properly aligned to 16-byte boundary for uint4 access. While this is likely true for the KV cache blocks (which are typically allocated with proper alignment), consider adding an assertion or documentation noting the alignment requirement. Misaligned access could cause undefined behavior or performance degradation on some AMD architectures.

ucm/store/nfsstore/device/cuda/cuda_device.cu
```diff
@@ -32,16 +32,26 @@
 
 inline __device__ void H2DUnit(uint8_t* __restrict__ dst, const volatile uint8_t* __restrict__ src)
 {
+#if defined(__CUDA_ARCH__)
     uint64_t a, b;
     asm volatile("ld.global.cs.v2.u64 {%0, %1}, [%2];" : "=l"(a), "=l"(b) : "l"(src));
     asm volatile("st.global.cg.v2.u64 [%0], {%1, %2};" ::"l"(dst), "l"(a), "l"(b));
+#else
+    *reinterpret_cast<uint4*>(dst) =
```
ygwpz, line 46: ⚠️ Warning: Similar to cuda_sm_kernel.cu, the `reinterpret_cast<uint4*>(dst)` removes both const and volatile qualifiers. The volatile qualifier on the destination pointer in D2HUnit was intentional to ensure proper visibility of writes to host memory. While the stream synchronization after the kernel provides the necessary fence, the explicit volatile in the original code served as a compiler barrier. On HIP, this plain copy should work correctly due to the subsequent sync, but the removal of volatile semantics should be documented as a deliberate design decision rather than an incidental consequence of the shim.

ucm/store/nfsstore/device/cuda/cuda_device.cu
```diff
@@ -32,16 +32,26 @@
...
+        *reinterpret_cast<const uint4*>(const_cast<const uint8_t*>(src));
+#endif
 }
 
 inline __device__ void D2HUnit(volatile uint8_t* __restrict__ dst, const uint8_t* __restrict__ src)
```
ygwpz, line 51: ⚠️ Warning: The `const_cast<uint8_t*>(dst)` removes the volatile qualifier. This mirrors the issue in H2DUnit above. The D2H path writes to host-registered memory, and the volatile qualifier ensured the compiler wouldn't optimize away or reorder these writes. The HIP fallback relies on stream synchronization for correctness, which is valid, but the comment should clarify that volatile semantics are intentionally replaced by stream sync barriers.

ucm/sparse/gsa_on_device/csrc/cuda/ham_dist/operator.h
```diff
@@ -1,7 +1,14 @@
...
+// torch keeps the cuda spelling for its public symbols on ROCm; the hipified
+// context header provides c10::cuda::getCurrentCUDAStream backed by HIP, while
+// the cuda-spelled header pulls in NVIDIA-only cuda_runtime_api.h/cusparse.h.
+#include <ATen/hip/HIPContext.h>
+#else
 #include <ATen/cuda/CUDAContext.h>
 #include <cuda.h>
+#endif
 #include <cuda_runtime.h>
 #include <torch/script.h>
```
ygwpz, line 14: 💡 Suggestion: The `#include <cuda_runtime.h>` after the conditional block relies on the hip_compat shim being on the include path for ROCm builds. This works correctly when the CMake configuration is proper, but could cause confusing build failures if the shim directory is not added to include paths. Consider adding a comment here noting this dependency, or adding an `#ifdef USE_ROCM` guard with an `#error` directive if the shim is missing (e.g., checking for `UNIFIEDCACHE_HIP_COMPAT_CUDA_RUNTIME_H` define).

ucm/shared/vendor/hip_compat/cuda_runtime.h
```diff
@@ -0,0 +1,73 @@
...
+#include <hip/hip_runtime.h>
+
+using cudaError_t = hipError_t;
+using cudaStream_t = hipStream_t;
+using cudaEvent_t = hipEvent_t;
+
+static constexpr hipError_t cudaSuccess = hipSuccess;
+static constexpr hipMemcpyKind cudaMemcpyHostToDevice = hipMemcpyHostToDevice;
+static constexpr hipMemcpyKind cudaMemcpyDeviceToHost = hipMemcpyDeviceToHost;
+static constexpr unsigned int cudaStreamNonBlocking = hipStreamNonBlocking;
```
ygwpz, line 52: 💡 Suggestion: Using `#define` macros for function aliases prevents proper function overload resolution and could cause issues if CUDA APIs add new signatures in future versions. For this limited set of APIs, the macro approach is acceptable. However, consider adding inline wrapper functions instead of macros for type safety and better debugging support. Example: `inline cudaError_t cudaMalloc(void** ptr, size_t size) { return hipMalloc(ptr, size); }` would preserve the cudaError_t return type explicitly.

ucm/shared/trans/rocm/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+if(NOT DEFINED CMAKE_HIP_ARCHITECTURES OR CMAKE_HIP_ARCHITECTURES STREQUAL "")
+    set(CMAKE_HIP_ARCHITECTURES "gfx90a")
```
ygwpz: 💡 Suggestion: The architecture defaults to gfx90a (MI250X/MI210) which is appropriate for datacenter AMD GPUs, but may cause confusion for users with consumer Radeon GPUs (gfx1100, gfx1201, etc.). Consider adding a comment or CMake warning message when defaulting, suggesting users set `CMAKE_HIP_ARCHITECTURES` explicitly. The PR description mentions gfx1100 validation but the default doesn't reflect this broader support.

ucm/sparse/gsa_on_device/csrc/rocm/ham_dist/CMakeLists.txt
```diff
@@ -0,0 +1,115 @@
+message(STATUS "Building ham_dist (hamming HIP/ROCm extension)...")
+
+# The project-global release flags strip all symbols (-s), which would remove
+# the CPython module init entry point from this extension and make it
+# unloadable. Drop -s for this subtree so the .so keeps PyInit_hamming.
+string(REPLACE "-s " "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+string(REGEX REPLACE "(^| )-s$" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+
+if(NOT DEFINED CMAKE_HIP_ARCHITECTURES OR CMAKE_HIP_ARCHITECTURES STREQUAL "")
```
ygwpz: 💡 Suggestion: The `string(REPLACE "-s " "" ...)` and `string(REGEX REPLACE ...)` pattern for removing the `-s` flag is somewhat fragile. It assumes `-s` appears with a trailing space or at the end of the string. If `-s` appears without a space (e.g., `-s -O2`), the first REPLACE handles it, but other edge cases might slip through. Consider using a more robust approach like `string(REGEX REPLACE "-s( |$)" "" ...)` to handle all positions uniformly.

ucm/sparse/test/gsa/test_hamming_rocm_ref.py
```diff
@@ -0,0 +1,189 @@
...
+                continue
+            block_slot = pos // block_size
+            offset = pos % block_size
+            phys = int(bt[bi, block_slot].item())
+            # The kernel reads the key block as (num_kv_head, block_size,
+            # num_chunk): base = phys*(num_kv_head*block_size*num_chunk),
+            # element (kv*block_size + offset)*num_chunk + chunk. Index a flat
+            # view of the key tensor the same way so the reference matches the
+            # kernel's layout regardless of the host tensor's nominal shape.
+            block = key[phys].reshape(-1)
```
ygwpz, line 77: 💡 Suggestion: The reference implementation uses a Python loop to compute popcount32, which is correct but slow. For test correctness verification this is acceptable. However, consider using `torch.bincount` or numpy's built-in popcount for better performance if this test needs to run on larger tensors. The current implementation is fine for the test sizes used (hd=576, block_size=64).

ucm/store/nfsstore/device/rocm/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+if(NOT DEFINED CMAKE_HIP_ARCHITECTURES OR CMAKE_HIP_ARCHITECTURES STREQUAL "")
+    set(CMAKE_HIP_ARCHITECTURES "gfx90a")
```
ygwpz: 💡 Suggestion: Same as the trans/rocm CMakeLists.txt - the gfx90a default should be documented or warned about. Users building on consumer AMD GPUs will need to override this.

setup.py
jeffdaily: Done in 53f8b25 -- added `rocm` to the PLATFORM list in that warning.

ucm/shared/trans/cuda/cuda_sm_kernel.cu
jeffdaily: I kept the plain copy and documented why in 53f8b25 rather than reintroducing `volatile`, because on AMD it doesn't do what `st.volatile` does on NVIDIA here. Measured on gfx90a (ROCm 7.2): a `volatile` access compiles to flat addressing plus the `glc` bit on loads (an L1 bypass at GPU-L2 scope) and no cache bit on stores -- i.e. GPU-L2 coherence, not host/system visibility. Host visibility for these H2D/D2H copies instead comes from the transfer buffers being fine-grained-coherent host registrations (`hipHostRegister` default) plus the `hipStreamSynchronize` after each copy (GPU caches flush at kernel completion). So the plain copy is visible to the host; `volatile` would be neither necessary nor sufficient. System-scope ordering, if it were ever needed, would be `__threadfence_system`, not the qualifier.

ucm/store/nfsstore/device/cuda/cuda_device.cu
jeffdaily: I kept the plain copy and documented why in 53f8b25 rather than reintroducing `volatile`, because on AMD it doesn't do what `st.volatile` does on NVIDIA here. Measured on gfx90a (ROCm 7.2): a `volatile` access compiles to flat addressing plus the `glc` bit on loads (an L1 bypass at GPU-L2 scope) and no cache bit on stores -- i.e. GPU-L2 coherence, not host/system visibility. Host visibility for these H2D/D2H copies instead comes from the transfer buffers being fine-grained-coherent host registrations (`hipHostRegister` default) plus the `hipStreamSynchronize` after each copy (GPU caches flush at kernel completion). So the plain copy is visible to the host; `volatile` would be neither necessary nor sufficient. System-scope ordering, if it were ever needed, would be `__threadfence_system`, not the qualifier.

ucm/store/nfsstore/device/cuda/cuda_device.cu
jeffdaily: I kept the plain copy and documented why in 53f8b25 rather than reintroducing `volatile`, because on AMD it doesn't do what `st.volatile` does on NVIDIA here. Measured on gfx90a (ROCm 7.2): a `volatile` access compiles to flat addressing plus the `glc` bit on loads (an L1 bypass at GPU-L2 scope) and no cache bit on stores -- i.e. GPU-L2 coherence, not host/system visibility. Host visibility for these H2D/D2H copies instead comes from the transfer buffers being fine-grained-coherent host registrations (`hipHostRegister` default) plus the `hipStreamSynchronize` after each copy (GPU caches flush at kernel completion). So the plain copy is visible to the host; `volatile` would be neither necessary nor sufficient. System-scope ordering, if it were ever needed, would be `__threadfence_system`, not the qualifier.

ucm/shared/trans/rocm/CMakeLists.txt
jeffdaily: Fixed in b697824. These arms pinned gfx90a before `enable_language(HIP)`, which preempted CMake's host-GPU auto-detection (`rocm_agent_enumerator`) -- a non-gfx90a AMD user who didn't pass `-DCMAKE_HIP_ARCHITECTURES` would have silently built gfx90a binaries that fail to load on their card. Now `enable_language(HIP)` auto-detects the host arch (deduped), with gfx90a used only as a fallback when nothing is detected (e.g. a CPU-only build host); `-DCMAKE_HIP_ARCHITECTURES` still overrides.

ucm/store/nfsstore/device/rocm/CMakeLists.txt
jeffdaily: Fixed in b697824. These arms pinned gfx90a before `enable_language(HIP)`, which preempted CMake's host-GPU auto-detection (`rocm_agent_enumerator`) -- a non-gfx90a AMD user who didn't pass `-DCMAKE_HIP_ARCHITECTURES` would have silently built gfx90a binaries that fail to load on their card. Now `enable_language(HIP)` auto-detects the host arch (deduped), with gfx90a used only as a fallback when nothing is detected (e.g. a CPU-only build host); `-DCMAKE_HIP_ARCHITECTURES` still overrides.

ucm/shared/trans/cuda/cuda_sm_kernel.cu
jeffdaily: The 16-byte alignment requirement is unchanged from the original CUDA path -- its `ld.global.cs.v4.b32` vectorized access requires the same 16-byte alignment, which the KV blocks satisfy. So the HIP path introduces no new constraint; keeping it as-is for parity with the CUDA code.

ucm/sparse/gsa_on_device/csrc/cuda/ham_dist/operator.h
jeffdaily: The rocm CMake arm puts the hip_compat shim dir on the include path (that's where `cuda_runtime.h` resolves for ROCm builds), so a misconfiguration fails loudly at compile time rather than silently. I'd rather not add an `#error` probe keyed on a shim-specific define, since it couples the source to the build layout. Leaving as-is.

ucm/shared/vendor/hip_compat/cuda_runtime.h
jeffdaily: The `#define` approach is intentional: it matches the cuda_to_hip compatibility-header convention used across these ports, and it also has to cover the type aliases (`cudaStream_t`, `cudaError_t`, ...) that inline wrappers can't. The aliased surface here is small and unambiguous (Malloc/Free/Memcpy/Stream/Event), so the macros don't create overload-resolution hazards in practice. Leaving as-is.

ucm/sparse/gsa_on_device/csrc/rocm/ham_dist/CMakeLists.txt
jeffdaily: The two statements already cover both positions -- `string(REPLACE "-s " ...)` handles `-s` anywhere with a trailing space, and `string(REGEX REPLACE "(^| )-s$" ...)` handles it at end-of-string -- and both leave `-shared`/`-static` untouched. So it's robust for the flag strings in play; leaving as-is.

ucm/sparse/test/gsa/test_hamming_rocm_ref.py
jeffdaily: This is the reference oracle for the test, so I've kept it as the simplest obviously-correct implementation rather than optimizing it; as you note, it's fine at the test sizes. Leaving as-is.

ucm/shared/trans/rocm/CMakeLists.txt
jeffdaily: Refined in 57b2443: I dropped the explicit gfx90a fallback. `enable_language(HIP)` already errors clearly when no GPU is detected and no arch is set, so a no-GPU build host now sets `-DCMAKE_HIP_ARCHITECTURES` explicitly rather than getting a silent gfx90a default.

ucm/store/nfsstore/device/rocm/CMakeLists.txt
jeffdaily: Refined in 57b2443: I dropped the explicit gfx90a fallback. `enable_language(HIP)` already errors clearly when no GPU is detected and no arch is set, so a no-GPU build host now sets `-DCMAKE_HIP_ARCHITECTURES` explicitly rather than getting a silent gfx90a default.

setup.py
jeffdaily: Done. The PLATFORM warning now lists rocm alongside cuda/ascend/ascend-a3/musa/maca, and the build match block routes PLATFORM=rocm to -DRUNTIME_ENVIRONMENT=rocm.

ucm/shared/trans/cuda/cuda_sm_kernel.cu
jeffdaily: Kept the plain store; on AMD, volatile is not the mechanism for host visibility. I checked the gfx90a codegen: the copy lowers to flat_store_dwordx4 / global_store_dwordx4 (128-bit vector stores), so the stores are emitted, not dead-store-eliminated. dst is a kernel output pointer that escapes the function, so the compiler cannot drop or sink the writes regardless of the qualifier. On AMD, volatile only sets the glc bit (coherent at agent/L2 scope); it does not add system-scope ordering, so it is neither necessary nor sufficient for host visibility. Visibility comes from the kernel-completion barrier plus the per-transfer hipStreamSynchronize, which the runtime uses to publish device writes to the fine-grained-coherent host-registered buffer. The CopyData round-trip tests confirm this on real gfx90a (the host reads back byte-identical data). Mid-kernel host visibility, if ever needed, would require __threadfence_system, not volatile. A comment in the file documents this.

ucm/shared/trans/cuda/cuda_sm_kernel.cu
jeffdaily: Done. Added a comment noting the uint4 access requires 16-byte alignment -- the same alignment the .v4.b32 vector PTX on the CUDA path already requires. The copy strides from block bases by the 32-byte transfer unit, so src and dst are aligned.

ucm/store/nfsstore/device/cuda/cuda_device.cu
jeffdaily: Here the dropped qualifier is on the src read of host-registered memory (the store in H2DUnit targets device memory). Each unit is read once with no reuse, so there is no hoist or reorder hazard; the H2D copy reads host data written before the launch, and the launch is ordered after those host writes. As on the store side, AMD volatile is an L1-bypass/glc hint at agent scope, not a cross-agent ordering primitive. Added the 16-byte alignment note here as well.

ucm/store/nfsstore/device/cuda/cuda_device.cu
jeffdaily: Same as the trans D2H store: kept the plain store. The gfx90a codegen emits the vector store (not eliminated), and host visibility comes from the kernel-completion barrier plus the per-transfer hipStreamSynchronize, not from volatile (which on AMD is only a glc/agent-scope hint). The in-file comment notes that the stream sync replaces the volatile intent. Exercised by the store round-trip tests on real hardware.

ucm/sparse/gsa_on_device/csrc/cuda/ham_dist/operator.h
jeffdaily: Left as-is. The #include <cuda_runtime.h> is shared by both the CUDA and ROCm paths (both need it), so splitting it under an #ifdef would just duplicate it. On a ROCm build the CMake for this subtree puts the hip_compat dir on the include path; if it were ever missing the build fails loudly with a clear "cuda_runtime.h: No such file" on a ROCm host that has no CUDA toolkit, so an explicit #error would be redundant.

ucm/shared/vendor/hip_compat/cuda_runtime.h
jeffdaily: Kept the macros. This shim is intentionally a thin cuda->hip alias layer over the small fixed runtime surface the backend uses, and it also aliases the type names (cudaStream_t, cudaEvent_t, ...) which inline wrappers cannot cover. The set is closed, so overload-resolution concerns do not arise here.

ucm/shared/trans/rocm/CMakeLists.txt
jeffdaily: Done. The arch is no longer pinned -- enable_language(HIP) honors -DCMAKE_HIP_ARCHITECTURES and otherwise auto-detects the host GPU. Added a comment telling consumer Radeon users building on a different machine than they deploy to to set the arch explicitly, plus a STATUS message that echoes the selected arch during configure.

ucm/sparse/gsa_on_device/csrc/rocm/ham_dist/CMakeLists.txt
jeffdaily: Done. Collapsed the two-step strip into one word-boundary regex, (^| )-s( |$) replaced with a space, so -s is removed only as a whole token (never -std= or -shared) in any position and adjacent flags stay separated.

ucm/sparse/test/gsa/test_hamming_rocm_ref.py
jeffdaily: Left as-is. This is the reference oracle for the test, so the plain per-bit loop is deliberately simple to keep it obviously correct; at the test sizes (hd=576, block_size=64) its runtime is not a concern.

ucm/store/nfsstore/device/rocm/CMakeLists.txt
jeffdaily: Done, same as trans/rocm/CMakeLists.txt: no arch pin (enable_language(HIP) auto-detects and honors -DCMAKE_HIP_ARCHITECTURES), plus a consumer-GPU note and a STATUS message echoing the selected arch.

PR #1022, opened 2026-06-12
[Bugfix] Enforce strict timeout for Posix AIO

dante159753:
## Summary

Enforce hard timeouts for Posix AIO cache operations so a stuck mount no longer leaves vLLM requests waiting forever on external KV cache loads.

## Changes

- Add task-level terminal timeout handling for Posix AIO tasks.
- Make AIO submit/completion, task wait/check, and timeout cleanup tolerate stuck kernel IO.
- Add timeout metrics and vLLM integration handling so load failures can be rescheduled through `kv_load_failure_policy=recompute`.
- Keep AIO timeout configuration simple by using the existing `timeout_ms` default instead of separate AIO timeout knobs.
- Add Posix store tests for open timeout, IO timeout, repeated timeout, and saturated stuck-task behavior.

## Why

When a mounted cache backend stops responding, existing AIO requests can be submitted successfully but never complete. Before this change, requests that hit external KV cache could remain blocked until the mount recovered, making inference requests hang under storage incidents.

## Impact

With the default `timeout_ms`, stuck AIO load tasks are force-completed as timeout failures. In vLLM, deployments using `kv_load_failure_policy=recompute` can continue serving by recomputing the affected KV blocks instead of waiting indefinitely. If the storage backend remains stuck, cache loads may be slower because they time out and fall back to recompute, but the service remains available.

## Verification

- A2 online vLLM+UCM hang simulation:
  - Setup: loop-backed ext4 mount through device mapper, UCM `storage_backends` on that mount, `posix_io_engine: aio`, `io_direct: true`, no explicit `timeout_ms`.
  - Current branch: `dmsetup suspend` for ~75s during external cache load; `/v1/models` stayed responsive; 8 concurrent completion requests returned `200 OK` in ~60.8s; post-resume requests completed in ~0.64s.
  - Baseline `origin/develop`: same setup; 8 concurrent completion requests all hit client timeout at ~55s while UCM repeatedly logged `Task(1) has not finished after (2000) ms`; post-resume requests completed after the mount recovered.


ucm/store/posix/cc/aio_impl.cc
```diff
@@ -181,23 +285,96 @@ void AioImpl::HarvestCompletions(std::vector<io_event>& events)
                 res.nBytes = -1;
                 res.error = -static_cast<int>(events[i].res);
             }
-            (*cb)(res);
-            delete cb;
+            if (cb) {
```
ygwpz, line 289: ⚠️ Warning: The null check for `cb` here could mask a logic error or memory corruption issue. For valid IOs submitted via ReadAsync/WriteAsync, the callback should always exist since it's stored in `iocb->aio_data` before submission. If `cb` is null, this indicates either:
1. An IO was submitted without a callback (which shouldn't happen)
2. Memory corruption of the aio_data field

Consider adding an assertion or error log when `cb` is null for non-null `iocbPtr`, as silently skipping the callback may leave the caller waiting indefinitely.

ucm/store/posix/cc/aio_impl.cc
```diff
@@ -181,23 +285,96 @@ void AioImpl::HarvestCompletions(std::vector<io_event>& events)
...
+        std::lock_guard<std::mutex> lk(tableMutex_);
+        auto it = iocbTable_.find(tag);
+        if (it == iocbTable_.end()) {
+            UC_DEBUG("AIO cancel task({}): no tracked in-flight IO.", tag);
+            return;
+        }
+        auto& vec = it->second;
+        for (auto vi = vec.begin(); vi != vec.end();) {
+            io_event ev{};
+            auto r = AioCancel(ctx_, *vi, &ev);
```
ygwpz, line 338: 🔴 Critical: When `AioCancel` returns 0 (success), the canceled operation's result is available in `ev`. However, the code unconditionally invokes the callback with `ECANCELED` error without checking `ev.res`. According to Linux AIO documentation, when `io_cancel` succeeds, `ev.res` contains the actual result of the canceled operation - it could be:
- A negative error code (including `ECANCELED` = 125)
- A positive value indicating partial completion

The current implementation ignores this information and always reports `ECANCELED`. This could misrepresent the actual completion status to the caller. Consider checking `ev.res` and using it if available:

```cpp
if (r == 0) {
    iocbToTag_.erase(*vi);
    cancelled.push_back(*vi);
    vi = vec.erase(vi);
    // Store ev for later use in determining actual result
}
...
for (auto* cbp : cancelled) {
    auto* cb = reinterpret_cast<Callback*>(cbp->aio_data);
    if (cb) {
        // Use actual result from ev if tracked, or ECANCELED as fallback
        (*cb)(Result{-1, ECANCELED});
        delete cb;
    }
    delete cbp;
}
```

ucm/store/posix/cc/io_engine_aio.h
```diff
@@ -109,11 +190,21 @@ class IoEngineAio : public Detail::TaskWrapper<TransTask, Detail::TaskHandle> {
...
+                IncrementAioTimeoutMetric();
+                MarkBackendUnhealthy("aio submit timeout");
+            } else {
+                IncrementIoErrorMetric();
+            }
+            handleFailure(result.fd);
+        }
     }
     template <bool dump>
     void Dispatch(TaskPtr t, WaiterPtr w)
```
ygwpz, line 166: 💡 Suggestion: Changed from stack-allocated `struct iocb cb` to heap-allocated `std::make_unique<struct iocb>()`. This adds allocation overhead for every IO operation.

For high-throughput scenarios (thousands of IOs per second), this could impact performance. Consider:
1. Using a pre-allocated pool of iocb structures
2. Reusing iocb objects across operations
3. Measuring the performance impact in benchmarks

The change was necessary for tracking, but worth monitoring for performance regression.

ucm/store/posix/cc/io_engine_aio.h
```diff
@@ -170,6 +277,58 @@ class IoEngineAio : public Detail::TaskWrapper<TransTask, Detail::TaskHandle> {
...
+    void ForceComplete(Detail::TaskHandle id, const WaiterPtr& w)
+    {
+        auto pending = w ? w->Pending() : 0;
+        UC_WARN("AIO task({}) force-completing; pending latch count before abort={}.", id, pending);
+        failureSet_.Insert(id);
+        timeoutSet_.Insert(id);
+        abandonedSet_.Insert(id);
+        aio_.CancelTask(id);
+        blockOperator_.CancelQueued(id);
+        if (w) { w->Abort(); }
```
ygwpz, line 282: 🔴 Critical: Potential race condition in `ForceComplete`. The sequence is:
1. Insert into `failureSet_`, `timeoutSet_`, `abandonedSet_`
2. Call `aio_.CancelTask(id)` which may invoke callbacks with ECANCELED
3. Call `blockOperator_.CancelQueued(id)` which may invoke callbacks
4. Call `w->Abort()` which does `counter_.exchange(0)`

However, callbacks from steps 2-3 may call `w->Done()` which decrements `counter_`. The race:
- If a callback calls `Done()` before `Abort()`, it decrements counter (e.g., counter becomes 5)
- Then `Abort()` does `exchange(0)`, setting counter to 0
- The callback's decrement is lost, but the counter is already 0

This is actually safe because `Abort()` will set counter to 0 anyway. But there's another issue:
- If `Abort()` runs first, setting counter to 0 and invoking `finish_()`
- Then a callback runs and tries to call `Done()` which sees counter=0 and returns early
- But the callback might still try to do cleanup work (like `CommitBlock`)

The `abandonedSet_` check in callbacks should catch this, but verify that ALL callback paths check `IsAbandoned()` before doing any work.

ucm/shared/infra/thread/latch.h
```diff
@@ -60,6 +60,20 @@ class Latch {
         if (this->counter_ == 0) { return; }
         this->cv_.wait(lk, [this] { return this->counter_ == 0; });
     }
+    // Abort may race with Done(); only the thread moving counter_ to zero runs finish_.
+    void Abort() noexcept
```
ygwpz, line 64: ⚠️ Warning: Race condition between `Abort()` and `Done()`.

In `Abort()`:
```cpp
auto prev = this->counter_.exchange(0, std::memory_order_acq_rel);
```

In `Done()`:
```cpp
auto counter = this->counter_.load(std::memory_order_acquire);
while (counter > 0) {
    auto desired = counter - 1;
    if (this->counter_.compare_exchange_weak(counter, desired, ...)) {
```

Race scenario:
1. `counter_` = 5, thread A calls `Done()`, reads counter=5
2. Thread B calls `Abort()`, exchanges counter to 0, prev=5, runs `finish_()`
3. Thread A's CAS fails because counter is now 0, retries loop
4. Thread A sees counter=0, exits loop without decrementing

This is correct behavior - Abort wins. However, consider documenting this explicitly:
- `Abort()` atomically zeros the counter and runs `finish_` exactly once
- `Done()` is idempotent when counter is already zero
- `finish_` may run from either `Abort()` or the last `Done()`, but never both

The comment "Abort may race with Done()" is good, but could be more specific about the guarantee: "Only one of Abort or the last Done will execute finish_".

ucm/store/posix/cc/backend_health.h
```diff
@@ -0,0 +1,447 @@
...
+            }
+            if (WaitProbe(state)) {
+                if (worker.joinable()) { worker.join(); }
+                if (state->status.Success()) {
+                    MarkRecovered();
+                    return;
+                }
+            } else {
+                if (!ShouldStop()) { IncrementProbeTimeoutMetric(); }
+                if (worker.joinable()) { worker.detach(); }
```
ygwpz: ⚠️ Warning: When `WaitProbe` returns false (probe timed out), the worker thread is detached. However, the worker thread will eventually complete and decrement `control->active` (line 209). This creates a timing issue:

1. Probe times out, worker is detached
2. `control->active` is still elevated (the worker hasn't decremented yet)
3. Next loop iteration checks `active >= kMaxProbeWorkers` and may skip starting a new probe
4. Eventually the detached worker finishes and decrements `active`

This could delay the next probe attempt by up to `kProbeIntervalMs`. Consider:
- Using a separate timeout counter that doesn't count detached workers
- Adding a "pending decrement" mechanism
- Or documenting that detached workers may temporarily block new probe starts

This is acceptable for recovery scenarios where timing isn't critical, but worth noting.

ucm/store/detail/template/task_wrapper.h
```diff
@@ -94,13 +95,32 @@ class TaskWrapper {
...
+            if (!drained) {
+                UC_ERROR(
+                    "Task({}) latch never drained; returning Timeout and abandoning in-flight"
+                    " IO.",
+                    taskId);
+            }
+            // Late IO callbacks must still see failure and avoid committing success.
             return Status::Timeout();
         }
         auto failure = failureSet_.Contains(taskId);
```
ygwpz, line 105: 💡 Suggestion: Inconsistency in timeout handling. 

The `Wait()` function has two timeout paths:
1. Direct timeout: when `w->WaitFor(timeoutMs_)` returns false (lines 94-115)
2. Indirect timeout: when `timeoutSet_.Contains(taskId)` is true (lines 117-123)

For direct timeout (line 94-115):
- `failureSet_.Insert(taskId)` is called (line 95)
- But `timeoutSet_.Insert(taskId)` is NOT called

For indirect timeout via `Cancel()` in io_engine_aio.h:
- Both `failureSet_.Insert(id)` and `timeoutSet_.Insert(id)` are called

This inconsistency could cause issues. Consider:
- Adding `timeoutSet_.Insert(taskId)` in the direct timeout path for consistency
- Or removing `timeoutSet_` usage and relying solely on the return status

The current behavior works, but the dual tracking mechanism adds complexity without clear benefit.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -421,6 +425,22 @@ def __init__(
         self.hash_block_size = self.block_size
         self.block_size *= self.cp_world_size
 
+    @staticmethod
+    def _record_counter(name: str, value: float = 1.0) -> None:
+        _record_counter(name, value)
+
+    def _record_load_error(self, metric_name: str, block_ids: Any) -> None:
+        invalid_blocks = set(block_ids)
+        new_invalid_blocks = invalid_blocks - self._invalid_block_ids
```
ygwpz, line 434: 💡 **Suggestion:** The `_invalid_block_ids` set is accessed without synchronization. If `_record_load_error` can be called concurrently from multiple threads, there's a potential race condition where:
1. Thread A computes `new_invalid_blocks = {1,2} - current_set`
2. Thread B computes `new_invalid_blocks = {1,3} - current_set`  
3. Both threads update the set, leading to double-counting in `connector_load_invalid_blocks_total`.

If vLLM callbacks are single-threaded, this is safe. Otherwise, consider using a lock or making the set update atomic.

ucm/store/detail/template/task_wrapper.h
```diff
@@ -94,13 +95,32 @@ class TaskWrapper {
             failureSet_.Insert(taskId);
             Cancel(t);
             constexpr size_t drainSliceMs = 2000;
-            while (!w->WaitForDuration(drainSliceMs)) {
-                UC_WARN("Task({}) has not finished after ({}) ms.", taskId, drainSliceMs);
+            constexpr size_t maxDrainRounds = 3;
+            bool drained = false;
+            for (size_t r = 0; r < maxDrainRounds; ++r) {
```
qyh111: 通用 TaskWrapper 被改成 6 秒后放弃 IO
[task_wrapper.h#L97-L115](https://github.com/ModelEngine-Group/unified-cache-management/blob/e3dc7e08f28c955576f25244c69e8cc9750ccad5/ucm/store/detail/template/task_wrapper.h#L97-L115)
这会同时影响 psync、cache、compress、ds3fs，并非仅 AIO。建议恢复原来的持续 drain；AIO 的 Cancel() 已调用 Abort()，本身就能立即退出 drain。

qyh111 (dismissed):

ucm/store/posix/cc/aio_impl.cc
```diff
@@ -84,10 +158,23 @@ AioImpl::~AioImpl()
...
+                iocbToTag_.size());
+    }
 }
 
-Status AioImpl::Setup()
+Status AioImpl::Setup(size_t timeoutMs)
 {
+    constexpr size_t defaultSweepIntervalMs = 100;
+    epollTimeoutMs_ = timeoutMs > 0 ? timeoutMs : defaultSweepIntervalMs;
+    sweepIntervalMs_ = defaultSweepIntervalMs;
```
qyh111, line 172: `sweepIntervalMs_` 是 100ms，但 `epollTimeoutMs_ = timeoutMs`。没有 completion event 时，deadline sweep 要等 `epoll_wait` 返回才执行；默认 timeout 下 `Check()` 路径可能比预期晚几十秒才完成，不符合严格超时语义。建议将 epoll timeout 限制为 sweep interval。

ucm/store/detail/template/task_wrapper.h
```diff
@@ -94,15 +96,29 @@ class TaskWrapper {
...
             }
-            failureSet_.Remove(taskId);
+            if (!drained) {
+                UC_ERROR(
+                    "Task({}) latch never drained; returning Timeout and abandoning in-flight"
+                    " IO.",
+                    taskId);
+            }
+            // For Aio, Late IO callbacks must still see failure and avoid committing success.
+            if (!IsAio()) { failureSet_.Remove(taskId); }
```
qyh111, line 102: 对 AIO，timeout/error 后都不再 remove task id。task id 单调递增时短期不会误伤当前任务，但存储故障反复发生会让集合持续增长。建议给 AIO 的 late-callback 防护单独建可清理状态，或在确认不会再有 late callback 后清理。

ucm/store/detail/template/task_wrapper.h
dante159753: Handled in `28417ec`: `timeoutSet_` has been removed, so we no longer maintain both `failureSet_` and `timeoutSet_`. A direct `Wait()` timeout still returns `Status::Timeout()`. If the AIO watchdog/sweep force-completes the task first, a later `Wait()` now observes `failureSet_` and returns `Status::Error()`. This keeps the state model simpler and avoids the previous inconsistency between two sets. Targeted tests passed on A2.

ucm/store/detail/template/task_wrapper.h
dante159753: Good point. In the latest `28417ec`, I reverted the more complex TaskState/refcount approach and kept the change smaller with an `IsAio()` branch. AIO keeps `failureSet_` after timeout/error so late callbacks can still see the failed task. Non-AIO paths remove the task id before returning, so this AIO-specific behavior does not leak into psync/cache/compress/ds3fs.

ucm/store/detail/template/task_wrapper.h
dante159753: 这里需要让failureSet_包含失败的taskId避免后续commit时候将超时写入的文件提交，当前的代码架构很难简单实现异步清理failureSet_，为了清理这个failureSet_添加一大堆逻辑感觉没有必要，毕竟只有极少数异常情况会残留taskid，空间和性能影响也微乎其微

ucm/store/posix/cc/aio_impl.cc
dante159753: Agreed that `cb == nullptr` should not happen for IOs submitted through `ReadAsync` / `WriteAsync`; `aio_data` is populated before submit. I would treat this as a defensive diagnostic rather than normal control flow. A small follow-up change could add an error log on the null-callback path while still reclaiming the iocb. I have not changed it in this push because it is orthogonal to the timeout behavior.

ucm/store/posix/cc/aio_impl.cc
dante159753: For this PR, `CancelTask()` is used as best-effort cleanup after the owner task has already been marked failed/timed out. The store side relies on the callback to unblock/reclaim resources; it does not use a successful cancel result to make the task successful, and `failureSet_` prevents a timed-out dump from committing. Reporting `ECANCELED` here is therefore intentional for the forced-timeout path. If we want `AioImpl` to preserve exact `io_cancel` results as a generic API cleanup, that can be a separate small change.

ucm/store/posix/cc/io_engine_aio.h
dante159753: Yes, this adds one allocation per IO. The heap allocation is deliberate here because the iocb address must remain stable after submit and until either completion or best-effort cancel. A pool would be a reasonable optimization if benchmarks show this allocation on the hot path, but I would keep it out of this timeout correctness PR to avoid adding another lifetime-management layer.

ucm/store/posix/cc/io_engine_aio.h
dante159753: I rechecked the callback paths after the simplification. `ForceComplete()` inserts `failureSet_` before calling `CancelTask()` and `CancelQueued()`. Late open callbacks check `failureSet_` before submitting AIO and commit false on dump. Late IO callbacks also commit dump blocks based on `!failureSet_.Contains(tid)`, so a timed-out task cannot commit success. `timeoutSet_` has been removed in `28417ec`, and the latch `Abort()` / `Done()` race is expected to be safe because `Abort()` wins by zeroing the counter. The late-open multi-shard test covers the commit-false behavior.

ucm/store/posix/cc/backend_health.h
dante159753: This thread is outdated for the current branch. The backend health/probe logic was reverted from the current diff, so there is no `WaitProbe` detached-worker path left to address in this PR.

ucm/integration/vllm/ucm_connector.py
dante159753: Current uses of `_record_load_error()` are from the connector worker flow, not from multiple concurrent Python threads. The set is only used to deduplicate invalid-block metrics and is reset by `get_block_ids_with_load_errors()`. I am not adding a lock in this PR. If vLLM starts invoking these connector callbacks concurrently, we should protect `_invalid_block_ids` with a small lock at that point.

ucm/store/posix/cc/aio_impl.cc
dante159753: Addressed in `fd9be59`. `epollTimeoutMs_` is now an internal fixed 10ms wakeup interval, not configurable and not derived from `timeout_ms`. `sweepIntervalMs_` stays at 100ms, while `submitTimeoutMs_` still uses `timeout_ms`. This keeps the completion thread waking frequently enough for deadline sweep even when there are no completion events. Targeted AIO tests passed on A2.

ucm/store/detail/template/task_wrapper.h
qyh111: 这里 `IsAio()` 只隔离了 failureSet_ 的清理策略，没有隔离 Wait() 超时后最多 drain 6 秒就返回的行为。非 AIO backend 仍然会执行同一段 maxDrainRounds 逻辑；如果底层 worker 已经开始执行且卡住，Cancel() 不一定能让 waiter Done，Wait() 仍会在 6 秒后返回 Timeout。建议把有限 drain/abandon 行为限定到 AIO，非 AIO 保持原来的持续 drain 语义。

qyh111 (approved):

ygwpz (approved):

PR #1023, opened 2026-06-12
[Fix][Opt]: Optimize status and memory release, refactor architecturefor asu_transport_impl.

yuanzhg078:
## Purpose
Optimize ASU transport task status handling and sub-batch resource cleanup, while simplifying AsuTransportImpl by splitting submission, SQE construction, completion, and config validation logic into clearer modules.

## Modifications
- AsuTransportImpl: refactored ProcessTask flow to submit, assign connections, build send buffers, send, then finalize from sub-batch states; added cancellation/shutdown cleanup for sub-batch resources; tightened PollTaskCompletions status handling and CQE result finalization.
- Submit flow: centralized sub-batch send failure marking, released resources for pre-send/send-failed sub-batches, handled missing send count attrs safely, and normalized partial failure status messages.
- SQE request: simplified sub-batch request preparation and removed duplicated validation helpers; entry/key/keep-alive request paths now initialize entryStatus directly and use ToKvOpcode.
- Config parser: moved SQE attr validation into config loading, including integer/bool validation and required positive kernel_count / quiet_count.

- Task completion/manager: simplified final status calculation, made empty sub-batch completion produce partial failure, removed TransportTaskManager::CancelAll.

## Test
Updated existing ASU transport unit tests for submit flow, SQE request building, IO scheduling, and task completion/resource cleanup.

### DIFF
--- a/ucm/transport/kv/asu/trans/src/asu_submit_flow.cpp
+++ b/ucm/transport/kv/asu/trans/src/asu_submit_flow.cpp
@@ -35,7 +35,9 @@ namespace {
 std::uint32_t GetSendCountAttr(const std::unordered_map<std::string, std::string>& attrs,
                                const std::string& name)
 {
-    return static_cast<std::uint32_t>(std::stoull(attrs.at(name), nullptr, 0));
+    auto it = attrs.find(name);
+    if (it == attrs.end()) { return 0; }
+    return static_cast<std::uint32_t>(std::stoull(it->second, nullptr, 0));
 }
 
 void SetSubBatchSendFailed(TransportSubBatchContext& subBatchContext, const Status& status)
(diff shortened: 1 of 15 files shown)


Infinite666:
Commit name typo Pptimize => Optimize.

ucm/transport/kv/asu/trans/src/asu_submit_flow.cpp
```diff
@@ -35,7 +35,9 @@ namespace {
 std::uint32_t GetSendCountAttr(const std::unordered_map<std::string, std::string>& attrs,
                                const std::string& name)
 {
-    return static_cast<std::uint32_t>(std::stoull(attrs.at(name), nullptr, 0));
+    auto it = attrs.find(name);
```
ygwpz, line 38: 💡 **Suggestion:** Returning 0 when an attribute is missing could lead to silent failures if the attribute is required. Consider:

1. If `send_count` is a **required** attribute, throwing an exception (original behavior) ensures the caller knows something is wrong.
2. If it's **optional**, returning 0 is fine, but consider adding a log message: `UC_DEBUG("Attribute '{}' not found, using default 0", name);`

This helps debugging if someone forgets to set a required attribute.

Infinite666 (approved):

PR #1024, opened 2026-06-12
[Feat] use host-pinned memory with dual CPU/device addresses for transport buffers

yumingyue624:
## Purpose
Switch ASU send/flag buffers from plain host memory to host-pinned
memory so that CPU code packs SQEs through the local mapping while
HCOMM/RDMA uses the device-visible mapping of the same allocation.

## Modifications
1. BufferManager: allocate host-pinned memory via aclrtMallocHost +
aclrtHostRegisterV2, obtain device pointer via
aclrtHostGetDevicePointer. ScatterGatherEntry gains device_addr.
RegisterMemory uses device address for host-pinned regions.
2. AsuTransportImpl: send/flag buffers use HOST_PINNED instead of HOST.
3. asu_submit_flow: pass device_addr to SendIoBatch.
4. sqe_request: use flagBuffer.device_addr for response_buffer_addr.
5. Tests: added host-pinned dual-address and device_addr assertions.

## Test
- buffer_manager_test: HostPinnedRegistersDeviceAddress.
- asu_submit_flow_test: BuildSubBatchSendBuffersUsesHostPinnedDeviceAddresses.
- sqe_request_test: packed response address matches device_addr.

### DIFF
--- a/ucm/shared/trans/ascend/ascend_buffer.cc
+++ b/ucm/shared/trans/ascend/ascend_buffer.cc
@@ -23,10 +23,35 @@
  * */
 #include "ascend_buffer.h"
 #include <acl/acl.h>
+#include <limits>
 #include <sys/mman.h>
 #include "logger/logger.h"
 
 namespace UC::Trans {
+
+namespace {
+
+constexpr std::uintptr_t HOST_REGISTER_PAGE_SIZE = 4096;
+
+void FreeHostMemory(void* host)
+{
+    auto ret = aclrtFreeHost(host);
+    if (ret != ACL_SUCCESS) { UC_ERROR("Failed to free host memory addr={} ret={}", host, ret); }
+}
+
+void* AlignUp(void* ptr, std::uintptr_t alignment)
+{
+    const auto addr = reinterpret_cast<std::uintptr_t>(ptr);
+    return reinterpret_cast<void*>((addr + alignment - 1) / alignment * alignment);
+}
+
+void ReleaseHostPinnedMemory(void* registeredHost, void* allocatedHost)
+{
+    Buffer::UnregisterHostBuffer(registeredHost);
+    FreeHostMemory(allocatedHost);
+}
+
+}  // namespace
 
 class HostHugePages : public std::enable_shared_from_this<HostHugePages> {
     struct ConstructorKey {};
@@ -122,6 +147,35 @@ std::shared_ptr<void> Trans::AscendBuffer::MakeHostBuffer(size_t size)
     auto ret = aclrtMallocHost(&host, size);
     if (ret == ACL_SUCCESS) { return std::shared_ptr<void>(host, aclrtFreeHost); }
     return nullptr;
+}
+
+std::shared_ptr<void> Trans::AscendBuffer::MakeHostPinnedBuffer(size_t size, void** pDevice)
+{
+    if (pDevice) { *pDevice = nullptr; }
+
+    constexpr auto kMaxSize = std::numeric_limits<size_t>::max();
+    if (size > kMaxSize - (HOST_REGISTER_PAGE_SIZE - 1)) { return nullptr; }
+
+    void* allocatedHost = nullptr;
+    const auto allocationSize = size + HOST_REGISTER_PAGE_SIZE - 1;
+    auto ret = aclrtMallocHost(&allocatedHost, allocationSize);
+    if (ret != ACL_SUCCESS) { return nullptr; }
+
+    void* host = AlignUp(allocatedHost, HOST_REGISTER_PAGE_SIZE);
+
+    void* device = nullptr;
+    auto status = Buffer::RegisterHostBuffer(host, size, &device);
+    if (status.Failure()) {
+        UC_ERROR("Failed to register host-pinned memory addr={} size={} status={}", host, size,
+                 status);
+        FreeHostMemory(allocatedHost);
+        return nullptr;
+    }
+
+    if (pDevice) { *pDevice = device; }
+    return std::shared_ptr<void>(host, [allocatedHost](void* registeredHost) {
+        ReleaseHostPinnedMemory(registeredHost, allocatedHost);
+    });
 }
 
 std::shared_ptr<void> Trans::AscendBuffer::MakeHostBuffer4DirectIo(size_t size)
--- a/ucm/shared/trans/buffer.h
+++ b/ucm/shared/trans/buffer.h
@@ -39,6 +39,7 @@ public:
 
     virtual std::shared_ptr<void> MakeHostBuffer(size_t size) = 0;
     virtual std::shared_ptr<void> MakeHostBuffer4DirectIo(size_t size) = 0;
+    virtual std::shared_ptr<void> MakeHostPinnedBuffer(size_t size, void** pDevice = nullptr) = 0;
     virtual Status MakeHostBuffers(size_t size, size_t number) = 0;
     virtual std::shared_ptr<void> GetHostBuffer(size_t size) = 0;
 
--- a/ucm/transport/kv/asu/trans/include/asu_transport/asu_transport.h
+++ b/ucm/transport/kv/asu/trans/include/asu_transport/asu_transport.h
@@ -72,9 +72,12 @@ struct TransportConfig {
     bool preconnect{true};
     bool bindCqPoller{true};
 
+    // Slot sizes are caller-visible capacities; BufferManager computes the
+    // aligned physical stride used for allocation and memory registration.
     std::size_t sendBufferSlotSize{4160};
     std::size_t sendBufferSlotNum{128};
-    std::size_t flagBufferSlotSize{128};
+    // Maximum memory required by a batch store/retrieve response flag buffer.
+    std::size_t flagBufferSlotSize{71};
     std::size_t flagBufferSlotNum{4096};
     std::size_t asuBatchLoadIoNum{110};
     std::size_t asuBatchStoreIoNum{110};
--- a/ucm/transport/kv/asu/trans/src/asu_submit_flow.cpp
+++ b/ucm/transport/kv/asu/trans/src/asu_submit_flow.cpp
@@ -130,24 +130,31 @@ Status AsuTransportImpl::BuildSubBatchSendBuffers(
             continue;
         }
 
-        if (subBatchContext.flagBuffer.addr == 0 || subBatchContext.flagBuffer.length == 0) {
-            const auto subBatchStatus =
-                Status::Error(StatusCode::NOT_INITIALIZED, "sub-batch flag buffer is not ready");
+        if (subBatchContext.channel == nullptr ||
+            !IsTransportBufferReady(subBatchContext.sendSge) ||
+            !IsTransportBufferReady(subBatchContext.flagBuffer)) {
+            const auto subBatchStatus = Status::Error(StatusCode::NOT_INITIALIZED,
+                                                      "sub-batch transport buffers are not ready");
             UC_ERROR(
-                "Sub-batch flag buffer is not ready index={} cid={} flag_addr={} flag_length={}",
-                index, subBatchContext.cid, subBatchContext.flagBuffer.addr,
-                subBatchContext.flagBuffer.length);
+                "Sub-batch transport buffers are not ready index={} cid={} channel={} "
+                "send_local_addr={} send_device_addr={} send_length={} send_slot={} "
+                "flag_local_addr={} flag_device_addr={} flag_length={} flag_slot={}",
+                index, subBatchContext.cid, subBatchContext.channel != nullptr,
+                subBatchContext.sendSge.local_addr, subBatchContext.sendSge.device_addr,
+                subBatchContext.sendSge.length, subBatchContext.sendSge.slot_index,
+                subBatchContext.flagBuffer.local_addr, subBatchContext.flagBuffer.device_addr,
+                subBatchContext.flagBuffer.length, subBatchContext.flagBuffer.slot_index);
             SetSubBatchSendFailed(subBatchContext, subBatchStatus);
             if (status.ok()) { status = subBatchStatus; }
             ReleaseSubBatchResources(subBatchContext);
             continue;
         }
 
-        ioBatches.push_back(
-            TransProvider::SendIoBatch{subBatchContext.channel->GetConnection(),
-                                       reinterpret_cast<void*>(subBatchContext.sendSge.addr),
-                                       reinterpret_cast<void*>(subBatchContext.flagBuffer.addr),
-                                       subBatchContext.sendSge.length});
+        ioBatches.push_back(TransProvider::SendIoBatch{
+            subBatchContext.channel->GetConnection(),
+            reinterpret_cast<void*>(subBatchContext.sendSge.device_addr),
+            reinterpret_cast<void*>(subBatchContext.flagBuffer.device_addr),
+            subBatchContext.sendSge.length});
         subBatchIndexes.emplace_back(index);
     }
 
--- a/ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
+++ b/ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
@@ -22,9 +22,13 @@
  * SOFTWARE.
  * */
 #include "asu_transport_impl.h"
+#include <acl/acl.h>
 #include <algorithm>
+#include <array>
 #include <chrono>
+#include <cstdint>
 #include <memory>
+#include <string>
 #include <thread>
 #include <utility>
 #include "aicpu_trans_provider.h"
@@ -37,6 +41,26 @@
 #include "transport_config_parser.h"
 
 namespace UC::ASU {
+
+namespace {
+
+constexpr std::size_t kFlagBufferHeaderCopySize = kCqeDwordCount * sizeof(std::uint32_t);
+
+Status CopyDeviceToHost(const ScatterGatherEntry& sge, void* host, std::size_t size)
+{
+    if (size > sge.length) {
+        return Status::Error(StatusCode::INVALID_ARGUMENT, "copy size exceeds buffer length");
+    }
+    const auto ret = aclrtMemcpy(host, size, reinterpret_cast<void*>(sge.device_addr), size,
+                                 ACL_MEMCPY_DEVICE_TO_HOST);
+    if (ret != ACL_SUCCESS) {
+        return Status::Error(StatusCode::INTERNAL_ERROR,
+                             "copy device memory to host failed ret=" + std::to_string(ret));
+    }
+    return Status::OK();
+}
+
+}  // namespace
 
 AsuTransportImpl::~AsuTransportImpl() { Shutdown(); }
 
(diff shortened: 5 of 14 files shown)


ucm/transport/kv/asu/trans/src/asu_submit_flow.cpp
```diff
@@ -119,12 +119,15 @@ Status AsuTransportImpl::BuildSubBatchSendBuffers(
             continue;
         }
 
-        if (subBatchContext.flagBuffer.addr == 0 || subBatchContext.flagBuffer.length == 0) {
+        if (subBatchContext.sendSge.device_addr == 0 || subBatchContext.flagBuffer.addr == 0 ||
```
Infinite666: What about sendSge.addr and sendSge.length?

ucm/transport/kv/asu/trans/src/buffer_manager.cpp
```diff
@@ -28,6 +28,79 @@
...
+namespace {
+
+struct BufferRegion {
+    std::shared_ptr<void> owner;
+    void* localAddr{nullptr};
+    void* deviceAddr{nullptr};
+    TransProvider::MemType providerMemType{TransProvider::MemType::MEM_HOST};
+};
+
+class BufferRegionCreator : public Trans::AscendBuffer {
```
Infinite666: Naming problem, the parent class is a Buffer and the child class is a Creator?

ucm/transport/kv/asu/trans/src/buffer_manager.cpp
```diff
@@ -60,22 +134,18 @@ Status BufferManager::Init(std::string name, MemoryType type, std::size_t slot_s
...
-        default:
-            return Status::Error(StatusCode::INVALID_ARGUMENT, name_ + ": unsupported memory type");
-    }
-
-    if (!memory_) {
-        return Status::Error(StatusCode::INTERNAL_ERROR, name_ + ": failed to allocate memory");
-    }
+    BufferRegionCreator regionCreator;
+    BufferRegion region;
+    auto allocStatus = regionCreator.MakeRegion(memory_type_, total, region);
```
Infinite666: Violating the inherit spirit, the child class has its own public function and will be called directly.

ucm/transport/kv/asu/trans/src/buffer_manager.cpp
```diff
@@ -28,6 +28,79 @@
...
+            aclrtHostUnregister(hostAddr);
+            aclrtFreeHost(hostAddr);
+            return Status::Error(StatusCode::INTERNAL_ERROR,
+                                 "failed to get host-pinned device address");
+        }
+
+        // The owner keeps the ACL registration alive until after HCOMM has
+        // unregistered the region in BufferManager's destructor.
+        auto owner = std::shared_ptr<void>(hostAddr, [](void* addr) {
+            aclrtHostUnregister(addr);
```
ygwpz: [Warning] Custom deleter lacks error handling

The custom deleter calls aclrtHostUnregister(addr) and aclrtFreeHost(addr) without checking return values. If aclrtHostUnregister fails, calling aclrtFreeHost on still-registered memory may cause undefined behavior or resource leaks.

Consider:
1. Logging errors in the deleter (even if silent)
2. Ensuring the cleanup order is safe per ACL documentation
3. Using aclrtFreeHost only if unregister succeeded

ucm/transport/kv/asu/trans/src/buffer_manager.cpp
```diff
@@ -28,6 +28,79 @@
...
+
+class BufferRegionCreator : public Trans::AscendBuffer {
+public:
+    Status MakeRegion(MemoryType type, std::size_t size, BufferRegion& region)
+    {
+        switch (type) {
+            case MemoryType::HOST: {
+                auto owner = MakeHostBuffer(size);
+                if (!owner) { return AllocationFailed("host"); }
+                region = {owner, owner.get(), owner.get(), TransProvider::MemType::MEM_HOST};
```
ygwpz: [Suggestion] device_addr semantic inconsistency for HOST memory type

For MemoryType::HOST, deviceAddr is set equal to localAddr (owner.get()). This is semantically ambiguous because plain host memory does not have a device-visible mapping. The comment in buffer_manager.h states device_addr is for device-visible address used by HCOMM/RDMA, but setting it equal to addr for HOST type contradicts this.

Consider:
1. Setting device_addr = 0 for HOST type (forcing callers to handle the case)
2. Or explicitly documenting that addr == device_addr for HOST memory means the same physical address is used

ucm/transport/kv/asu/trans/src/asu_submit_flow.cpp
ygwpz: [Critical] Missing validation for sendSge.length

The validation now checks sendSge.device_addr, flagBuffer.addr, flagBuffer.device_addr, and flagBuffer.length, but sendSge.length is not validated. If sendSge.length is 0 or invalid, the SendIoBatch will have an incorrect length parameter, potentially causing buffer overflow or underflow in RDMA operations.

Add: sendSge.length == 0 to the validation condition.

ucm/transport/kv/asu/trans/src/buffer_manager.cpp
```diff
@@ -141,8 +210,10 @@ Status BufferManager::Allocate(std::size_t size, ScatterGatherEntry& sge)
     if (idx == IndexPool::npos) {
         return Status::Error(StatusCode::RESOURCE_BUSY, name_ + ": no free slots");
     }
-    void* addr = static_cast<char*>(memory_.get()) + idx * slot_size_;
-    sge.addr = reinterpret_cast<std::uint64_t>(addr);
+    const auto offset = idx * slot_size_;
+    sge.addr = reinterpret_cast<std::uint64_t>(static_cast<char*>(memory_.get()) + offset);
+    sge.device_addr =
+        reinterpret_cast<std::uint64_t>(static_cast<char*>(device_memory_) + offset);
```
ygwpz: [Warning] Potential null pointer arithmetic if device_memory_ is null

The device_addr calculation uses static_cast<char*>(device_memory_) + offset. If device_memory_ is nullptr (e.g., due to a failed Init that did not properly set it), this pointer arithmetic produces a garbage address value.

While Allocate() checks memory_ is valid at line 201, it does not verify device_memory_ is non-null. Consider adding: if (!device_memory_) return Status::Error(...)

ucm/transport/kv/asu/trans/src/buffer_manager.h
```diff
@@ -34,7 +34,10 @@
 namespace UC::ASU {
 
 struct ScatterGatherEntry {
+    // Local address used by CPU code for SQE packing and completion polling.
     std::uint64_t addr{0};
+    // Device-visible address used by HCOMM/HIXL and remote RDMA operations.
+    std::uint64_t device_addr{0};
```
ygwpz, line 41: [Performance] Struct size increase may impact cache efficiency

Adding device_addr increases ScatterGatherEntry from ~24 bytes to ~32 bytes (assuming 64-bit system). In high-throughput scenarios with many pending SQEs, this 33% size increase could:
1. Increase memory footprint for sub-batch contexts
2. Reduce cache efficiency when processing batches
3. Impact memory bandwidth in tight loops

This is acceptable given the feature requirement, but worth noting for performance-sensitive deployments.

ucm/transport/kv/asu/trans/src/sqe_request.cpp
```diff
@@ -199,7 +199,7 @@ KvBatchStoreRequest BuildBatchStoreRequest(
     request.kv_ns_id = GetTransportConfigAttr<std::uint32_t>(attrs, "kv_ns_id");
     request.dtype = GetTransportConfigAttr<std::uint8_t>(attrs, "dtype");
     request.dspec = GetTransportConfigAttr<std::uint8_t>(attrs, "dspec");
-    request.response_buffer_addr = flagBuffer.addr;
+    request.response_buffer_addr = flagBuffer.device_addr;
```
ygwpz, line 190: [Suggestion] Architectural consistency for user-provided buffer addresses

The response_buffer_addr now correctly uses flagBuffer.device_addr for RDMA operations. However, the entry.buffer_addr at line 212 uses entries[index].buffer.region.addr (CPU address). If user-provided KVBuffer regions are also host-pinned memory, they should similarly use device addresses for RDMA operations.

This creates a semantic inconsistency: internal buffers use device_addr for RDMA, but user buffers use addr. Consider:
1. Adding a device_addr field to MemoryRegion struct
2. Or documenting that user-provided buffers must use device-visible addresses if they are host-pinned
3. Or providing a registration API that captures both addresses

ucm/transport/kv/asu/trans/src/buffer_manager.cpp
yumingyue624: Fixed. Removed BufferRegionCreator and inheritance from AscendBuffer. BufferRegion now owns creation logic and uses AscendBuffer through composition.

ucm/transport/kv/asu/trans/src/buffer_manager.cpp
yumingyue624: Fixed. Removed the inheritance relationship. Callers now use:
BufferRegion::Create(memory_type_, total, region_);
BufferManager stores the complete BufferRegion directly.

ucm/transport/kv/asu/trans/src/asu_submit_flow.cpp
yumingyue624: Fixed. Renamed addr to local_addr and added unified validation for channel, local_addr, device_addr, length, and slot_index of both send and flag buffers.

yumingyue624: Fixed. Renamed addr to local_addr and added unified validation for channel, local_addr, device_addr, length, and slot_index of both send and flag buffers.

ucm/transport/kv/asu/trans/src/buffer_manager.cpp
yumingyue624: Setting deviceAddr = localAddr for HOST memory is intentional: it keeps the BufferRegion layout uniform and avoids special-casing downstream code. Since HOST memory is never used for RDMA, the value of deviceAddr is semantically irrelevant in this case. Added an inline comment to clarify.

ucm/transport/kv/asu/trans/src/buffer_manager.cpp
yumingyue624: Done. Added error logging for both aclrtHostUnregister and aclrtFreeHost in the deleter.

ucm/shared/trans/ascend/ascend_buffer.cc
```diff
@@ -28,6 +28,28 @@
...
+    auto ret = aclrtFreeHost(host);
+    if (ret != ACL_SUCCESS) { UC_ERROR("Failed to free host memory addr={} ret={}", host, ret); }
+}
+
+void ReleaseHostPinnedMemory(void* host)
+{
+    auto ret = aclrtHostUnregister(host);
+    if (ret != ACL_SUCCESS) {
+        UC_ERROR("Failed to unregister host-pinned memory addr={} ret={}", host, ret);
+        return;
```
ygwpz: 🔴 **Critical Bug - Memory Leak**: If `aclrtHostUnregister` fails, this function returns without calling `FreeHostMemory(host)`. This causes a memory leak because the host memory allocated by `aclrtMallocHost` is never freed. The memory will remain allocated until the process exits. Fix: Always call `FreeHostMemory(host)` regardless of whether unregister succeeds or fails:
```cpp
void ReleaseHostPinnedMemory(void* host)
{
    auto ret = aclrtHostUnregister(host);
    if (ret != ACL_SUCCESS) {
        UC_ERROR("Failed to unregister host-pinned memory addr={} ret={}", host, ret);
    }
    FreeHostMemory(host);  // Always free, even if unregister failed
}
```

ucm/transport/kv/asu/trans/include/asu_transport/asu_transport.h
```diff
@@ -72,9 +72,12 @@ struct TransportConfig {
     bool preconnect{true};
     bool bindCqPoller{true};
 
+    // Slot sizes are caller-visible capacities; BufferManager computes the
+    // aligned physical stride used for allocation and memory registration.
     std::size_t sendBufferSlotSize{4160};
     std::size_t sendBufferSlotNum{128};
-    std::size_t flagBufferSlotSize{128};
+    // Maximum memory required by a batch store/retrieve response flag buffer.
+    std::size_t flagBufferSlotSize{71};
```
ygwpz, line 80: ⚠️ **Warning**: `flagBufferSlotSize` changed from 128 to 71. This is a significant reduction (~44%). Please verify that 71 bytes is sufficient for all batch store/retrieve response flag buffers. If this value is too small, it could cause buffer overflows during protocol operations. Consider adding a compile-time or runtime assertion to validate that the actual flag buffer usage never exceeds this limit.

ucm/shared/trans/ascend/ascend_buffer.cc
yumingyue624: Done. Removed the early return after aclrtHostUnregister failure. The failure is still logged, and FreeHostMemory(host) is now always called to avoid leaking the aclrtMallocHost allocation.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -436,23 +463,54 @@ void AsuTransportImpl::PollTaskCompletions(const TransportTaskContextPtr& ctx)
...
+            responseData = flagHeader.data();
+        }
+
+        if (const auto status = protocolManager_->PollResponseCid(responseData, completedCid);
             !status.ok()) {
             continue;
         }
         if (completedCid == 0 || completedCid != subBatchContext.cid) { continue; }
 
+        if (!IsCpuAccessible(subBatchContext.flagBuffer.memory_type)) {
```
Infinite666: Can just use subBatchContext.flagBuffer.memory_type == DEVICE

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -436,23 +463,54 @@ void AsuTransportImpl::PollTaskCompletions(const TransportTaskContextPtr& ctx)
...
+            CompleteSubBatch(*ctx, subBatchContext, status);
+        };
+
         std::uint16_t completedCid = 0;
-        if (const auto status = protocolManager_->PollResponseCid(
-                reinterpret_cast<void*>(subBatchContext.flagBuffer.addr), completedCid);
+        const void* responseData = nullptr;
+        std::array<std::uint8_t, kFlagBufferHeaderCopySize> flagHeader{};
+        std::vector<std::uint8_t> flagBuffer;
+        if (IsCpuAccessible(subBatchContext.flagBuffer.memory_type)) {
```
Infinite666: Can replaced by if memory_type == DEVICE xxx else xxx, so we can remove IsCpuAccessible.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -38,6 +42,29 @@
 
 namespace UC::ASU {
 
+namespace {
+
+constexpr std::size_t kFlagBufferHeaderCopySize = kCqeDwordCount * sizeof(std::uint32_t);
+
+Status CopyDeviceToHost(const ScatterGatherEntry& sge, void* host, std::size_t size,
+                        const char* name)
```
Infinite666: No need to pass name here, we can just return error code and print error log outside.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -436,23 +463,54 @@ void AsuTransportImpl::PollTaskCompletions(const TransportTaskContextPtr& ctx)
...
         std::uint16_t completedCid = 0;
-        if (const auto status = protocolManager_->PollResponseCid(
-                reinterpret_cast<void*>(subBatchContext.flagBuffer.addr), completedCid);
+        const void* responseData = nullptr;
+        std::array<std::uint8_t, kFlagBufferHeaderCopySize> flagHeader{};
+        std::vector<std::uint8_t> flagBuffer;
+        if (IsCpuAccessible(subBatchContext.flagBuffer.memory_type)) {
+            responseData = reinterpret_cast<void*>(subBatchContext.flagBuffer.local_addr);
+        } else {
+            auto status = CopyDeviceToHost(subBatchContext.flagBuffer, flagHeader.data(),
```
Infinite666: Can we reduce it to using it only one copy?

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
yumingyue624: Done. Replaced the CPU-accessible helper check with an explicit `memory_type == MemoryType::ASCEND_DEVICE`
 branch for the flag buffer header/full CQE copy path.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
yumingyue624: Done. Removed `IsCpuAccessible`; device buffers now use the staging/copy path, while non-device buffers use  `local_addr` directly.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
yumingyue624: Done. Removed the `name` parameter from `CopyDeviceToHost`; it now only returns the base copy status. The call sites log the specific flag-buffer header/full-buffer context before completing the sub-batch with the error.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
yumingyue624: The two-step copy is intentional: polling only needs the CQE header, so incomplete CQEs copy 16 bytes instead of the full flag buffer. The full buffer is copied only after the cid matches and the CQE is ready.

Infinite666 (dismissed):

Infinite666 (dismissed):

ucm/shared/trans/ascend/ascend_buffer.cc
```diff
@@ -124,6 +145,33 @@ std::shared_ptr<void> Trans::AscendBuffer::MakeHostBuffer(size_t size)
...
+{
+    if (pDevice) { *pDevice = nullptr; }
+
+    void* host = nullptr;
+    auto ret = aclrtMallocHost(&host, size);
+    if (ret != ACL_SUCCESS) { return nullptr; }
+
+    if (reinterpret_cast<std::uintptr_t>(host) % HOST_REGISTER_PAGE_SIZE != 0) {
+        UC_ERROR("Host-pinned memory is not 4K page aligned addr={} size={}", host, size);
+        FreeHostMemory(host);
```
mag1c-h: If the address returned by malloc is not aligned, it will cause the startup to fail, which is not a reasonable handling strategy.

ucm/shared/trans/buffer.h
```diff
@@ -39,6 +39,7 @@ class Buffer {
 
     virtual std::shared_ptr<void> MakeHostBuffer(size_t size) = 0;
     virtual std::shared_ptr<void> MakeHostBuffer4DirectIo(size_t size) = 0;
+    virtual std::shared_ptr<void> MakeHostPinnedBuffer(size_t size, void** pDevice = nullptr) = 0;
```
mag1c-h, line 42: Why do we need to add a new interface when `MakeHostPinnedBuffer = MakeHostBuffer + RegisterHostBuffer`?

ucm/shared/trans/ascend/ascend_buffer.cc
```diff
@@ -23,11 +23,39 @@
...
+
+void* AlignUp(void* ptr, std::uintptr_t alignment)
+{
+    const auto addr = reinterpret_cast<std::uintptr_t>(ptr);
+    return reinterpret_cast<void*>((addr + alignment - 1) / alignment * alignment);
+}
+
+void ReleaseHostPinnedMemory(void* registeredHost, void* allocatedHost)
+{
+    auto ret = aclrtHostUnregister(registeredHost);
```
mag1c-h: use `Buffer::UnregisterHostBuffer` is better.

ucm/shared/trans/buffer.h
yumingyue624: The main difference is cleanup ownership: the returned buffer must unregister the host mapping before freeing the aclrtMallocHost allocation. Keeping this as a dedicated interface avoids making every caller manually pair MakeHostBuffer + RegisterHostBuffer + unregister/free and reduces the chance of leaks or mismatched cleanup.

ucm/shared/trans/ascend/ascend_buffer.cc
yumingyue624: Done. Reworked this path to guarantee 4K alignment instead of failing on an unaligned aclrtMallocHost result.

ucm/shared/trans/ascend/ascend_buffer.cc
yumingyue624: Done. Replaced the direct aclrtHostUnregister call with Buffer::UnregisterHostBuffer in the host-pinned release path. The aligned host-pinned allocation/release probe and related ASU host-pinned tests passed in the Ascend container.

mag1c-h (approved):

PR #1025, opened 2026-06-12
[Feat] Add opt-in reshape cache event sync

dante159753:
﻿## Summary

This PR splits the cache dump reshape-event synchronization optimization into its own branch, without the dump queue pipeline refactor from `cache-dump-event-optimize`.

Changes:
- Add opt-in `enable_reshape_cache_event_sync` config, defaulting to `false` to preserve current behavior.
- When enabled, layerwise dump first tries to reuse vLLM/vLLM-Ascend `reshape_cache_event` from `attn_metadata[layer_name]`, then falls back to `attn_metadata.reshape_cache_event`.
- If no usable reshape event exists, UCM keeps the existing current-stream event fallback; if event creation fails, it falls back to device synchronization.
- Preserve borrowed torch event references in `Device` so raw event handles remain valid until UCM clears event handles.
- Add focused metrics for event-source selection:
  - `dump_event_reshape_cache_direct_used_total`
  - `dump_event_reshape_cache_layer_used_total`
  - `dump_event_current_stream_used_total`
  - `dump_event_sync_fallback_used_total`
- Update config example and docs.

## Why

For layerwise cache dump, vLLM-Ascend may already record a `reshape_cache_event` immediately after KV cache writes are ready. Waiting on that event lets UCM's C++ cache copy stream synchronize at the KV-ready point instead of recording a later current-stream event in Python.

The switch is opt-in because model/version behavior differs. The default remains the pre-existing path.

## Scope

This PR intentionally does **not** include dump queue optimization or `DumpQueue` changes. The diff is limited to connector/device Python code plus config, metric config, and docs.

## Verification

Ran locally on Windows worktree `cache-dump-event-sync-only`:

- `git diff --check origin/develop...HEAD`
- `python -m py_compile ucm/integration/vllm/device.py ucm/integration/vllm/ucm_connector.py`
- `python -m black --check ucm/integration/vllm/device.py ucm/integration/vllm/ucm_connector.py`
- `python -m isort --check-only --profile=black ucm/integration/vllm/device.py ucm/integration/vllm/ucm_connector.py`
- `codespell docs/source/user-guide/metrics/metrics.md docs/source/user-guide/prefix-cache/pipeline_store.md examples/metrics/metrics_configs.yaml examples/ucm_config_example.yaml ucm/integration/vllm/device.py ucm/integration/vllm/ucm_connector.py --skip ucm/csrc/**,ucm.egg-info/**,.github/**,ucm/sparse/gsa_on_device/csrc/** -L CANN,cann,NNAL,nnal,ASCEND,ascend,EnQue,CopyIn,Collet,DElt,re-use`
- Parsed `examples/metrics/metrics_configs.yaml` and `examples/ucm_config_example.yaml` with PyYAML.
- Confirmed PR diff does not include dump queue files.

Note: full local `pre-commit run --files ...` could not complete on this machine because the actionlint hook attempted to download Go from `go.dev` and hit a local SSL/network failure. The changed files do not include GitHub Actions files; black/isort/codespell were run directly with matching configured versions.

### DIFF
--- a/ucm/integration/vllm/device.py
+++ b/ucm/integration/vllm/device.py
@@ -13,7 +13,7 @@ import subprocess
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from itertools import accumulate
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import torch
 from vllm.platforms import current_platform
@@ -26,11 +26,15 @@ logger = init_logger(__name__)
 class Device(ABC):
     def __init__(self):
         self.events = {}
+        self.borrowed_events = []
 
     @abstractmethod
     def get_event_handle(self) -> int:
         """Return event handle for stream sync. 0 means no event (use synchronize instead)."""
         pass
+
+    def get_event_handle_from_event(self, event: Any) -> int:
+        return 0
 
     @abstractmethod
     def synchronize(self):
@@ -114,11 +118,23 @@ class CudaDevice(Device):
             logger.error(f"get cuda event handle failed. {e}")
             return 0
 
+    def get_event_handle_from_event(self, event: Any) -> int:
+        try:
+            handle = getattr(event, "cuda_event", None)
+            if handle is None or int(handle) == 0:
+                return 0
+            self.borrowed_events.append(event)
+            return int(handle)
+        except Exception as e:
+            logger.error(f"get cuda event handle from existing event failed. {e}")
+            return 0
+
     def synchronize(self):
         torch.cuda.current_stream().synchronize()
 
     def destroy_event_handles(self):
         self.events.clear()
+        self.borrowed_events.clear()
 
     def destroy_event_handle(self, handle: int):
         self.events.pop(handle, None)
@@ -248,6 +264,22 @@ class NpuDevice(Device):
             logger.error(f"get npu event handle failed. {e}")
             return 0
 
+    def get_event_handle_from_event(self, event: Any) -> int:
+        try:
+            handle = getattr(event, "npu_event", None)
+            if handle is None:
+                parameter = getattr(event, "_as_parameter_", None)
+                if callable(parameter):
+                    parameter = parameter()
+                handle = getattr(parameter, "value", None)
+            if handle is None or int(handle) == 0:
+                return 0
+            self.borrowed_events.append(event)
+            return int(handle)
+        except Exception as e:
+            logger.error(f"get npu event handle from existing event failed. {e}")
+            return 0
+
     def synchronize(self):
         torch.npu.current_stream().synchronize()
 
@@ -260,6 +292,7 @@ class NpuDevice(Device):
             except Exception as e:
                 logger.error(f"destroy npu event failed. {e}")
         self.events.clear()
+        self.borrowed_events.clear()
 
     def destroy_event_handle(self, handle: int):
         import acl
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -323,6 +323,7 @@ class UCMDirectConnector(KVConnectorBase_V1):
         self.tp_rank = self._vllm_config.parallel_config.rank
         self.block_size = self._vllm_config.cache_config.block_size
         self.is_mla = self._vllm_config.model_config.is_deepseek_mla
+        self.use_mla = getattr(self._vllm_config.model_config, "use_mla", self.is_mla)
         self.num_layers = self._vllm_config.model_config.get_num_layers(
             self._vllm_config.parallel_config
         )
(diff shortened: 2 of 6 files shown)


ucm/integration/vllm/device.py
```diff
@@ -114,11 +118,23 @@ def get_event_handle(self) -> int:
...
             return 0
 
+    def get_event_handle_from_event(self, event: Any) -> int:
+        try:
+            handle = getattr(event, "cuda_event", None)
+            if handle is None or int(handle) == 0:
+                return 0
+            self.borrowed_events.append(event)
+            return int(handle)
+        except Exception as e:
```
ygwpz, line 128: ⚠️ **Warning - Exception handling too broad**

Using `except Exception` catches all exceptions including system exceptions like `KeyboardInterrupt`, `SystemExit`. This could mask critical errors.

Recommended specific exceptions:
```python
except (AttributeError, TypeError, ValueError) as e:
```

This applies to both CudaDevice (line 128) and NpuDevice (line 279) implementations.

ucm/integration/vllm/device.py
```diff
@@ -248,6 +264,22 @@ def get_event_handle(self) -> int:
...
+            if handle is None:
+                parameter = getattr(event, "_as_parameter_", None)
+                if callable(parameter):
+                    parameter = parameter()
+                handle = getattr(parameter, "value", None)
+            if handle is None or int(handle) == 0:
+                return 0
+            self.borrowed_events.append(event)
+            return int(handle)
+        except Exception as e:
```
ygwpz, line 279: ⚠️ **Warning - Exception handling too broad**

Same issue as CudaDevice - using `except Exception` is too broad. Should catch specific exceptions like `AttributeError, TypeError, ValueError`.

ucm/integration/vllm/device.py
```diff
@@ -114,11 +118,23 @@ def get_event_handle(self) -> int:
...
+            return 0
+
     def synchronize(self):
         torch.cuda.current_stream().synchronize()
 
     def destroy_event_handles(self):
         self.events.clear()
+        self.borrowed_events.clear()
 
     def destroy_event_handle(self, handle: int):
```
ygwpz, line 139: ⚠️ **Warning - Inconsistent cleanup in destroy_event_handle**

The `destroy_event_handle` method only removes events from `self.events` dict but does not remove corresponding entries from `self.borrowed_events`.

If individual event handles are destroyed (rather than batch clearing via `destroy_event_handles`), borrowed event references could accumulate, potentially causing memory growth over time.

Consider either:
1. Adding logic to track handle-to-event mapping for borrowed events
2. Documenting that `destroy_event_handle` is only for UCM-created events, not borrowed ones
3. Using a different structure for `borrowed_events` with proper tracking

ucm/integration/vllm/device.py
```diff
@@ -26,12 +26,16 @@
 class Device(ABC):
     def __init__(self):
         self.events = {}
+        self.borrowed_events = []
```
ygwpz, line 29: 💡 **Suggestion - Thread safety consideration**

The `borrowed_events` list is accessed without synchronization. If this device is used in a multi-threaded context (e.g., multiple workers sharing the same device instance), there could be race conditions when:
- Multiple threads call `get_event_handle_from_event` concurrently (append)
- One thread clears while another is appending

Consider adding:
1. A lock for `borrowed_events` operations
2. Documentation about thread-safety assumptions (single-threaded per device)
3. A thread-safe collection if multi-threaded access is expected

ucm/integration/vllm/ucm_connector.py
```diff
@@ -323,6 +323,7 @@ def __init__(
         self.tp_rank = self._vllm_config.parallel_config.rank
         self.block_size = self._vllm_config.cache_config.block_size
         self.is_mla = self._vllm_config.model_config.is_deepseek_mla
+        self.use_mla = getattr(self._vllm_config.model_config, "use_mla", self.is_mla)
```
ygwpz, line 326: 💡 **Suggestion - Unused variable**

The variable `self.use_mla` is set here but never used in the visible code changes. If this is intentional (for future use), consider adding a comment explaining the purpose. If not needed, it should be removed to avoid confusion.

Note: `self.is_mla` is already used elsewhere in the code.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -826,14 +830,61 @@ def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
...
+                event = getattr(layer_metadata, "reshape_cache_event", None)
+                if event is not None:
+                    return event, "layer"
+            except (KeyError, TypeError, AttributeError):
+                pass
+
+        event = getattr(attn_metadata, "reshape_cache_event", None)
+        if event is not None:
+            return event, "direct"
+        return None, "direct_missing"
```
ygwpz, line 851: ⚠️ **Warning - Metric gap for missing reshape cache events**

When `enable_reshape_cache_event_sync` is True but the reshape cache event is missing (event_source = "none" or "direct_missing"), and we fall back to current stream or device sync, there is no specific metric tracking this "expected-but-missing" case.

This creates an observability gap - operators cannot distinguish between:
- Intentional fallback (feature disabled)
- Expected reshape event missing (vLLM version/model mismatch)

Consider adding metrics like:
- `dump_event_reshape_cache_none_total` (attn_metadata is None)
- `dump_event_reshape_cache_missing_total` (event_source = "direct_missing")

ucm/integration/vllm/ucm_connector.py
```diff
@@ -826,14 +830,61 @@ def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
...
+            if event_handle != 0:
+                if event_source == "direct":
+                    ucmmetrics.update_stats(
+                        "dump_event_reshape_cache_direct_used_total", 1.0
+                    )
+                elif event_source == "layer":
+                    ucmmetrics.update_stats(
+                        "dump_event_reshape_cache_layer_used_total", 1.0
+                    )
+                return event_handle
```
ygwpz, line 880: ⚠️ **Warning - Metric inconsistency in handle extraction failure**

When `enable_reshape_cache_event_sync` is True and we successfully get a `reshape_cache_event` but `get_event_handle_from_event` returns 0 (handle extraction failed), we fall back to the current stream path and track "dump_event_current_stream_used_total".

The metric doesn't reflect that we attempted reshape cache sync first but failed at handle extraction. Consider:
1. Adding a metric for "dump_event_reshape_cache_handle_failed_total"
2. Or updating the existing metric logic to distinguish this fallback path

PR #1032, opened 2026-06-16
Dram pool merge

ayaka836:
## Modifications 
transport部分代码
p2p: add core transport interfaces：基类和部分通用函数定义
p2p: add transport manager：北向接口类transport manager实现
control: add tcp control plane：tcp控制面实现
manager交互流程
image

控制面流程
image

### DIFF
--- a/ucm/transport/p2p/include/one_sided/core/transport.h
+++ b/ucm/transport/p2p/include/one_sided/core/transport.h
@@ -1,0 +1,84 @@
+#pragma once
+
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace transport {
+
+using ManagerID = std::string;
+using MemoryHandle = uint64_t;
+using Metadata = std::vector<uint8_t>;
+
+constexpr MemoryHandle kInvalidMemoryHandle = 0;
+
+struct Endpoint {
+    std::string host = "127.0.0.1";
+    uint16_t port = 0;
+
+    std::string toString() const {
+        return host + ":" + std::to_string(port);
+    }
+};
+
+enum class Opcode {
+    Read,
+    Write,
+};
+
+enum class Status {
+    Ok,
+    InvalidArgument,
+    NotFound,
+    AlreadyExists,
+    Failed,
+};
+
+enum class MemoryType {
+    Host,
+    Device,
+};
+
+struct MemoryRegion {
+    void* addr = nullptr;
+    uint64_t length = 0;
+    MemoryType type = MemoryType::Host;
+    int device_id = -1;
+};
+
+struct InitAttrs {
+    virtual ~InitAttrs();
+};
+
+struct TransferOp {
+    void* local_addr = nullptr;
+    uint64_t remote_addr = 0;
+    uint64_t length = 0;
+};
+
+struct Transfer {
+    Opcode opcode = Opcode::Read;
+    ManagerID target_manager;
+    std::vector<TransferOp> ops;
+};
+
+class Transport {
+   public:
+    virtual ~Transport();
+
+    virtual const char* name() const = 0;
+    virtual Status init(const InitAttrs& options);
+    virtual Status shutdown();
+
+    virtual Status registerMemory(const MemoryRegion& memory);
+    virtual Status unregisterMemory(const MemoryRegion& memory);
+    virtual Status exportMetadata(const ManagerID& manager_id, Metadata& out);
+    virtual Status importMetadata(const ManagerID& manager_id, const Metadata& metadata);
+    virtual Status connectPeer(const ManagerID& manager_id);
+    virtual Status submitTransfer(const Transfer& request);
+};
+
+using TransportPtr = std::shared_ptr<Transport>;
+
+}  // namespace transport
--- a/ucm/transport/p2p/src/one_sided/core/transport_common.cpp
+++ b/ucm/transport/p2p/src/one_sided/core/transport_common.cpp
@@ -1,0 +1,135 @@
+#include "transport_common.h"
+
+#include <climits>
+
+namespace transport::detail {
+
+uint64_t PtrToU64(const void* ptr) {
+    return static_cast<uint64_t>(reinterpret_cast<uintptr_t>(ptr));
+}
+
+void* U64ToPtr(uint64_t value) {
+    return reinterpret_cast<void*>(static_cast<uintptr_t>(value));
+}
+
+bool AppendU64(Metadata& out, uint64_t value) {
+    for (int shift = 56; shift >= 0; shift -= 8) {
+        out.push_back(static_cast<uint8_t>((value >> shift) & 0xff));
+    }
+    return true;
+}
+
+bool ReadU64(const Metadata& input, size_t& offset, uint64_t& value) {
+    if (offset > input.size() || input.size() - offset < sizeof(uint64_t)) {
+        return false;
+    }
+    value = 0;
+    for (size_t i = 0; i < sizeof(uint64_t); ++i) {
+        value = (value << 8) | input[offset + i];
+    }
+    offset += sizeof(uint64_t);
+    return true;
+}
+
+bool AppendU32(Metadata& out, uint32_t value) {
+    out.push_back(static_cast<uint8_t>((value >> 24) & 0xff));
+    out.push_back(static_cast<uint8_t>((value >> 16) & 0xff));
+    out.push_back(static_cast<uint8_t>((value >> 8) & 0xff));
+    out.push_back(static_cast<uint8_t>(value & 0xff));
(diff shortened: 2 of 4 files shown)


ucm/transport/p2p/src/control/tcp_ctrl.cpp
```diff
@@ -0,0 +1,421 @@
...
+    if (status != Status::Ok) {
+        channel.close();
+        return status;
+    }
+
+    status = addPeer(manager_id, std::move(channel));
+    if (status != Status::Ok) {
+        return status;
+    }
+    status = startPeerReceive(manager_id, std::move(metadata_request_handler));
```
yuanzhg078: The **startPeerReceiv**e function is invoked twice consecutively with identical arguments; the second call is redundant and should be removed to avoid unintended side effects or resource conflicts.

ucm/transport/p2p/src/core/transport_manager.cpp
```diff
@@ -0,0 +1,283 @@
...
+#include <utility>
+#include <vector>
+
+namespace transport {
+
+TransportManager::TransportManager(ManagerID manager_id)
+    : manager_id_(std::move(manager_id)) {
+    (void)parseManagerID(manager_id_, local_endpoint_);
+    control_ = std::make_shared<TcpControlPlane>();
+    (void)control_->init(ControlPlaneInitAttrs{});
```
yuanzhg078: The constructor discards all init/listen errors with (void) , creating zombie objects — please either throw on failure or use an explicit Status init() .

ucm/transport/p2p/include/one_sided/core/transport.h
```diff
@@ -0,0 +1,90 @@
...
+enum class Opcode {
+    Send,
+    Read,
+    Write,
+};
+
+enum class Status {
+    Ok,
+    InvalidArgument,
+    NotFound,
```
Infinite666, line 33: Do we really need the status of NotFound and AlreadyExists? Is hard for upper module to deal with so many different status code, upper module only needs to know if the calling is success or not, and we can add error logs inside transport to clarify the different abnormal scenes.

ucm/transport/p2p/src/control/tcp_ctrl_peer.cpp
```diff
@@ -0,0 +1,320 @@
...
+                        std::shared_ptr<Impl::PendingRequest> pending;
+                        {
+                            std::lock_guard<std::mutex> lock(impl_->mutex);
+                            const auto pending_it = impl_->pending_requests.find(control_message.request_id);
+                            if (pending_it != impl_->pending_requests.end()) {
+                                pending = pending_it->second;
+                            }
+                        }
+                        if (pending) {
+                            std::lock_guard<std::mutex> lock(impl_->mutex);
```
yuanzhg078: The pending request lookup and response update should be done under a single `impl_->mutex` lock

yuanzhg078: The pending request lookup and response update should be done under a single `impl_->mutex` lock

ucm/transport/p2p/src/core/transport_manager.cpp
```diff
@@ -0,0 +1,283 @@
...
+        }
+    }
+    return Status::Ok;
+}
+
+ControlEndpoint TransportManager::localEndpoint() const {
+    return local_endpoint_;
+}
+
+ManagerID TransportManager::endpointKey(const ControlEndpoint& endpoint) const {
```
Infinite666: 1. We can write an inner function inside ControlEndpoint like
`std::string toString() const {return host + ":" +std::to_string(port); }`
so we can just call endpoint.toString().
2. We can just put ManagerId inside ControlEndpoint, so we can just use endpoint.id.
3. Is there any circumstances that we need to use host or port alone? If not, there's no need to split them at the  very first.

ucm/transport/p2p/src/control/tcp_ctrl_peer.cpp
```diff
@@ -0,0 +1,320 @@
...
+        impl_->peer_connections[peer] = std::move(connection);
+        channel.impl_->socket = kInvalidSocket;
+    }
+    return Status::Ok;
+}
+
+Status TcpControlPlane::startPeerReceive(const ManagerID& peer, MetadataExchangeHandler metadata_handler) {
+    {
+        std::lock_guard<std::mutex> lock(impl_->mutex);
+        if (impl_->peer_connections.find(peer) == impl_->peer_connections.end()) {
```
yuanzhg078: **startPeerReceive** exceeds 230 lines and handles too many responsibilities in one place.

ucm/transport/p2p/src/core/manager_channel.cpp
```diff
@@ -0,0 +1,293 @@
...
+        const auto status = item.transport->exportMetadata(metadata);
+        if (status != Status::Ok) {
+            return status;
+        }
+        advertisement.records.push_back(TransportMetadataRecord{item.protocol, std::move(metadata)});
+    }
+    return EncodePeerAdvertisement(advertisement, out);
+}
+
+Status TransportManager::buildLocalMetadata(Metadata& metadata, const std::string& transport_name) const {
```
Infinite666: Why not calling exportInstalledMetadata directly?

ucm/transport/p2p/src/core/manager_channel.cpp
```diff
@@ -0,0 +1,293 @@
...
+               Metadata& local_metadata) {
+            return handleMetadataExchange(manager_id, transport_name, remote_metadata, local_metadata);
+        });
+    if (status == Status::Ok) {
+        status = control_->exchangeMetadata(manager_id, transport_name, local, remote);
+    }
+    if (status == Status::Ok) {
+        ManagerID resolved_manager = manager_id;
+        status = establishPeerFromMetadata(remote, resolved_manager, transport_name);
+        if (status == Status::Ok && resolved_manager != manager_id) {
```
yuanzhg078: The resolved_manager != manager_id validation happens after establishPeerFromMetadata() has already updated peer_states_ and imported transport metadata

ucm/transport/p2p/src/core/transport_manager.cpp
```diff
@@ -0,0 +1,284 @@
...
+#include "rdma/rdma_transport.h"
+#include "hixl/hixl_transport.h"
+
+#include "core/manager_meta.h"
+#include "transport_common.h"
+
+#include <cstddef>
+#include <cstdint>
+#include <chrono>
+#include <functional>
```
ygwpz: 🔴 **Critical:** The constructor ignores return values from `parseManagerID` and `control_->init()`. If these fail, subsequent operations may crash. Consider: `if (parseManagerID(...).Failure()) { throw std::runtime_error("Invalid manager_id"); }`

ucm/transport/p2p/src/control/tcp_ctrl_io.cpp
```diff
@@ -0,0 +1,168 @@
...
+}
+
+Status RecvAll(Socket socket, void* data, size_t length) {
+    auto* cursor = static_cast<char*>(data);
+    while (length > 0) {
+        const auto chunk = static_cast<int>(std::min<size_t>(length, 64 * 1024));
+        const int received = recv(socket, cursor, chunk, 0);
+        if (received <= 0) {
+            return Status::Failed;
+        }
```
ygwpz: ⚠️ **Warning:** `errno` is checked after `recv` returns error, but should be captured immediately: `ssize_t received = recv(...); if (received < 0) { int err = errno; ... }`. Another call could modify errno between recv and the check.

ucm/transport/p2p/src/one_sided/core/transport_common.cpp
```diff
@@ -0,0 +1,160 @@
...
+    out.insert(out.end(), name.begin(), name.end());
+    AppendU32(out, static_cast<uint32_t>(metadata.size()));
+    out.insert(out.end(), metadata.begin(), metadata.end());
+    return true;
+}
+
+bool ReadMetadataRecord(const Metadata& input, size_t& offset, std::string& name, Metadata& metadata) {
+    uint32_t name_size = 0;
+    if (!ReadU32(input, offset, name_size) || offset > input.size() || input.size() - offset < name_size) {
+        return false;
```
yuanzhg078, line 118: **ReadMetadataRecord** duplicates the same length-prefixed string parsing logic already used in like **ReadString.** Could this be refactored to share a common implementation with ReadString?

ygwpz (dismissed):
Dram pool merge adds new P2P transport infrastructure with control plane and manager components. The architecture appears well-structured with proper separation of concerns. Documentation is comprehensive.

PR #1033, opened 2026-06-16
[BugFix] lookup for all ranks

flesher0813:
## Purpose
Fix GQA cache lookup correctness by checking cached block availability across all TP ranks instead of only rank 0. This prevents scheduler-side false hits where rank 0 files exist but other rank-specific files have been removed by GC, which can later cause worker load failures.

## Modifications 

- Update get_num_new_matched_tokens() to use the minimum prefix hit across all ranks as the effective external cache hit.
- Add _lookup_external_prefix_all_ranks to search for all ranks' blocks

## Test
Tested with 2tp/8tp/2dp 2tp/4tp 2dcp/2tp 2pcp with QwQ-32B or Qwen2.5-1.5B

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -414,6 +414,9 @@ class UCMDirectConnector(KVConnectorBase_V1):
         defer_scheduler_store = getattr(self, "_defer_scheduler_store", False)
         if role == KVConnectorRole.SCHEDULER:
             self.request_hasher = RequestHasher(vllm_config, 0)
+            self._rank_hashers = [
+                RequestHasher(vllm_config, rank) for rank in range(1, self.tp_size)
+            ]
             self._seed = self.request_hasher("UCM_HASH_SEED")
             # init scheduler-size connector
             if not defer_scheduler_store:
@@ -514,6 +517,36 @@ class UCMDirectConnector(KVConnectorBase_V1):
             ret.append(hash_value)
 
         return ret
+
+    def _lookup_external_prefix_all_ranks(self, external_block_ids: list[bytes]) -> int:
+        rank0_hit_idx = self.store.lookup_on_prefix(external_block_ids)
+        if rank0_hit_idx < 0 or self.is_mla or not self._rank_hashers:
+            return rank0_hit_idx
+
+        candidate_block_ids = external_block_ids[: rank0_hit_idx + 1]
+        num_other_ranks = len(self._rank_hashers)
+        block_ids_other_ranks: list[bytes] = []
+        for block_id in candidate_block_ids:
+            for rank_hasher in self._rank_hashers:
+                block_ids_other_ranks.append(rank_hasher(block_id))
+
+        founds = self.store.lookup(block_ids_other_ranks)
+        expected_founds = len(candidate_block_ids) * num_other_ranks
+        if len(founds) != expected_founds:
+            logger.warning(
+                "Lookup all ranks returned unexpected result length: "
+                f"expected {expected_founds}, got {len(founds)}."
+            )
+            return -1
+
+        for block_idx in range(len(candidate_block_ids)):
+            begin = block_idx * num_other_ranks
+            end = begin + num_other_ranks
+            rank_founds = founds[begin:end]
+            if len(rank_founds) != num_other_ranks or not all(rank_founds):
+                return block_idx - 1
+
+        return len(candidate_block_ids) - 1
 
     def _create_store(
         self,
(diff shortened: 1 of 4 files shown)


ucm/integration/vllm/ucm_connector.py
```diff
@@ -441,6 +445,23 @@ def generate_hash(
 
         return ret
 
+    def _lookup_external_prefix_all_ranks(self, external_block_ids: list[bytes]) -> int:
+        if self.is_mla or len(self._rank_hashers) <= 1:
+            return self.store.lookup_on_prefix(external_block_ids)
+
+        min_hit_idx = len(external_block_ids) - 1
+        for rank, rank_hasher in enumerate(self._rank_hashers):
+            rank_block_ids = (
```
mag1c-h: This loop increases the time spent on Lookup operations. In scenarios with high TP rank, will the performance test pass?

ucm/integration/vllm/ucm_connector.py
```diff
@@ -383,6 +383,10 @@ def __init__(
         defer_scheduler_store = getattr(self, "_defer_scheduler_store", False)
         if role == KVConnectorRole.SCHEDULER:
             self.request_hasher = RequestHasher(vllm_config, 0)
+            self._rank_hashers = [
+                RequestHasher(vllm_config, rank % self.tp_size)
```
ygwpz: 💡 **Suggestion:** The `_rank_hashers` creation uses `rank % self.tp_size`, but this is redundant since `rank` is already in `range(1, self.tp_size)`. Consider simplifying to `for rank in range(1, self.tp_size)` directly, unless there's a specific reason for the modulo operation.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -441,6 +445,27 @@ def generate_hash(
...
+        block_ids_other_ranks: list[bytes] = []
+        for block_id in candidate_block_ids:
+            for rank_hasher in self._rank_hashers:
+                block_ids_other_ranks.append(rank_hasher(block_id))
+
+        founds = self.store.lookup(block_ids_other_ranks)
+        for block_idx in range(len(candidate_block_ids)):
+            begin = block_idx * num_other_ranks
+            end = begin + num_other_ranks
+            if not all(founds[begin:end]):
```
ygwpz: ⚠️ **Warning:** The `all()` check iterates through `founds[begin:end]`. If `founds` is empty or the slice is empty, `all([])` returns `True`, which might not be the intended behavior. Consider adding an explicit check for empty slices.

ucm/store/posix/cc/space_manager.cc
```diff
@@ -58,12 +58,36 @@ Status SpaceManager::Setup(const Config& config)
...
+
+    std::shared_ptr<std::vector<uint8_t>> results;
+    std::shared_ptr<std::atomic<int32_t>> status;
+    std::shared_ptr<Latch> waiter;
+
+    const auto ok = Status::OK().Underlying();
+
+    try {
+        results = std::make_shared<std::vector<uint8_t>>(num, false);
+        status = std::make_shared<std::atomic<int32_t>>(ok);
```
ygwpz, line 71: ⚠️ **Warning:** If `prefixLookupSrv_.NWorker()` returns 0, `waiter->Set(0)` would cause `waiter->Wait()` to return immediately without doing any lookup. The results would remain all false. Consider adding a check for zero workers.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -1353,6 +1382,17 @@ def __init__(
 
         if role == KVConnectorRole.SCHEDULER:
             self.request_hasher = RequestHasher(vllm_config, 0)
+            rank_ids = list(
```
ygwpz, line 1723: 💡 **Suggestion:** The `rank_ids` calculation uses `dict.fromkeys` to deduplicate. Consider documenting the expected behavior when `dcp_world_size > tp_size` - this could produce unexpected results.

ucm/store/posix/cc/space_manager.cc
```diff
@@ -58,12 +58,36 @@ Status SpaceManager::Setup(const Config& config)
 
 Expected<std::vector<uint8_t>> SpaceManager::Lookup(const Detail::BlockId* blocks, size_t num)
 {
-    std::vector<uint8_t> results(num, false);
-    auto res = LookupOnPrefix(blocks, num);
-    if (!res) [[unlikely]] { return res.Error(); }
-    const auto index = res.Value();
-    for (ssize_t i = 0; i <= index; ++i) { results[i] = true; }
-    return results;
+    if (num == 0) { return std::vector<uint8_t>{}; }
```
ygwpz, line 61: ⚠️ **Warning:** The new `Lookup` implementation uses `shared_ptr` for thread-safe access. However, if `prefixLookupSrv_.NWorker()` returns 0, `waiter->Set(0)` would return immediately. Consider adding: `if (nWorker == 0) { return std::vector<uint8_t>(num, false); }`

PR #1038, opened 2026-06-18
Add Ascend Cache IO aggregation copy path

NaganooMei:
## Purpose

This PR adds an IO aggregation path for DeepSeek V4 full-attention (FA) CacheStore transfers on Ascend.

The original path submits one asynchronous H2D/D2H copy for every KV tensor fragment. A V4 FA shard contains many fragments, so these small submissions introduce considerable CPU and runtime scheduling overhead. The new path treats one CacheStore shard as one IO object:

- Load: copy the contiguous host shard to a device staging buffer, then use FFTS SDMA to scatter it to the FA tensor fragments.
- Dump: use FFTS SDMA to gather the FA tensor fragments into a device staging buffer, then copy the contiguous shard to host memory.

## Behavior and scope

- `UCMFAWAConnector` enables `cache_io_aggregation` by default only for the V4 FA store.
- An explicit `cache_io_aggregation: false` setting is preserved and disables IO aggregation for FA.
- The V4 WA store always disables IO aggregation.
- The option can become effective only when built with `RUNTIME_ENVIRONMENT=ascend` (`PLATFORM=ascend`). CUDA, simulation, and `ascend-a3` builds mask it to `false`.
- Under connector-managed configuration, IO aggregation is therefore enabled only for V4 FA on Ascend. Other cases fall back to their existing path; A3 keeps the Cache SDMA Direct default path.
- Cache IO aggregation and Cache SDMA Direct are mutually exclusive. GDR is also rejected when IO aggregation is enabled.

## Implementation

- Add an Ascend-only IO aggregation stream and route CacheStore shard-level multi-fragment copies through it.
- Add per-lane H2D/D2H copy streams, FFTS streams, device staging buffers, and events.
- Reuse `FftsSdmaDispatcher` for scatter/gather between a staging buffer and the KV tensor addresses.
- Use 4 fixed internal lanes, pipeline depth 2, and at most 8 ready FFTS lanes per dispatch. These values are not controlled by `cache_stream_number` in this PR.
- Keep the normal Ascend, CUDA, simulation, GDR, and SDMA Direct paths unchanged when IO aggregation is disabled.
- Map shared CacheStore buffers with `MAP_POPULATE` before host-buffer registration.

## Performance

Test setup:

- DeepSeek V4, DP1 / TP8
- 95% prefix pre-populated
- 1 output token
- Cache capacity was not saturated

| Input length | Concurrency | Requests | TTFT without aggregation (ms) | TTFT with IO aggregation (ms) | Improvement |
| --- | ---: | ---: | ---: | ---: | ---: |
| 32K | 8 | 24 | 2176.9 | 1833.8 | 15.76% |
| 64K | 4 | 12 | 2524.2 | 1977.9 | 21.64% |
| 128K | 2 | 6 | 2875.8 | 2353.6 | 18.16% |


ucm/shared/trans/ascend/ffts/ffts_sdma_dispatcher.cc
```diff
@@ -0,0 +1,154 @@
...
+
+    rtFftsPlusTaskInfo_t task{};
+    task.fftsPlusSqe = &sqe;
+    task.descBuf = contexts_.data();
+    task.descBufLen = sizeof(rtFftsPlusComCtx_t) * contexts_.size();
+    task.descAddrType = RT_FFTS_PLUS_CTX_DESC_ADDR_TYPE_HOST;
+    task.argsHandleInfoNum = 0;
+    task.argsHandleInfoPtr = nullptr;
+
+    completed_ = true;
```
ygwpz, line 125: 🔴 **Critical:** The `completed_` flag is set to `true` BEFORE the actual `rtFftsPlusTaskLaunchWithFlag` call. If the launch fails (line 126-129), `completed_` remains `true`, which will block subsequent `AddMemcpy` calls (since `AddMemcpy` checks `if (completed_)`). This could leave the dispatcher in an unusable state after a failed launch.

**Fix:** Move `completed_ = true;` to AFTER the success check, or reset it on failure:
```cpp
auto ret = rtFftsPlusTaskLaunchWithFlag(...);
if (ret == RT_ERROR_NONE) {
    completed_ = true;
    return Status::OK();
}
// Keep completed_ = false on failure so dispatcher can be reused
UC_ERROR(...);
return Status{...};
```

ucm/shared/trans/ascend/ascend_io_aggregation_stream.cc
```diff
@@ -0,0 +1,164 @@
...
+
+}  // namespace
+
+AscendIoAggregationStream::AscendIoAggregationStream() = default;
+
+AscendIoAggregationStream::~AscendIoAggregationStream() = default;
+
+Status AscendIoAggregationStream::Setup()
+{
+    return Status::InvalidParam("Cache IO aggregation stream requires stream options");
```
ygwpz: 💡 **Suggestion:** The error message "Cache IO aggregation stream requires stream options" is somewhat vague. Consider making it more informative, e.g., "Cache IO aggregation stream requires StreamOptions with cacheIOAggregation=true and tensorSizes specified". This helps users understand what's missing.

ucm/shared/trans/ascend/ffts/ffts_sdma_dispatcher.cc
NaganooMei: Thanks, fixed. `completed_` is now set only after `rtFftsPlusTaskLaunchWithFlag` succeeds, so a failed launch will not leave the dispatcher in a completed state.

ucm/shared/trans/ascend/ascend_io_aggregation_stream.cc
NaganooMei: Thanks for the suggestion. I updated the default `Setup()` error message to explicitly say that `IoAggregationStreamConfig` is required.

ucm/shared/trans/ascend/io_aggregation/ascend_shard_io_aggregator.cc
```diff
@@ -0,0 +1,307 @@
...
+        return Status::InvalidParam("invalid shard IO aggregation tensor number({})", sizes.size());
+    }
+
+    object.specs.clear();
+    object.specs.reserve(sizes.size());
+    auto* stagingBase = static_cast<uint8_t*>(staging);
+    size_t offset = 0;
+    for (size_t i = 0; i < sizes.size(); ++i) {
+        const auto size = sizes[i];
+        if (size == 0 || devices[i] == nullptr) {
```
mag1c-h: There are scenarios in the base class's implementation function that skip nullptr.

NaganooMei: Thanks for pointing this out. Fixed in 8f14687.
BuildScatterSpecs() and BuildGatherSpecs() now skip individual fragments whose size is zero or whose device entry is nullptr, while still advancing the staging-buffer offset by the fragment size. This matches the base Stream vector-overload behavior, while keeping the outer pointer validation unchanged.
The branch has also been rebased onto the latest upstream develop (286678b).

mag1c-h (approved):

PR #1048, opened 2026-06-24
﻿[Feat] ASU key field: 8-byte key in low 8B of 16B slot

yumingyue624:
## Purpose
Align with the updated ASU KV protocol spec: keys are 8 bytes, packed into the low 8 bytes of the 16-byte wire slot (high 8B zero).

## Modifications
1. kv_protocol.cpp: write keys at dword offset +2 within each 16B key slot (Store/Retrieve: target[14]; BatchStore/BatchRetrieve entry: base[3]; Delete/Exist entry: base[2]). High 8B stays zero from caller pre-clearing.
2. kv_protocol.cpp: key size limit 16 -> 8 in all ValidateRequest paths.
3. Tests: shorten all test keys to <= 8 bytes; fix expected-buffer key offsets to match.

## Test
- asu.test: 313 tests passing on aarch64 (GCC 12.3.1).

### DIFF
--- a/ucm/store/asu/cc/asu_store.cc
+++ b/ucm/store/asu/cc/asu_store.cc
@@ -28,10 +28,8 @@
 #include <cctype>
 #include <cstddef>
 #include <functional>
-#include <iomanip>
 #include <memory>
 #include <numeric>
-#include <sstream>
 #include <stdexcept>
 #include <string>
 #include <utility>
@@ -59,9 +57,12 @@ std::uint64_t HashAsuKey(const Detail::BlockId& block)
 
 std::string MakeAsuKey(const Detail::BlockId& block)
 {
-    std::ostringstream os;
-    os << std::hex << std::setfill('0') << std::setw(16) << HashAsuKey(block);
-    return os.str();
+    const auto hash = HashAsuKey(block);
+    std::string key(sizeof(hash), '\0');
+    for (std::size_t i = 0; i < key.size(); ++i) {
+        key[i] = static_cast<char>((hash >> (i * 8)) & 0xff);
+    }
+    return key;
 }
 
 Status ConvertStatus(const AsuStatus& status)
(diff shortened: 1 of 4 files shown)


ucm/transport/kv/asu/trans/src/kv_protocol.cpp
```diff
@@ -161,8 +161,8 @@ Status KvStoreProtocol::PackSqe(const SqeRequest& req, std::uint32_t* target)
     target[10] = r.offset;
     target[11] = (r.lr ? (1U << 31) : 0) | (r.length & 0xFFFFFF);
 
-    std::size_t key_len = std::min(r.key.size(), static_cast<std::size_t>(16));
-    if (key_len > 0) { std::memcpy(&target[12], r.key.data(), key_len); }
+    std::size_t key_len = std::min(r.key.size(), static_cast<std::size_t>(8));
+    if (key_len > 0) { std::memcpy(&target[14], r.key.data(), key_len); }
```
nrj868: Low 8 bytes should be `target[12]` ?

ucm/store/asu/cc/asu_store.cc
```diff
@@ -60,7 +60,8 @@ std::uint64_t HashAsuKey(const Detail::BlockId& block)
 std::string MakeAsuKey(const Detail::BlockId& block)
 {
     std::ostringstream os;
-    os << std::hex << std::setfill('0') << std::setw(16) << HashAsuKey(block);
+    os << std::hex << std::setfill('0') << std::setw(8)
+       << static_cast<std::uint32_t>(HashAsuKey(block));
```
nrj868: Do not cast hash value to `std::uint32_t`, it will increase the likelihood of hash collisions
make asu key as `std::array<std::bytes, 8>`

ucm/transport/kv/asu/trans/src/kv_protocol.cpp
yumingyue624: Done. Fixed to memcpy(&target[12], ...) — the key is now placed in the low 8 bytes (dword 12-13) of the 16B slot, with the high 8 bytes kept zero, matching the little-endian protocol spec (store.xlsx: Dword 12 = keybytes 15:0).

ucm/store/asu/cc/asu_store.cc
yumingyue624: Done. MakeAsuKey now keeps the full 64-bit hash as an 8-byte binary key instead of truncating it to uint32_t. Kept CacheKey as std::string for compatibility with current ASU interfaces.

nrj868 (approved):
LGTM

PR #1049, opened 2026-06-24
control: add tcp control plane

ayaka836:
## Modifications 
Add control plane TCP implementation
image

### DIFF
--- a/ucm/transport/p2p/include/one_sided/control/metadata_channel.h
+++ b/ucm/transport/p2p/include/one_sided/control/metadata_channel.h
@@ -1,0 +1,67 @@
+#pragma once
+
+#include <atomic>
+#include <cstdint>
+#include <functional>
+#include <mutex>
+#include <string>
+#include <thread>
+#include "core/transport.h"
+
+namespace transport {
+
+class MetadataChannel final {
+public:
+    using MetadataRequestHandler =
+        std::function<Status(const Metadata& remote_metadata, Metadata& local_metadata)>;
+
+    MetadataChannel();
+    ~MetadataChannel();
+
+    MetadataChannel(const MetadataChannel&) = delete;
+    MetadataChannel& operator=(const MetadataChannel&) = delete;
+    MetadataChannel(MetadataChannel&&) = delete;
+    MetadataChannel& operator=(MetadataChannel&&) = delete;
+
+    Status Init(const Endpoint& endpoint, MetadataRequestHandler handler);
+    Status ExchangeMetadata(const Endpoint& endpoint, const Metadata& metadata,
+                            Metadata& remote_metadata);
+    void Close();
+
+private:
+    class SocketHandle final {
+    public:
+        SocketHandle();
+        explicit SocketHandle(int socket);
+        ~SocketHandle();
+
+        SocketHandle(const SocketHandle&) = delete;
+        SocketHandle& operator=(const SocketHandle&) = delete;
+        SocketHandle(SocketHandle&& other) noexcept;
+        SocketHandle& operator=(SocketHandle&& other) noexcept;
+
+        bool Valid() const;
+        int Get() const;
+        int Release();
+        void Reset(int socket = -1);
+
+    private:
+        int socket_;
+    };
+
+    Status Listen(const Endpoint& endpoint);
+    Status StartAccepting();
+    Status Connect(const Endpoint& endpoint);
+    Status AcceptSocket(SocketHandle& socket);
+
+    Endpoint endpoint_;
+    SocketHandle listen_socket_;
+    SocketHandle socket_;
+    mutable std::mutex mutex_;
+    std::thread accept_thread_;
+    std::atomic<bool> stop_accept_{false};
+    uint32_t max_receive_frame_size_ = 4 * 1024 * 1024;
+    MetadataRequestHandler metadata_request_handler_;
+};
+
+}  // namespace transport
--- a/ucm/transport/p2p/include/one_sided/core/transport.h
+++ b/ucm/transport/p2p/include/one_sided/core/transport.h
@@ -17,7 +17,7 @@ struct Endpoint {
     std::string host = "127.0.0.1";
     uint16_t port = 0;
 
-    std::string toString() const { return host + ":" + std::to_string(port); }
+    std::string ToString() const { return host + ":" + std::to_string(port); }
 };
 
 enum class Opcode {
@@ -70,16 +70,16 @@ class Transport {
 public:
     virtual ~Transport() = default;
 
-    virtual const char* name() const = 0;
-    virtual Status init(const InitAttrs& options) = 0;
-    virtual Status shutdown() = 0;
+    virtual const char* Name() const = 0;
+    virtual Status Init(const InitAttrs& options) = 0;
+    virtual Status Shutdown() = 0;
 
-    virtual Status registerMemory(const MemoryRegion& memory, MemoryHandle& handle) = 0;
-    virtual Status unregisterMemory(MemoryHandle handle) = 0;
-    virtual Status exportMetadata(const ManagerID& manager_id, Metadata& out) = 0;
-    virtual Status importMetadata(const ManagerID& manager_id, const Metadata& metadata) = 0;
-    virtual Status connectPeer(const ManagerID& manager_id) = 0;
-    virtual Status execute(const Operation& request) = 0;
+    virtual Status RegisterMemory(const MemoryRegion& memory, MemoryHandle& handle) = 0;
+    virtual Status UnregisterMemory(MemoryHandle handle) = 0;
+    virtual Status ExportMetadata(const ManagerID& manager_id, Metadata& out) = 0;
+    virtual Status ImportMetadata(const ManagerID& manager_id, const Metadata& metadata) = 0;
+    virtual Status ConnectPeer(const ManagerID& manager_id) = 0;
+    virtual Status Execute(const Operation& request) = 0;
 };
 
 using TransportPtr = std::shared_ptr<Transport>;
--- a/ucm/transport/p2p/src/one_sided/control/metadata_channel.cpp
+++ b/ucm/transport/p2p/src/one_sided/control/metadata_channel.cpp
@@ -1,0 +1,342 @@
+#include "control/metadata_channel.h"
+#include <algorithm>
+#include <arpa/inet.h>
+#include <cstring>
+#include <mutex>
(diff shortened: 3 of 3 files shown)


ucm/transport/p2p/src/one_sided/control/tcp_ctrl_int.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include "control/tcp_ctrl.h"
```
pyxyzc: What is `tcp_ctrl_int.h` used for?

ucm/transport/p2p/src/one_sided/control/tcp_ctrl.cpp
```diff
@@ -0,0 +1,360 @@
...
+    if (status != Status::Ok) {
+        channel.close();
+        return status;
+    }
+
+    Metadata response_frame;
+    status = ReceiveFrame(channel.impl_->socket, response_frame, impl_->max_receive_frame_size);
+    channel.close();
+    if (status != Status::Ok) {
+        return Status::Failed;
```
yuanzhg078: Why discard the actual status and return Status::Failed instead, losing the error context from ReceiveFrame?

ucm/transport/p2p/src/one_sided/control/tcp_ctrl.cpp
```diff
@@ -0,0 +1,360 @@
...
+            ControlMetadataResponse metadata_response{metadata_status, std::move(local_metadata)};
+            Metadata response_payload;
+            if (EncodeControlMetadataResponse(metadata_response, response_payload) == Status::Ok) {
+                const auto send_status = SendFrame(accepted_socket, response_payload.data(), response_payload.size());
+                if (send_status != Status::Ok) {
+                    CloseSocket(accepted_socket);
+                    continue;
+                }
+            }
+            CloseSocket(accepted_socket);
```
pyxyzc: Plz make sure that the current control plane is designed to handle only one request per connection :)

ucm/transport/p2p/src/one_sided/control/tcp_ctrl.cpp
```diff
@@ -0,0 +1,360 @@
...
+        while (!impl_->stop_accept.load(std::memory_order_relaxed)) {
+            int accepted_socket = kInvalidSocket;
+            const auto status = acceptSocket(accepted_socket);
+            if (status != Status::Ok) {
+                continue;
+            }
+
+            Metadata request_frame;
+            const auto receive_status =
+                ReceiveFrame(accepted_socket, request_frame, impl_->max_receive_frame_size);
```
pyxyzc: Should we consider the circum that the peer connect and then keep the connection open but doesn't send any data?

ucm/transport/p2p/src/one_sided/control/tcp_ctrl_int.h
ayaka836: The impl pattern was used to keep the exposed header files cleaner. However, since the control plane is an internal module, impl is indeed unnecessary here. I will optimize it away and remove this file.

ucm/transport/p2p/src/one_sided/control/tcp_ctrl.cpp
ayaka836: The control plane does not maintain persistent connections; theoretically, the data exchange with each peer occurs only once.

ucm/transport/p2p/src/one_sided/control/tcp_ctrl.cpp
ayaka836: During the exchange, the initiator will perform a send followed by a receive on the same connection.

ucm/transport/p2p/src/one_sided/control/tcp_ctrl.cpp
pyxyzc: I mean we should set timeout for the accepted_socket, otherwise it will block under the circum that the peer connect but not send any msg.

ayaka836: The control plane specifies the data transmission protocol, ensures that connections are not held without active data transfer.

ucm/transport/p2p/src/one_sided/control/tcp_ctrl.cpp
ayaka836: fixed.

ucm/transport/p2p/include/one_sided/control/tcp_ctrl.h
```diff
@@ -0,0 +1,49 @@
...
+
+    TcpControlPlane();
+    ~TcpControlPlane();
+
+    TcpControlPlane(const TcpControlPlane&) = delete;
+    TcpControlPlane& operator=(const TcpControlPlane&) = delete;
+    TcpControlPlane(TcpControlPlane&&) = delete;
+    TcpControlPlane& operator=(TcpControlPlane&&) = delete;
+
+    Status listen(const Endpoint& endpoint);
```
nrj868: Fix naming：  `Listen`   `ExchangeMetadata`

ucm/transport/p2p/include/one_sided/control/tcp_ctrl.h
```diff
@@ -0,0 +1,49 @@
...
+#include <cstdint>
+#include <functional>
+#include <mutex>
+#include <string>
+#include <thread>
+#include "core/transport.h"
+
+namespace transport {
+
+class TcpControlPlane final {
```
pyxyzc: I need to confirm the design that `TcpctrlPlane`—the component within `TcpMgr`, is resp for exchanging metadata. If true, would it be better to rename it? The name "control plane" is easily misunderstood as implying a separate process.

nrj868: Simply using `ControlPlane` would be better
Change filename as well

ucm/transport/p2p/include/one_sided/control/tcp_ctrl.h
```diff
@@ -0,0 +1,49 @@
...
+    TcpControlPlane(const TcpControlPlane&) = delete;
+    TcpControlPlane& operator=(const TcpControlPlane&) = delete;
+    TcpControlPlane(TcpControlPlane&&) = delete;
+    TcpControlPlane& operator=(TcpControlPlane&&) = delete;
+
+    Status listen(const Endpoint& endpoint);
+
+    Status exchangeMetadata(const Endpoint& endpoint, const Metadata& metadata,
+                            Metadata& remote_metadata);
+    Status startAccepting();
```
nrj868: Combine `listen` `startAccepting` into a signle  `Start` method, change stopAccepting to `Stop`
Pass in  local endpoint info、handler through `Init` or `Setup`

ucm/transport/p2p/include/one_sided/control/tcp_ctrl.h
```diff
@@ -0,0 +1,49 @@
...
+    ~TcpControlPlane();
+
+    TcpControlPlane(const TcpControlPlane&) = delete;
+    TcpControlPlane& operator=(const TcpControlPlane&) = delete;
+    TcpControlPlane(TcpControlPlane&&) = delete;
+    TcpControlPlane& operator=(TcpControlPlane&&) = delete;
+
+    Status listen(const Endpoint& endpoint);
+
+    Status exchangeMetadata(const Endpoint& endpoint, const Metadata& metadata,
```
nrj868: What happens if all instance call `exchangeMetadata` at same time

ucm/transport/p2p/include/one_sided/control/tcp_ctrl.h
```diff
@@ -0,0 +1,49 @@
...
+                            Metadata& remote_metadata);
+    Status startAccepting();
+    void setMetadataRequestHandler(MetadataRequestHandler handler);
```
pyxyzc: Plz consider impl encapsulation by passing in both the `localEndpoint` and the `handler` when starting the control plane.

Thus we can expose only a single `serve` interface (which encompasses the existing setMetadataRequestHandler/listen/startAccepting) to the outside.

ucm/transport/p2p/include/one_sided/control/tcp_ctrl.h
```diff
@@ -0,0 +1,49 @@
...
+    void stopAccepting();
+    void close();
+
+private:
+    Status connect(const Endpoint& endpoint);
+    Status acceptSocket(int& socket);
+
+    Endpoint endpoint_;
+    int listen_socket_ = -1;
```
nrj868: Using RAII for socket to avoid manually close

ucm/transport/p2p/include/one_sided/control/tcp_ctrl.h
ayaka836: Taking two instances as an example, two independent connections will be established. The manager will receive metadata on both connections: one returned by the peer in response, and the other proactively sent by the peer. During ImportMetadata, the manager acquires a lock, so the metadata is imported only once.

ucm/transport/p2p/include/one_sided/control/tcp_ctrl.h
ayaka836: fixed.

ucm/transport/p2p/include/one_sided/control/tcp_ctrl.h
ayaka836: fixed.

ucm/transport/p2p/include/one_sided/control/tcp_ctrl.h
ayaka836: Changed to MetadataChannel.

ucm/transport/p2p/include/one_sided/control/tcp_ctrl.h
ayaka836: fixed.

ucm/transport/p2p/include/one_sided/control/tcp_ctrl.h
ayaka836: fixed.

ucm/transport/p2p/include/one_sided/control/tcp_ctrl.h
ayaka836: Changed to MetadataChannel.

nrj868 (approved):
lgtm

PR #1051, opened 2026-06-24
[Feat] Adapt ASU client transport registration

Fengli5355:
## Purpose
Adapt AsuClient to AIV transport provider.

## Modifications 
1. Add C ABI entry points for creating ASU client / transport instances and loading client config, so kv-test can load them through the runtime proxy.
2. Support selecting AICPU / FAKE / AIV transport providers through build options.
3. Extend transport provider config field parsing.
4. Update RegisterRegions / BindRegisteredRegions to preserve and pass tokenId.
5. Use the real active connection handle when registering device memory, and keep the connection alive until unregister.
6. Clean up provider memory handles during unregister.
7. Fix SQE request building to resolve MR keys from the registered handle map.
8. Update related AsuClient and SQE tests.

### DIFF
--- a/ucm/transport/kv/asu/client/src/asu_client_impl.cpp
+++ b/ucm/transport/kv/asu/client/src/asu_client_impl.cpp
@@ -342,7 +342,12 @@ Status AsuClientImpl::RegisterRegionsOnce(const std::vector<MemoryRegion>& regio
     for (std::size_t index = 0; index < regions.size(); ++index) {
         RegisteredMemory registeredRegion;
         registeredRegion.region = regions[index];
-        if (index < results.size()) { registeredRegion.handle = results[index].handle; }
+        if (index < results.size()) {
+            registeredRegion.handle = results[index].handle;
+            registeredRegion.lkey = results[index].lkey;
+            registeredRegion.rkey = results[index].rkey;
+            registeredRegion.tokenId = results[index].tokenId;
+        }
         registeredRegions.emplace_back(registeredRegion);
     }
 
@@ -899,6 +904,9 @@ Status AsuClientImpl::BindRegisteredResources(AsuId asuId,
         RegisteredMemory registeredRegion;
         registeredRegion.region = resource.region;
         registeredRegion.handle = resource.result.handle;
+        registeredRegion.lkey = resource.result.lkey;
+        registeredRegion.rkey = resource.result.rkey;
+        registeredRegion.tokenId = resource.result.tokenId;
         registeredRegions.emplace_back(registeredRegion);
     }
 
--- a/ucm/transport/kv/asu/client/src/asu_client_impl.cpp
+++ b/ucm/transport/kv/asu/client/src/asu_client_impl.cpp
@@ -336,18 +336,23 @@ Status AsuClientImpl::RegisterRegionsOnce(const std::vector<MemoryRegion>& regio
         return WithContext(status, "asuIndex=0 asuId=" + std::to_string(snapshot->asuIds.front()) +
                                        " region_count=" + std::to_string(regions.size()));
     }
+    if (results.size() != regions.size()) {
+        return WithContext(Status::Error(StatusCode::INTERNAL_ERROR,
+                                         "register result count does not match region count"),
+                           "asuIndex=0 asuId=" + std::to_string(snapshot->asuIds.front()) +
+                               " region_count=" + std::to_string(regions.size()) +
+                               " result_count=" + std::to_string(results.size()));
+    }
 
     std::vector<RegisteredMemory> registeredRegions;
     registeredRegions.reserve(regions.size());
     for (std::size_t index = 0; index < regions.size(); ++index) {
         RegisteredMemory registeredRegion;
         registeredRegion.region = regions[index];
-        if (index < results.size()) {
-            registeredRegion.handle = results[index].handle;
-            registeredRegion.lkey = results[index].lkey;
-            registeredRegion.rkey = results[index].rkey;
-            registeredRegion.tokenId = results[index].tokenId;
-        }
+        registeredRegion.handle = results[index].handle;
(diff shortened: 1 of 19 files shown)


ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -178,6 +197,11 @@ Status AsuTransportImpl::Shutdown()
     for (const auto& ctx : taskManager_.GetAll()) {
         if (ctx != nullptr) { (void)taskManager_.Remove(ctx->taskId); }
     }
+    {
```
yuanzhg078, line 201: UnregisterRegions() or **Shutdown()** should call transProvider_->UnregisterMemory(...) to release it. You only clear map here.
Currently, UnregisterRegions() has the unregister logic, but **Shutdown() only clears the local maps, so the shutdown path misses the actual unregister operation.**

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -302,12 +326,57 @@ Status AsuTransportImpl::RegisterRegions(const std::vector<MemoryRegion>& region
...
+        RegisteredMemory regMem;
+        regMem.region = region;
+        regMem.handle = handle;
+        regMem.tokenId = tokenId;  // Only UB is supported for the current version.
+        registeredRegions_[handle] = regMem;
+        registeredRegionStates_[handle] =
+            RegisteredRegionState{connectionHandle, memHandles[0], std::move(connectionChannel)};
+        results.emplace_back(RegisterResult{Status::OK(), handle, 0, 0, tokenId});
     }
     return Status::OK();
```
yuanzhg078, line 419: Scenarios where `RegisterMemory` fails for some regions should be considered; a "partial failure" status needs to be returned to the client.

ucm/transport/kv/asu/client/src/asu_client_impl.cpp
```diff
@@ -334,7 +334,12 @@ Status AsuClientImpl::RegisterRegionsOnce(const std::vector<MemoryRegion>& regio
     for (std::size_t index = 0; index < regions.size(); ++index) {
         RegisteredMemory registeredRegion;
         registeredRegion.region = regions[index];
-        if (index < results.size()) { registeredRegion.handle = results[index].handle; }
+        if (index < results.size()) {
```
Wwwzff: is it possible that idx >= results.size()? If it's not what we expect, add a log or just assert it with an err message

ucm/transport/kv/asu/trans/src/aiv_trans_provider.h
```diff
@@ -29,9 +29,9 @@
 
 namespace UC::ASU {
 
-class AIVTransProvider : public TransProvider {
+class AIVTransProviderAdapter : public TransProvider {
```
Infinite666, line 32: Why do we rename this class?

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -301,13 +340,77 @@ Status AsuTransportImpl::RegisterRegions(const std::vector<MemoryRegion>& region
...
+        regMem.tokenId = tokenId;  // Only UB is supported for the current version.
+        registeredRegions_[handle] = regMem;
+        registeredRegionStates_[handle] =
+            RegisteredRegionState{connectionHandle, memHandles[0], std::move(connectionChannel)};
+        registeredHandles.emplace_back(handle);
+        registeredDescs.push_back(
+            TransProvider::UnregisterMemoryDesc{connectionHandle, memHandles[0]});
+        results.emplace_back(RegisterResult{Status::OK(), handle, 0, 0, tokenId});
+    }
+    if (hasFailure) {
```
Infinite666, line 409: If you want to unregister for registered mem, you should use break instead of continue for the very first failure. Or you can choose to allow partial success.

ucm/transport/kv/asu/trans/src/asu_transport_impl.h
```diff
@@ -121,6 +121,12 @@ class AsuTransportImpl final : public AsuTransport {
 
     void SetTransProvider(std::unique_ptr<TransProvider> provider);
 
+    struct RegisteredRegionState {
```
Infinite666, line 124: 1. Why do we need to stick a channel with a MR?
2. Seems like memhandle and mrhandle are very similar, can we try to merge them?

ucm/transport/kv/asu/trans/src/aiv_trans_provider.h
Fengli5355: This rename solves a bug that AIV and UCM uses the same namespace and class name, which leads to a conliction of ABI. Either UCM or AIV needs to modify the class name.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
Fengli5355: A new PR will fix this.

yuanzhg078: Done. Add break now. We will commit changes in the next pr.

ucm/transport/kv/asu/trans/src/asu_transport_impl.h
yuanzhg078: 2: Agreed. I unified the handle type and now use MRHandle as void* in types.h, removing the separate MemHandle alias.  We will commit changes in the next pr.

yuanzhg078: 1：Because AIV treats device memory registration as connection-context state. The connection handle is needed by AIV to create and later release the MR/token for that device memory.

Infinite666 (approved):

PR #1056, opened 2026-06-26
[Feat] Use fixed 8-byte CacheKey instead of std::string

yumingyue624:
## Purpose
Replace variable-length string keys with a fixed 8-byte byte array (std::array<std::byte, 8>) to simplify packing and eliminate runtime size checks in the KV protocol path.

## Modifications
1. types.h: redefine CacheKey as std::array<std::byte, 8>; add CacheKeyView() helper for string_view interop.
2. kv_protocol: replace empty/size key validation with all-zero check; simplify all PackEntry/PackSqe to fixed memcpy.
3. router: switch HashFunction to string_view; use CacheKeyView for hash inputs in all router implementations.
4. asu_store: build CacheKey via memcpy from hash.
5. Tests: adapt key construction and assertions to fixed-size keys.

## Test
All test passed.

### DIFF
--- a/ucm/transport/kv/asu/client/src/asu_client_impl.cpp
+++ b/ucm/transport/kv/asu/client/src/asu_client_impl.cpp
@@ -78,11 +78,19 @@ std::string FirstFailedSubTaskContext(const ClientTaskContext& ctx)
     return "client_task_id=" + std::to_string(ctx.taskId) + " op=" + ClientOpTypeName(ctx.opType);
 }
 
+std::vector<UC::KV::CacheKey> ToRouterKeys(const std::vector<CacheKey>& keys)
+{
+    std::vector<UC::KV::CacheKey> routerKeys;
+    routerKeys.reserve(keys.size());
+    for (const auto& key : keys) { routerKeys.emplace_back(std::string(CacheKeyView(key))); }
+    return routerKeys;
+}
+
 std::vector<UC::KV::CacheKey> ExtractEntryKeys(const std::vector<KVBuffer>& entries)
 {
     std::vector<UC::KV::CacheKey> keys;
     keys.reserve(entries.size());
-    for (const auto& entry : entries) { keys.emplace_back(entry.key); }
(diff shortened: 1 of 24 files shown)


ucm/transport/kv/common/include/kv_common/router.h
```diff
@@ -27,15 +27,17 @@
 #include <functional>
 #include <memory>
 #include <string>
+#include <string_view>
 #include <unordered_map>
 #include <utility>
 #include <vector>
+#include "asu_transport/types.h"
```
nrj868: `Router` will be used by DramPool too, using ASU's header file will break its independence.
Suggest to change `CacheKey` to string_view

ucm/transport/kv/kv-test/include/kv_test/cache_key_utils.h
```diff
@@ -0,0 +1,69 @@
+#pragma once
+
+#include <algorithm>
+#include <cstddef>
```
Fengli5355: Common functions could be extracted and placed in this file. Check redundant definitions.

nrj868 (approved):

ucm/transport/kv/asu/client/src/asu_client_impl.cpp
```diff
@@ -78,11 +78,19 @@ std::string FirstFailedSubTaskContext(const ClientTaskContext& ctx)
     return "client_task_id=" + std::to_string(ctx.taskId) + " op=" + ClientOpTypeName(ctx.opType);
 }
 
+std::vector<UC::KV::CacheKey> ToRouterKeys(const std::vector<CacheKey>& keys)
+{
+    std::vector<UC::KV::CacheKey> routerKeys;
+    routerKeys.reserve(keys.size());
+    for (const auto& key : keys) { routerKeys.emplace_back(std::string(CacheKeyView(key))); }
```
nrj868, line 85: Nit: Optmize this during perf

PR #1062, opened 2026-06-29
tcp: add standalone tcp transport

ayaka836:
#### Module Purpose
This MR adds TcpMessageChannel, a TCP control message channel.It encapsulates TCP connection establishment, connection management, frame encoding/decoding, and background IO handling.
##### Init(const Endpoint& local)
Initializes the local listening endpoint, creates the listen socket, and starts the background IO thread.
##### Send(const Endpoint& peer, const void* data, size_t length)
Sends one message frame to the specified peer.
- If a default connection already exists, it is reused for sending.
- If no usable connection exists, actively connects to the peer and registers the new connection as the default sending connection.
- The message frame format is: sender endpoint + payload.
- A successful Send() only means the data was written to the local TCP socket. It does not provide message-level ACK or retransmission semantics.
##### Receive(Endpoint& peer, Metadata& data)
Blocks until one message frame is received.
- Returns the peer endpoint and payload.
- The background IO thread parses complete frames from the TCP byte stream and pushes them into the receive queue.

### DIFF
--- a/ucm/transport/p2p/src/two_sided/tcp/tcp_message_channel.cpp
+++ b/ucm/transport/p2p/src/two_sided/tcp/tcp_message_channel.cpp
@@ -1,0 +1,530 @@
+#include "two_sided/tcp/tcp_message_channel.h"
+#include <algorithm>
+#include <arpa/inet.h>
+#include <cerrno>
+#include <condition_variable>
+#include <cstdint>
+#include <cstring>
+#include <deque>
+#include <future>
+#include <mutex>
+#include <netdb.h>
+#include <sys/epoll.h>
+#include <sys/socket.h>
+#include <thread>
+#include <unistd.h>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+#include "transport_common.h"
+
+#ifndef MSG_NOSIGNAL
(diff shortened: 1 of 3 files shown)


ucm/transport/p2p/src/two_sided/tcp/tcp_message_channel.cpp
```diff
@@ -0,0 +1,530 @@
...
+        Message message;
+        if (!DecodeMessage(frame, message.peer, message.data)) {
+            drop_socket = true;
+            continue;
+        }
+
+        {
+            std::lock_guard<std::mutex> lock(mutex_);
+            auto connection_it = connections_.find(socket);
+            if (connection_it == connections_.end()) {
```
yuanzhg078, line 376: HandleConnectionEvent: skip receive_queue_.emplace_back when connections_.find(socket) == end() or BindPeerLocked fails, otherwise an unregistered peer leaks into the receive queue.

ayaka836: A decoded frame is treated as a valid message even if the connection is no longer registered. The connection may be dropped, but the decoded message is still delivered to the receive queue by design.

nrj868 (approved):

PR #1081, opened 2026-07-09
[WIP] Add buffer manager base for buffer delegator

pyxyzc:
## Purpose

This change introduces three local structures:

- `BufferMgr`
- `BufferPool`
- `Buffer`

The goal is to provide a small RAII-based HBM buffer manager for buffer delegator. The manager allocates one fixed-size device memory pool during initialization and serves fixed-size aligned scratch buffers from that pool.

### BufferMgr

`BufferMgr` is the top-level entry point owned by `AsuStore`.

### BufferPool

`BufferPool` owns the actual device memory pool.

### Buffer

`Buffer` is a move-only RAII handle for one acquired slot.

Slot addresses are calculated as:

```cpp
deviceAddr = baseAddr + slotId * slotSize
```

## Modifications 

- `buffer_delegator.h`


ucm/store/delegator/cc/buffer_manager.h
```diff
@@ -0,0 +1,111 @@
...
+    struct Allocator {
+        std::function<void*(std::size_t)> allocate;
+        std::function<void(void*)> free;
+    };
+
+    struct Config {
+        std::size_t alignmentBytes{0};
+        std::size_t alignedSize{0};
+        std::size_t poolSizeBytes{0};
+        Allocator allocator;
```
Infinite666, line 27: Why is the allocator inside config?

ucm/store/delegator/cc/buffer_manager.h
```diff
@@ -0,0 +1,111 @@
...
+    };
+
+    BufferMgr() = default;
+    ~BufferMgr() = default;
+
+    BufferMgr(const BufferMgr&) = delete;
+    BufferMgr& operator=(const BufferMgr&) = delete;
+
+    Status Init(Config config);
+    Expected<Buffer> Acquire();
```
Infinite666, line 37: Seems like BufferMgr only have one `Acquire` method exposed to the outside class. Maybe we should discuss about the usage of BufferMgr or you can put the design in pr description.

ucm/store/delegator/cc/buffer_manager.h
```diff
@@ -0,0 +1,111 @@
...
+
+private:
+    bool initialized_{false};
+    std::size_t alignmentBytes_{0};
+    std::size_t alignedSize_{0};
+    std::size_t poolSizeBytes_{0};
+    std::shared_ptr<BufferPool> pool_;
+};
+
+class BufferPool : public std::enable_shared_from_this<BufferPool> {
```
Infinite666, line 55: Take BufferManager in ASU as a reference, it's based on IndexPool, which is a lock free logical slot manager.

ucm/store/delegator/cc/buffer_manager.h
```diff
@@ -0,0 +1,111 @@
...
+    Buffer(const Buffer&) = delete;
+    Buffer& operator=(const Buffer&) = delete;
+
+    Buffer(Buffer&& other) noexcept;
+    Buffer& operator=(Buffer&& other) noexcept;
+
+    explicit operator bool() const noexcept { return IsValid(); }
+    bool IsValid() const noexcept { return pool_ != nullptr; }
+
+    std::uint64_t DeviceAddr() const noexcept { return deviceAddr_; }
```
Infinite666, line 94: No need to specify the addr is a device addr.

ucm/store/delegator/cc/buffer_manager.cc
```diff
@@ -0,0 +1,178 @@
...
+    alignmentBytes_ = config.alignmentBytes;
+    alignedSize_ = config.alignedSize;
+    poolSizeBytes_ = slotCount * config.alignedSize;
+    initialized_ = true;
+    return Status::OK();
+}
+
+Expected<Buffer> BufferMgr::Acquire()
+{
+    if (!initialized_) {
```
Infinite666, line 152: Is there any secenario that we may call Acquire before Init?

ucm/store/delegator/cc/buffer_manager.cc
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    alignmentBytes_ = config.alignmentBytes;
+    alignedSize_ = config.alignedSize;
+    poolSizeBytes_ = slotCount * config.alignedSize;
+    initialized_ = true;
+    return Status::OK();
+}
+
+Expected<Buffer> BufferMgr::Acquire()
```
Infinite666, line 150: Is there any other logic needs to be handled by BufferMgr? If not, this layer is too thin.

ucm/store/delegator/cc/buffer_manager.h
pyxyzc: Will add design notes in PR later :)

PR #1091, opened 2026-07-13
[test] Add model-aware E2E H2D and S2D bandwidth benchmarks

NaganooMei:
## Purpose

Add model-aware E2E bandwidth benchmarks for validating UCM dump and load performance across Cache|Fake and Cache|Posix pipelines.

Users only need to adjust the configuration at the top of each script before running the benchmark.

## Modifications

- Add `cache_fake_bw.py` for Cache|Fake dump/load bandwidth testing.
- Add `cache_posix_bw.py` for Cache|Posix dump/load bandwidth testing.
- Support GLM-5.2, MiniMax-M2.7 TP8, and DeepSeek-V4 profiles.
- Support layerwise and non-layerwise transfers.
- GLM-5.2 and MiniMax-M2.7 TP8 support both transfer modes.
- DeepSeek-V4 always uses its complete non-layerwise tensor layout.
- Generate non-layerwise layouts using 78 layers for GLM-5.2 and 62 layers for MiniMax-M2.7 TP8.
- Model MLA behavior by letting worker 0 dump data and all workers load the same cache blocks.
- Model GQA behavior by letting each worker dump and load its own cache blocks.
- Use one worker process per rank for both dump and load.
- Create one scheduler in worker 0 to check backend readiness before load starts.
- Use the `spawn` multiprocessing start method.
- Add optional NUMA-aware CPU affinity for Python workers and UCM store threads.
- Add worker failure detection and cleanup for interrupted or suspended runs.
- Add five warmup epochs excluded from statistics.
- Report average, minimum, P50, P90, P99, and maximum latency and bandwidth.
- Use a 32 GiB Cache buffer.
- Use 16 measured dump epochs and 128 measured load epochs.
- Add a 15 ms interval between adjacent epochs.
- Use `psync` as the default Posix I/O engine.
- Keep SDMA Direct and CPU affinity disabled by default.
- Document user-adjustable parameters next to their definitions.

## Usage

Select the model, transfer mode, worker count, and layer count at the top of the script.

GLM-5.2 example:

    model_name = "glm-5.2"
    use_layerwise = True
    worker_number = 16
    layer_number = 78

MiniMax-M2.7 TP8 example:

    model_name = "minimax-m2.7 tp8"
    use_layerwise = False
    worker_number = 8
    layer_number = 62

The built-in MiniMax tensor sizes are for TP8. Other TP configurations must adjust the tensor size list accordingly.

DeepSeek-V4 example:

    model_name = "dsv4"
    worker_number = 8

DeepSeek-V4 always uses its complete non-layerwise tensor layout, regardless of `use_layerwise` and `layer_number`.

Run the benchmarks:

    python3 ucm/store/test/e2e/cache_fake_bw.py
    python3 ucm/store/test/e2e/cache_posix_bw.py

## Test

- Passed Python AST syntax validation for both benchmark scripts.
- Passed model profile and transfer-mode validation.
- Passed `git diff --check`.
- Passed pre-commit formatting checks.
- Runtime execution requires an Ascend NPU environment.


ucm/store/test/e2e/cache_posix_bw.py
```diff
@@ -0,0 +1,437 @@
...
+
+# Fill tensor_size_list with the per-layer tensor byte sizes of the target
+# deployment before running a profile. MLA writes once from worker 0 and all
+# workers load the same block ids, while GQA workers use their own block ids.
+MODEL_PROFILES = {
+    "glm-5.2": {
+        "worker_mode": "mla",
+        "worker_number": 8,
+        "share_buffer_enable": True,
+        "tensor_size_list": [131072, 65536, 32768],
```
flesher0813: should be  [131072, 16384, 32768]

NaganooMei: thanks, fixed.

ygwpz (approved):

qyh111 (approved):

PR #1095, opened 2026-07-14
[Feat] Implement LRU eviction policy

hcygit1:
## What

- Add an LRU eviction policy based on `std::list` and `std::unordered_map`.
- Move accessed entries to the most-recently-used position.
- Select eviction candidates from the least-recently-used end.
- Skip entries that are not ready, referenced, or still leased.
- Validate eviction ratios and align ratio rounding with the position eviction policy.

## Tests

- Add, duplicate add, delete, and reinsert tests.
- Verify LRU ordering and access updates.
- Verify eviction ratio handling.
- Verify invalid and oversized ratios.
- Verify non-ready, referenced, and leased entries are skipped.
- Verify selected entries are marked as `DELETING`.

## Verification

- Targeted LRU compilation and behavior tests passed.
- clang-format 20 check passed.
- `git diff --check` passed.
- Full CMake/CTest was not run locally because the current macOS environment is incompatible with the project's Linux linker flags.

### DIFF
--- a/ucm/store/dram/cc/lru_eviction_policy.h
+++ b/ucm/store/dram/cc/lru_eviction_policy.h
@@ -79,11 +79,11 @@ public:
     std::vector<BlockId> GetEvictionResults(double evictRatio) override
     {
         std::vector<BlockId> victims;
-        if (evictRatio <= 0.0 || index_.empty()) { return victims; }
+        if (!std::isfinite(evictRatio) || evictRatio <= 0.0 || index_.empty()) { return victims; }
 
-        evictRatio = std::min(evictRatio, 1.0);
-        const auto target = static_cast<std::size_t>(
-            std::ceil(static_cast<double>(index_.size()) * evictRatio));
+        const double boundedRatio = std::min(evictRatio, 1.0);
+        const auto target =
+            static_cast<std::size_t>(static_cast<double>(index_.size()) * boundedRatio);
         const auto now = std::chrono::system_clock::now();
 
         for (auto it = lruList_.rbegin(); it != lruList_.rend() && victims.size() < target; ++it) {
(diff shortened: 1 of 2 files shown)


ucm/store/dram/cc/lru_eviction_policy.h
```diff
@@ -0,0 +1,116 @@
...
+private:
+    using ListIter = std::list<BlockId>::iterator;
+
+    struct LruNode {
+        EntryPtr entry;
+        ListIter iter;
+    };
+
+    std::list<BlockId> lruList_;
+    std::unordered_map<BlockId, LruNode, UC::Detail::BlockIdHasher> index_;
```
Infinite666: Can we use list to store the entries and map to store the iterators? If we can do that, we only need to use list in `GetEvictionResults` instead of using both.

hcygit1: ok

ucm/store/test/case/dram/dram_lru_eviction_policy_test.cc
```diff
@@ -0,0 +1,266 @@
...
+
+namespace {
+using Clock = std::chrono::system_clock;
+using TimePoint = Clock::time_point;
+using UC::DramStore::Entry;
+using UC::DramStore::EntryPtr;
+using UC::DramStore::EntryStatus;
+using UC::DramStore::LruEvictionPolicy;
+
+EntryPtr MakeEntry(UC::Detail::BlockId key, EntryStatus status = EntryStatus::READY,
```
Infinite666: Use the common function in `dram_test_common.h` for `MakeEntry` and `KeyFromHex`.

hcygit1: Updated. The LRU eviction policy tests now reuse MakeEntry and KeyFromHex from dram_test_common.h. The test-specific arguments were adjusted to match the common helper signature.

Infinite666 (approved):

PR #1100, opened 2026-07-15
[Feat] hixlTransport support multi-instance

ayaka836:
## Purpose
When HIXL communicates over RoCE, the underlying network resources are bound to the device/NIC associated with the current ACL runtime context. As a result, a single HIXL instance can only represent the resources of one device/NIC and cannot utilize multiple available NICs on the host process. 
This commit adds multi-instance support to the HIXL transport.

## Modifications 
* `HixlInitAttrs` now supports configuring multiple HIXL instances. Each instance creates an independent HIXL engine and runs in its own worker thread, where the corresponding ACL device context is initialized during thread startup.
* If an instance is configured with a port less than 0, an available local port is automatically selected during initialization.
* After importing peer metadata, the transport builds a route to that peer. Route selection will be further topology-aware in future updates. The current strategy is:

  * Avoid selecting a local instance that belongs to the same host and device as the remote instance.
  * Among the remaining candidates, randomly select one of the local instances with the fewest existing routes.
  * Use the selected route for all subsequent `Connect` and transfer operations to that peer.
* The current implementation assumes that the remote peer exports a single HIXL instance. If multiple remote instances are advertised, only the first one is used as a fallback.

### DIFF
--- a/ucm/transport/p2p/include/one_sided/core/transport_init_attrs.h
+++ b/ucm/transport/p2p/include/one_sided/core/transport_init_attrs.h
@@ -3,14 +3,20 @@
 #include <cstdint>
 #include <map>
 #include <string>
+#include <vector>
 #include "core/transport.h"
 
 namespace transport {
 
 struct HixlInitAttrs : public InitAttrs {
-    std::string local_engine;
-    std::map<std::string, std::string> options;
-    int device_id = -1;
+    struct Instance {
+        int32_t port = 0;
+        int32_t device_id = -1;
+        std::map<std::string, std::string> options;
+    };
+
+    std::string ip = "127.0.0.1";
+    std::vector<Instance> instances;
     int32_t connect_timeout_ms = 1000;
     int32_t transfer_timeout_ms = 1000;
 };
--- a/ucm/transport/p2p/include/one_sided/hixl/hixl_transport.h
+++ b/ucm/transport/p2p/include/one_sided/hixl/hixl_transport.h
@@ -1,19 +1,24 @@
 #pragma once
 
+#include <cstddef>
 #include <cstdint>
-#include <map>
+#include <memory>
 #include <mutex>
+#include <shared_mutex>
 #include <string>
 #include <unordered_map>
-#include "acl/acl.h"
+#include <vector>
 #include "core/transport.h"
 #include "core/transport_init_attrs.h"
-#include "hixl/hixl.h"
+#include "hixl/hixl_types.h"
 
 namespace transport {
 
-struct HixlTransportMetadata {
-    std::string local_engine;
+class HixlInstance;
+
+struct HixlInstanceInfo {
+    Endpoint endpoint;
+    int32_t device_id = -1;
 };
 
 class HixlTransport final : public Transport {
@@ -25,8 +30,8 @@ public:
     HixlTransport& operator=(const HixlTransport&) = delete;
 
     TransportProtocol Protocol() const override;
-    Status Init(const InitAttrs& options) override;
-    Status Init(const HixlInitAttrs& options);
+    Status Init(const InitAttrs& attrs) override;
+    Status Init(const HixlInitAttrs& attrs);
     Status Shutdown() override;
     Status RegisterMemory(const MemoryRegion& memory, MemoryHandle& handle) override;
     Status UnregisterMemory(MemoryHandle handle) override;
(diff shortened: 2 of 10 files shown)


ucm/transport/p2p/include/one_sided/core/transport_init_attrs.h
```diff
@@ -3,14 +3,20 @@
...
 
 namespace transport {
 
 struct HixlInitAttrs : public InitAttrs {
-    std::string local_engine;
-    std::map<std::string, std::string> options;
-    int device_id = -1;
+    struct Instance {
+        int32_t port = 0;
+        int device_id = -1;
```
Wwwzff: **int32_t** xxx
**int** xxxx

use a unified type instead

ucm/transport/p2p/include/one_sided/hixl/hixl_transport.h
```diff
@@ -39,32 +53,81 @@ class HixlTransport final : public Transport {
...
+        HixlTask task;
+        std::promise<Status> result;
+    };
+
+    struct HixlInstance {
+        Endpoint local_endpoint;
+        int device_id = -1;
+        std::unique_ptr<hixl::Hixl> hixl;
+        std::thread worker;
+        std::mutex mutex;
```
Wwwzff: underlist variables which are protected by mutex and add a comment

ucm/transport/p2p/src/one_sided/hixl/hixl_transport.cpp
```diff
@@ -1,28 +1,99 @@
...
+
+    addrinfo* results = nullptr;
+    if (getaddrinfo(host.c_str(), "0", &hints, &results) != 0) { return Status::Failed; }
+
+    Status status = Status::Failed;
+    for (auto* item = results; item != nullptr; item = item->ai_next) {
+        const int candidate = socket(item->ai_family, item->ai_socktype, item->ai_protocol);
+        if (candidate < 0) { continue; }
+
+        if (bind(candidate, item->ai_addr, static_cast<socklen_t>(item->ai_addrlen)) == 0) {
```
Wwwzff: it seems to be a redundant cast for ai_addrlen

ucm/transport/p2p/src/one_sided/hixl/hixl_transport.cpp
```diff
@@ -47,71 +118,171 @@ HixlTransport::~HixlTransport()
...
-                 static_cast<int>(get_context_status));
-        if (previous_context != nullptr) { (void)aclrtSetCurrentContext(previous_context); }
-        return Status::Failed;
+    auto options_copy = options;
+    std::promise<Status> start_result;
+    auto start_future = start_result.get_future();
+
+    instance.worker = std::thread([&instance, start_result = std::move(start_result),
+                                   options_copy = std::move(options_copy)] mutable {
+        const auto set_device_status = aclrtSetDevice(instance.device_id);
```
Wwwzff: should call aclrtResetDevice when returns or there're resource leaks

ucm/transport/p2p/src/one_sided/hixl/hixl_transport.cpp
```diff
@@ -47,71 +118,171 @@ HixlTransport::~HixlTransport()
...
+        if (set_device_status != ACL_ERROR_NONE) {
+            UC_ERROR("[Transport][HIXL] set device failed: aclrtSetDevice({}) returned {}",
+                     instance.device_id, static_cast<int>(set_device_status));
+            start_result.set_value(Status::Failed);
+            return;
+        }
+
+        instance.hixl = std::make_unique<hixl::Hixl>();
```
Wwwzff: It seems that the hixl are initialized on thread but deconstructed on the main process.

ucm/transport/p2p/src/one_sided/hixl/hixl_transport.cpp
```diff
@@ -47,71 +118,171 @@ HixlTransport::~HixlTransport()
...
-    if (previous_context != nullptr && previous_context != context) {
-        const auto restore_status = aclrtSetCurrentContext(previous_context);
-        if (restore_status != ACL_ERROR_NONE) {
-            UC_ERROR("transport hixl restore context failed: aclrtSetCurrentContext returned {}",
-                     static_cast<int>(restore_status));
-            hixl_.Finalize();
-            return Status::Failed;
+    {
+        std::lock_guard<std::mutex> lock(instance.mutex);
+        instance.initialized = false;
```
Wwwzff: how about using atomic bool instead?

ucm/transport/p2p/src/one_sided/hixl/hixl_transport.cpp
```diff
@@ -47,71 +118,171 @@ HixlTransport::~HixlTransport()
...
+        const auto set_device_status = aclrtSetDevice(instance.device_id);
+        if (set_device_status != ACL_ERROR_NONE) {
+            UC_ERROR("[Transport][HIXL] set device failed: aclrtSetDevice({}) returned {}",
+                     instance.device_id, static_cast<int>(set_device_status));
+            start_result.set_value(Status::Failed);
+            return;
+        }
+
+        instance.hixl = std::make_unique<hixl::Hixl>();
+        std::map<hixl::AscendString, hixl::AscendString> hixl_options;
```
Wwwzff: pls reserve size for this container

ucm/transport/p2p/src/one_sided/hixl/hixl_transport.cpp
```diff
@@ -47,71 +118,171 @@ HixlTransport::~HixlTransport()
...
+Status HixlTransport::Init(const InitAttrs& attrs)
 {
-    const auto* attrs = dynamic_cast<const HixlInitAttrs*>(&options);
-    return attrs == nullptr ? Status::InvalidArgument : Init(*attrs);
+    const auto* hixl_attrs = dynamic_cast<const HixlInitAttrs*>(&attrs);
+    return hixl_attrs == nullptr ? Status::InvalidArgument : Init(*hixl_attrs);
 }
 
-Status HixlTransport::Init(const HixlInitAttrs& options)
+Status HixlTransport::StartInstance(HixlInstance& instance,
```
Infinite666: Can we make HixlInstance as a inner class of HixlTransport? HixlTransport can locate the needed instance and call instance.xxx(). That maybe a more clear structure.

ucm/transport/p2p/include/one_sided/hixl/hixl_transport.h
```diff
@@ -39,32 +53,81 @@ class HixlTransport final : public Transport {
...
+        std::condition_variable cv;
+        std::deque<std::shared_ptr<QueuedTask>> tasks;
+        bool stopping = false;
+        bool initialized = false;
+    };
+
+    struct RemoteInstance {
```
Infinite666: There is an accumulation of too many similar, relatively small structs, can we think it through and code more readable?

ucm/transport/p2p/include/one_sided/core/transport_init_attrs.h
ayaka836: Fixed.

ucm/transport/p2p/include/one_sided/hixl/hixl_transport.h
ayaka836: Fixed.

ucm/transport/p2p/src/one_sided/hixl/hixl_transport.cpp
ayaka836: Fixed.

ucm/transport/p2p/src/one_sided/hixl/hixl_transport.cpp
ayaka836: Fixed.

ucm/transport/p2p/src/one_sided/hixl/hixl_transport.cpp
ayaka836: Fixed.

ucm/transport/p2p/include/one_sided/hixl/hixl_transport.h
ayaka836: Fixed.

ucm/transport/p2p/src/one_sided/hixl/hixl_transport.cpp
ayaka836: std::map does not support reserve() because it is implemented as a red-black tree, not contiguous storage or a hash table.

ucm/transport/p2p/src/one_sided/hixl/hixl_transport.cpp
ayaka836: It is not necessary. initialized, together with stopping, the task queue, and the condition variable, forms a composite state that must be kept consistent under the same mutex.

Infinite666 (approved):

PR #1103, opened 2026-07-15
[Test] A2 scheduler probe 1

dante159753:
## What changed

Adds one temporary documentation-only probe file.

## Why

This PR is used to verify that the A2 on-demand scheduler detects queued PR gate jobs, splits card 7 during the hourly window, and starts available self-hosted runners.

## Impact

No product or runtime behavior changes. The PR and branch will be removed after scheduler validation.

## Verification

- `git diff --check origin/develop...HEAD`
- GitHub PR Gate, especially `test-e2e-pc-a2`


dante159753:
Temporary A2 scheduler probe completed: the hourly window dispatched this job successfully, and the runner returned the card to whole-device mode after the queue drained.

PR #1106, opened 2026-07-15
[Feat] Add register KV cache regions in UCM connector and ASU store

yuanzhg078:
## Purpose
Enable the vLLM UCM connector to collect and register KV-cache memory regions with ASU pipeline stores.
## Modifications
Update UCMDirectConnector to collect KV-cache memory regions after pipeline-store initialization.
Register K and V cache tensors as separate regions when the pipeline store requires KV-cache registration.
Support supported KV-cache tensor layouts, skip empty or duplicate regions, and reject unsupported cache types or tensor shapes.
Add NeedRegisterKVCaches and RegisterKVCaches to the UCM store interface.
Expose KV-cache registration through the pipeline-store C++ and Python bindings.
Make ASU stores opt in to registration; other stores report that it is not required.
## Test
Added an ASU store test confirming KV-cache registration is enabled and currently succeeds as a no-op.
Verified Python syntax for the updated connector and pipeline-store wrapper.

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -401,6 +401,9 @@ class UCMDirectConnector(KVConnectorBase_V1):
         )
 
         self.store = self._create_store(self.kv_cache_layout, store_cores)
+        if self.store.need_register_kv_caches():
+            registrations = self._collect_kv_cache_registrations()
+            self.store.register_kv_caches(registrations)
 
         if worker_cores:
             try:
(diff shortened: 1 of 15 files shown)


ucm/store/pipeline/connector.py
```diff
@@ -157,6 +157,22 @@ def wait(self, task: Task) -> None:
...
+            base_addrs
+            if isinstance(base_addrs, np.ndarray)
+            else np.array(base_addrs, dtype=np.uint64)
+        )
+        sizes_np = (
+            sizes if isinstance(sizes, np.ndarray) else np.array(sizes, dtype=np.uint64)
+        )
+        self.store_.RegisterPersistentRegions(addrs_np, sizes_np)
+
+    def unregister_kv_cache_regions(self) -> None:
```
Fengli5355: Where is this function called?

ucm/integration/vllm/ucm_connector.py
```diff
@@ -401,6 +401,13 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
         )
 
         self.store = self._create_store(self.kv_cache_layout, store_cores)
+        config = self.connector_configs[0].get("ucm_connector_config", {})
```
Infinite666: Do not specify the difference between stores, the better way is make register as a common interface for every store, if a store don't need it, it will be a empty function.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -412,6 +419,43 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
+        range_groups: list[list[tuple[int, int]]] = []
+        unique_regions: set[tuple[int, int]] = set()
+
+        def add_tensor(group_index: int, tensor: torch.Tensor) -> None:
+            addr = tensor.data_ptr()
+            size = tensor.numel() * tensor.element_size()
+            region = (addr, size)
+            if addr == 0 or size <= 0:
+                return
+            if region in unique_regions:
```
Infinite666: In what kind of scenario will we have duplicate region?

yuanzhg078: Normally, two independent allocations are made per layer; when **KV cache sharing** is used in vllm ascend, multiple layers share a single allocation.

ucm/store/pipeline/connector.py
yuanzhg078: Done! The ASU store unregisters the regions internally during destruction, so I have removed this redundant Python-level wrapper.

ucm/store/ucmstore_v1.h
```diff
@@ -156,6 +160,19 @@ class StoreV1 {
     StoreV1() = default;
 };
 
+struct KVCacheRegion {
+    std::uintptr_t addr{0};
+    std::size_t size{0};
+};
+
+class KVCacheRegistrar {
```
Infinite666: Why do we still need this? Just add a function inside StoreV1.

Infinite666 (dismissed):

ucm/store/ucmstore_v1.h
yuanzhg078: Fix now.

ucm/integration/vllm/ucm_connector.py
yuanzhg078: Done

Infinite666 (approved):

Wwwzff (approved):

PR #1116, opened 2026-07-17
[Feat] Add dram pool buffer manager

Infinite666:
## Purpose
1. Implement `BufferManager` of dram pool, it contains multiple `BufferPool` for different buffer sizes. It helps to choose `BufferPool` based on the given size.
2. Complete the TODOs relied on `BufferManager` in metadata management.
**TODO: Adaption for SIMU env to run gtest in pipeline.**

## Modifications
1. `buffer.h`: Common `Buffer` struct for `BufferManager` and `Entry`.
3. `buffer_manager.cc/.h`: Implement `BufferManager` class.
4. `eviction_policy.h`: Modify the return value of `GetEvictionResults` from `BlockId` to `Entry` for the later free of buffer slot.
5. `metadata.cc/.h`: Move the gc thread from `MetadataManager` to `DramPoolServer`, so that all the background threads are maintained by `DramPoolServer`.
6. `drampool_server.cpp/.h`: Replaced `BufferPoolList` by `BufferManager`.

## Test
```
[==========] Running 71 tests from 5 test suites.
[----------] Global test environment set-up.
[----------] 19 tests from UCLruEvictionPolicyTest (5 ms total)
[----------] 22 tests from UCShardMetadataTest (0 ms total)
[----------] 4 tests from UCMetadataManagerTest (6 ms total)
[----------] 14 tests from UCPosEvictionPolicyTest (0 ms total)
[----------] 12 tests from UCTtlEvictionPolicyTest (0 ms total)

[----------] Global test environment tear-down
[==========] 71 tests from 5 test suites ran. (4660 ms total)
[  PASSED  ] 71 tests.
```

PR #1126, opened 2026-07-20
[Feat] Add DramPool KV protocol

yumingyue624:
## Purpose

* Share one KV wire protocol between DramPool and DramStore

## Modifications

* Add Dump, Load, and Lookup request packing and validation
* Add packed per-item responses with a pollable readiness status
* Add an API for checking response readiness before unpacking results

## Test
```
  [==========] Running 32 tests from 1 test suite.
  [----------] 32 tests from KvProtocolTest (0 ms total)
  [==========] 32 tests from 1 test suite ran. (0 ms total)
  [  PASSED  ] 32 tests.
```


ucm/store/test/case/dram/kv_protocol_test.cpp
```diff
@@ -0,0 +1,751 @@
+/**
```
Infinite666, line 1: The test files should be inside `ucm/store/test/case/dram` folder.

ucm/store/test/case/dram/kv_protocol_test.cpp
```diff
@@ -0,0 +1,751 @@
...
+#include <cstdint>
+#include <cstring>
+#include <gtest/gtest.h>
+#include <string>
+#include <vector>
+
+namespace UC::DramPool {
+namespace {
+
+BlockId MakeKey(std::uint8_t seed)
```
Infinite666: Can reuse `KeyFromHex` / `https://github.com/ModelEngine-Group/unified-cache-management/pull/1116` function to get fake key or entry in `dram_test_common.h`.

ucm/store/test/case/dram/kv_protocol_test.cpp
yumingyue624: Done. Moved the KV protocol tests to ucm/store/test/case/dram/drampool

ucm/store/dram/cc/kv_protocol.h
```diff
@@ -0,0 +1,282 @@
...
+    virtual Status UnpackResponse(const void* data, std::uint16_t result_count,
+                                  KvResponse& out) const = 0;
+
+    // Server side: unpack request wire -> struct; pack response struct -> wire.
+    virtual Status UnpackRequest(const void* data, std::size_t size,
+                                 std::unique_ptr<KvRequest>& out) const = 0;
+    virtual Status PackResponse(void* data, const KvResponse& resp) const = 0;
+};
+
+class KvDumpProtocol : public KvProtocol {
```
Infinite666: Could you consider whether using a template would be a better option?

yumingyue624: After evaluating the template-based approach, the request classes still required many duplicated protocol-specific function declarations and implementations, without providing a clear simplification. Therefore, we kept the current polymorphic design for better readability and maintainability.

ucm/store/test/case/dram/kv_protocol_test.cpp
yumingyue624: Done.

Infinite666 (approved):

PR #1133, opened 2026-07-21
[Feat] Add delegator copy stream and buffer pool

pyxyzc:
## Purpose

Introduce delegator infrastructure for asynchronous device-to-device transfers.

  ## Modifications

  - Add `Delegator::CopyStream`:
    - Create and manage multiple Ascend copy streams.
    - Support asynchronous device-to-device copies with parameter validation.
    - Synchronize and destroy managed streams during cleanup.

  - Extend `BufferPool::Slot`:
    - Expose the slot's byte offset relative to the pool base address.

## Test

image

image


ucm/store/delegator/cc/delegator_copy_stream.cc
```diff
@@ -0,0 +1,113 @@
...
+        if (ret != ACL_SUCCESS) {
+            result = AclStatus("aclrtSynchronizeStream", ret);
+        }
+    }
+    return result;
+}
+
+void CopyStream::Reset()
+{
+    if (device_id_ >= 0) { (void)BindDevice(); }
```
Infinite666: Why do we call aclrtSetDevice in Reset?

ucm/store/delegator/cc/delegator_copy_stream.cc
```diff
@@ -0,0 +1,113 @@
...
+    if (stream == nullptr || destination == nullptr || source == nullptr || size == 0 ||
+        size > destination_capacity) {
+        return Status::InvalidParam("invalid delegator D2D copy");
+    }
+    const auto ret = aclrtMemcpyAsync(destination, destination_capacity, source, size,
+                                      ACL_MEMCPY_DEVICE_TO_DEVICE, stream);
+    return ret == ACL_SUCCESS ? Status::OK() : AclStatus("aclrtMemcpyAsync", ret);
+}
+
+Status CopyStream::Synchronize()
```
Infinite666: We only need the semantics of synchronize all streams?

ucm/store/delegator/cc/delegator_copy_stream.cc
```diff
@@ -0,0 +1,113 @@
...
+    return Status::Error(std::string(operation) + " failed: " + std::to_string(error));
+}
+
+}  // namespace
+
+CopyStream::~CopyStream() { Reset(); }
+
+Status CopyStream::Setup(std::int32_t device_id, std::size_t stream_number)
+{
+    if (!streams_.empty()) { return Status::DuplicateKey(); }
```
Infinite666: The return error duplicate key doesn't seems very relavant.

ucm/store/delegator/cc/delegator_copy_stream.h
```diff
@@ -0,0 +1,59 @@
...
+    Status DeviceToDeviceAsync(aclrtStream stream, void* destination,
+                               std::size_t destination_capacity, const void* source,
+                               std::size_t size);
+    Status Synchronize();
+    std::size_t Size() const { return streams_.size(); }
+
+private:
+    void Reset();
+
+    std::vector<aclrtStream> streams_;
```
Infinite666, line 56: Can we try to reuse the AscendStream which is already encapsulated?

ucm/store/delegator/cc/delegator_buffer_manager.h
```diff
@@ -0,0 +1,51 @@
...
+ * SOFTWARE.
+ */
+#pragma once
+
+#include <cstddef>
+#include "pool/buffer_pool.h"
+
+namespace UC::Delegator {
+
+class BufferManager {
```
Infinite666: Looks like the logic of BufferManager and BufferPool is very similar to me, maybe we can try to add a new function in BufferPool to get the offset of slot, so we can use BufferPool directly?

pyxyzc: Done. Directly use BufferPool :)

ucm/store/delegator/cc/delegator_copy_stream.h
pyxyzc: AscendStream lack of D2D ability thus retain the current design, maybe refactor later :)

ucm/store/delegator/cc/delegator_copy_stream.cc
pyxyzc: Done.

pyxyzc: And add device check :)

ucm/store/delegator/cc/delegator_copy_stream.cc
pyxyzc: Done. And add device check :)

ucm/store/delegator/cc/delegator_copy_stream.cc
pyxyzc: Add single & all interface :)

Infinite666 (approved):

Wwwzff (approved):

PR #1135, opened 2026-07-21
[Feat] AsuClient modifies the process of registration

Fengli5355:
## Purpose
Adapt AsuClient to senarios where there are multiple ASUs.
For temprary implementation, the very first transport (ASU) does the actual registering and unregistering; the followers does binding and unbinding to avoid redundant operations.
In the future, we plan to remove the *owner* transport and let Client chooses which transport does the action of reg and unreg. This needs several modifications or refactors in AIV / AICPU TransProvider.

## Modifications 
1. In commit 28b3fc4, asu_port is modified from an integer to a vector.
2. In commit e90178e, we add Bind and Unbind interface in Client, Transport, and TransProvider which is currently placeholder to be discussed.

## Test
Only for discussion in this stage.

### DIFF
--- a/ucm/transport/kv/asu/client/src/asu_client_impl.cpp
+++ b/ucm/transport/kv/asu/client/src/asu_client_impl.cpp
@@ -355,6 +355,7 @@ Status AsuClientImpl::RegisterRegionsOnce(const std::vector<MemoryRegion>& regio
     }
 
     Status finalStatus = Status::OK();
+    std::vector<std::pair<AsuId, std::shared_ptr<AsuTransport>>> boundTransports;
     for (std::size_t asuIndex = 1; asuIndex < snapshot->asuIds.size(); ++asuIndex) {
         auto iter = snapshot->transports.find(snapshot->asuIds[asuIndex]);
         if (iter == snapshot->transports.end()) {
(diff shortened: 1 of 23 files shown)


ucm/transport/kv/asu/trans/include/aiv_transport/aiv_transport.h
```diff
@@ -66,6 +66,15 @@ class AIVTransport {
...
                                   std::vector<MRHandle>& mrHandles) = 0;
 
+    virtual Status BindMemory(const std::vector<RegisteredMemory>& regions,
+                              std::vector<MRHandle>& mrHandles) = 0;
+
+    struct UnbindMemoryDesc {
+        MRHandle mrHandle;
+    };
+
+    virtual std::vector<Status> UnbindMemory(const std::vector<UnbindMemoryDesc>& memoryDescs) = 0;
```
yuanzhg078, line 76: You can pass the mrHandle directly here; there is no need to use UnbindMemoryDesc.

ucm/transport/kv/asu/client/src/asu_client_impl.cpp
```diff
@@ -796,13 +817,41 @@ Status AsuClientImpl::UnregisterRegionsOnce(const std::vector<MRHandle>& handles
...
+        const auto id = snapshot->asuIds.front();
+        const auto iter = snapshot->transports.find(id);
+        if (iter == snapshot->transports.end()) {
+            if (finalStatus.ok()) {
+                finalStatus = WithContext(
+                    Status::Error(StatusCode::NOT_FOUND, "first asu transport not found"),
+                    "asuId=" + std::to_string(id));
+            }
+        } else {
+            auto status = iter->second->UnregisterRegions(handles);
```
yuanzhg078, line 849: When a follower unbind fails, finalStatus is already non-OK, but the owner is still unregistered; since the cache is not removed afterward, registeredResources_ retains a handle that is no longer valid on the owner.

ucm/transport/kv/asu/client/src/asu_client_impl.cpp
```diff
@@ -375,24 +376,44 @@ Status AsuClientImpl::RegisterRegionsOnce(const std::vector<MemoryRegion>& regio
...
                                 " region_count=" + std::to_string(registeredRegions.size()));
-        } else if (status.ok() && childResults.size() != registeredRegions.size() &&
-                   finalStatus.ok()) {
-            finalStatus =
-                WithContext(PartialFailed("one or more asu region bindings failed"),
-                            "asuIndex=" + std::to_string(asuIndex) +
-                                " asuId=" + std::to_string(snapshot->asuIds[asuIndex]) +
-                                " region_count=" + std::to_string(registeredRegions.size()) +
-                                " result_count=" + std::to_string(childResults.size()));
+        } else if (status.ok()) {
```
Infinite666, line 379: Seems like this status here is always ok.

Fengli5355: This status could be the status of Provider::BindMemory, or Provider::GetMemTokenId, which is not always ok().

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -207,6 +207,16 @@ Status AsuTransportImpl::Shutdown()
     }
     {
         std::lock_guard<std::mutex> lock(registeredRegionsMu_);
+        std::vector<MRHandle> boundHandles;
+        boundHandles.reserve(boundRegisteredRegionHandles_.size());
+        for (const auto& item : boundRegisteredRegionHandles_) {
```
yuanzhg078, line 212: `boundRegisteredRegionHandles_` duplicates information already stored in `registeredRegions_`: the bound local handle is available as `registeredRegions_[ownerHandle].handle`.

PR #1136, opened 2026-07-21
[CI] build: integrate p2p transport with CMake

ayaka836:
## Modifications 
- Add transport  CMake
  - Non-Ascend: Skip P2P build.
  - Ascend with complete HIXL dependencies: Compile Manager, TCP/metadata, and HixlTransport.
  - Ascend with missing HIXL, libcann_hixl, or libmetadef: Still compile and install Manager, TCP, and metadata, but do not compile or link HIXL-related code.Calling InstallTransport(TransportProtocol::Hixl, ...) in this case returns Status::Unsupported(). 
- Reuse UCM status in p2p transport

### DIFF
--- a/ucm/transport/CMakeLists.txt
+++ b/ucm/transport/CMakeLists.txt
@@ -1,2 +1,4 @@
+if(RUNTIME_ENVIRONMENT STREQUAL "ascend")
+    add_subdirectory(p2p)
+endif()
 add_subdirectory(kv)
-add_subdirectory(p2p)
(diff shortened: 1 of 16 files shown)


ucm/transport/CMakeLists.txt
```diff
@@ -1,2 +1,4 @@
+if(BUILD_UCM_ASU)
```
Infinite666: Transport module should not bound with BUILD_UCM_ASU.

ucm/transport/p2p/CMakeLists.txt
```diff
@@ -0,0 +1,166 @@
+if(NOT RUNTIME_ENVIRONMENT STREQUAL "ascend")
```
Infinite666, line 1: Use runtime env to determine which kind of transport can be built.

Infinite666 (approved):

Infinite666 (dismissed):

Infinite666 (approved):

PR #1146, opened 2026-07-23
[FEAT] Add variable-size buffer pool

hcygit1:
## Purpose

Add a variable-size buffer pool for allocating differently sized buffers from one contiguous memory region.

## Modifications

- Add a hardened OffsetAllocator implementation.
- Add VariableBufferPool with configurable allocation alignment.
- Support host, host-pinned, and Ascend device memory.
- Add allocation offset, size tracking, zeroing, and thread safety.
- Add CMake integration and shared pool test fixtures.
- Add allocator and variable buffer pool unit tests.

## Test

- VariableBufferPool: 17/17 passed
- OffsetAllocator 32-bit mode: 22/22 passed
- OffsetAllocator 16-bit mode: 22/22 passed
- Pool regression tests: 51/51 passed
- ASan/UBSan: passed

Real Ascend/CANN validation is pending on the server environment.


ucm/shared/pool/variable_buffer_pool.cc
```diff
@@ -0,0 +1,241 @@
...
+    return true;
+}
+
+Status VariableBufferPool::BufferRegion::Create(MemoryType memory_type, std::size_t size,
+                                                BufferRegion& region)
+{
+    Trans::AscendBuffer ascendBuffer;
+    switch (memory_type) {
+        case MemoryType::HOST: {
+            auto owner = ascendBuffer.MakeHostBuffer(size);
```
Infinite666: Please refer to the latest `buffer_pool.cc` to use `Device` instead of `AscendBuffer`.

ucm/shared/pool/variable_buffer_pool.h
```diff
@@ -0,0 +1,128 @@
...
+ * successful Allocate with exactly one successful Free and must not call Free
+ * concurrently for the same handle. If memory clearing fails, the allocation
+ * remains live and the caller must retry Free or Reset the pool. Init and Reset
+ * must not run concurrently with allocation methods.
+ */
+class VariableBufferPool {
+    static constexpr std::size_t kDefaultAllocationAlignment = 64;
+
+public:
+    enum class MemoryType {
```
Infinite666: If we are using the same class like `buffer_pool.h`, please move them to a common file.

ucm/shared/pool/variable_buffer_pool.h
```diff
@@ -0,0 +1,128 @@
...
+    bool IsInitialized() const { return static_cast<bool>(region_) && allocator_ != nullptr; }
+
+    const std::string& GetName() const { return name_; }
+    MemoryType GetMemoryType() const { return memory_type_; }
+    std::size_t GetTotalSize() const { return total_capacity_; }
+    void* GetLocalAddr() const { return region_.local_addr; }
+    void* GetDeviceAddr() const { return region_.device_addr; }
+
+private:
+    struct BufferRegion {
```
Infinite666: Same like `MemoryType`.

ucm/shared/pool/variable_buffer_pool.cc
```diff
@@ -0,0 +1,241 @@
...
+            Reset();
+            return status;
+        }
+    }
+    return Status::OK();
+}
+
+Status VariableBufferPool::Allocate(std::size_t requested_size, AllocationHandle& handle)
+{
+    std::lock_guard<std::mutex> lock(mutex_);
```
Infinite666: If you need mutex lock, you can put it inside the allocator to make the lock range smaller.

ucm/shared/pool/variable_buffer_pool.h
```diff
@@ -0,0 +1,128 @@
...
+    static constexpr std::size_t kDefaultAllocationAlignment = 64;
+
+public:
+    enum class MemoryType {
+        HOST = 0,
+        HOST_PINNED = 1,
+        ASCEND_DEVICE = 2,
+    };
+
+    struct AllocationHandle {
```
Infinite666: Maybe `BufferHandle` is a better name for code understanding.

ucm/shared/pool/detail/offset_allocator.h
```diff
@@ -0,0 +1,123 @@
...
+    Allocator(uint32 size, uint32 maxAllocs = DEFAULT_MAX_ALLOCS);
+    ~Allocator();
+
+    Allocator(const Allocator&) = delete;
+    Allocator& operator=(const Allocator&) = delete;
+    // Allocator ownership remains bound to the arena that provides synchronization.
+    Allocator(Allocator&&) = delete;
+    Allocator& operator=(Allocator&&) = delete;
+
+    void reset();
```
Infinite666: Please make sure the coding style is consistent, the method name should start with a upper letter.

ucm/shared/pool/detail/offset_allocator.h
```diff
@@ -0,0 +1,123 @@
...
+#endif
+
+static constexpr uint32 NUM_TOP_BINS = 32;
+static constexpr uint32 BINS_PER_LEAF = 8;
+static constexpr uint32 TOP_BINS_INDEX_SHIFT = 3;
+static constexpr uint32 LEAF_BINS_INDEX_MASK = 0x7;
+static constexpr uint32 NUM_LEAF_BINS = NUM_TOP_BINS * BINS_PER_LEAF;
+
+struct Allocation {
+    static constexpr uint32 NO_SPACE = 0xffffffff;
```
Infinite666: The static variables should be moved out from the struct.

ucm/shared/pool/detail/offset_allocator.cc
```diff
@@ -0,0 +1,458 @@
...
+    // Only splitting a free region consumes a spare metadata node.
+    if (remainderSize > 0 && m_freeOffset == 0) {
+        return {Allocation::NO_SPACE, Allocation::NO_SPACE_METADATA};
+    }
+
+    node.dataSize = size;
+    node.used = true;
+    unlinkNodeFromBin(binIndex, nodeIndex);
+    m_freeStorage -= nodeTotalSize;
+#ifdef DEBUG_VERBOSE
```
Infinite666: Please use `UC_DEBUG` instead.

ucm/shared/pool/detail/offset_allocator.h
```diff
@@ -0,0 +1,123 @@
...
+    static constexpr NodeIndex NO_SPACE_METADATA = std::numeric_limits<NodeIndex>::max();
+
+    uint32 offset = NO_SPACE;
+    NodeIndex metadata = NO_SPACE_METADATA;
```
Infinite666: Maybe `node` or `nodeIndex` is a better name.

ucm/shared/pool/variable_buffer_pool.cc
```diff
@@ -0,0 +1,241 @@
...
+
+        const auto byteOffset =
+            static_cast<std::size_t>(handle.allocation.offset) * allocation_alignment_;
+        allocatedSize = static_cast<std::size_t>(allocatedUnits) * allocation_alignment_;
+        if (byteOffset > total_capacity_ || allocatedSize > total_capacity_ - byteOffset) {
+            return Status::InvalidParam(name_ + ": allocation is outside the buffer region");
+        }
+
+        localAddress = static_cast<char*>(region_.local_addr) + byteOffset;
+        enableZero = enable_zero_;
```
Infinite666: Just use the private variable directly is ok.

ucm/shared/pool/variable_buffer_pool.cc
```diff
@@ -0,0 +1,241 @@
...
+{
+    std::size_t allocatedSize = 0;
+    void* localAddress = nullptr;
+    bool enableZero = false;
+
+    {
+        std::lock_guard<std::mutex> lock(mutex_);
+        if (!IsInitialized()) { return Status::Error("buffer pool not initialized"); }
+
+        const auto allocatedUnits = allocator_->allocationSize(handle.allocation);
```
Infinite666: 1. If the logics below is only needed when enable zero is true, why do we calculate them before the judgement?
2. Can we try to simplify or get rid of this logic by recording the desired attributes when allocating?

ucm/shared/pool/variable_buffer_pool.cc
hcygit1: Done. The shared buffer creation logic now uses Trans::Device{}.MakeBuffer() instead of AscendBuffer.

ucm/shared/pool/variable_buffer_pool.h
hcygit1: Done. The shared memory type is now defined as BufferMemoryType in buffer_region.h, and both buffer pools reuse it.

ucm/shared/pool/variable_buffer_pool.h
hcygit1: Done. BufferRegion and its creation/reset logic have been moved to the shared buffer_region.h/.cc files.

ucm/shared/pool/variable_buffer_pool.cc
hcygit1: Done. The mutex is now managed inside OffsetAllocator, which reduces the lock scope to individual allocator operations.

ucm/shared/pool/variable_buffer_pool.h
hcygit1: Done. AllocationHandle has been renamed to BufferHandle.

ucm/shared/pool/detail/offset_allocator.h
hcygit1: Done. The allocator methods now start with uppercase letters to match the project coding style.

ucm/shared/pool/detail/offset_allocator.h
hcygit1: Done. The sentinel constants have been moved from Allocation to namespace scope.

ucm/shared/pool/detail/offset_allocator.cc
hcygit1: Done. The original debug output has been replaced with UC_DEBUG.

ucm/shared/pool/detail/offset_allocator.h
hcygit1: Done. metadata has been renamed to nodeIndex, and the related sentinel has been renamed to NO_SPACE_NODE_INDEX.

ucm/shared/pool/variable_buffer_pool.cc
hcygit1: Done. The unnecessary temporary variables have been removed, and the private handle members are now used directly.

ucm/shared/pool/variable_buffer_pool.cc
hcygit1: Done. The non-zeroing path now frees the allocation directly. The address and allocated size required for zeroing are recorded in BufferHandle during allocation and reused during Free.

ucm/shared/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 add_subdirectory(vendor)
 add_subdirectory(infra)
 add_subdirectory(trans)
+if(RUNTIME_ENVIRONMENT STREQUAL "ascend" OR RUNTIME_ENVIRONMENT STREQUAL "simu"
```
Infinite666: Just put `add_subdirectory(pool/detail)` under `add_subdirectory(pool)`.

ucm/shared/pool/buffer_region.h
```diff
@@ -0,0 +1,52 @@
...
+
+namespace UC {
+
+enum class BufferMemoryType {
+    HOST = 0,
+    HOST_PINNED = 1,
+    ASCEND_DEVICE = 2,
+};
+
+namespace PoolDetail {
```
Infinite666: Can remove the namespace since all the other class or struct in pool don't have their specific namespace.

ucm/shared/pool/variable_buffer_pool.cc
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    if (!enable_zero_) {
+        if (!allocator_->Free(handle.allocation_)) {
+            return Status::InvalidParam(name_ + ": invalid allocation handle");
+        }
+        return Status::OK();
+    }
+
+    if (allocator_->GetAllocationSize(handle.allocation_) == 0) {
```
Infinite666: Looks like this function is more like a validation, you can put this validation logic inside `Free`.

ucm/shared/pool/detail/offset_allocator.cc
```diff
@@ -0,0 +1,453 @@
...
+}
+
+uint32 Allocator::GetAllocationSize(Allocation allocation) const
+{
+    if (allocation.nodeIndex == NO_SPACE_NODE_INDEX ||
+        static_cast<uint32>(allocation.nodeIndex) >= m_maxAllocs) {
+        return 0;
+    }
+
+    std::lock_guard<std::mutex> lock(m_mutex);
```
Infinite666: Refer to the comment in `buffer_region.cc`, if we move this validation logic inside `Free`, we don't need to acquire for lock.

ucm/shared/pool/detail/offset_allocator.cc
```diff
@@ -0,0 +1,453 @@
...
+    if (!m_nodes) { return 0; }
+
+    const Node& node = m_nodes[allocation.nodeIndex];
+    if (!node.used || node.dataOffset != allocation.offset) { return 0; }
+    return node.dataSize;
+}
+
+StorageReport Allocator::GetStorageReport() const
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
```
Infinite666: This is a read only operation, we only need to require for shared lock.

ucm/shared/pool/detail/offset_allocator.cc
```diff
@@ -0,0 +1,453 @@
...
+            SmallFloat::floatToUint((topBinIndex << TOP_BINS_INDEX_SHIFT) | leafBinIndex);
+        assert(freeStorage >= largestFreeRegion);
+    }
+
+    return {freeStorage, largestFreeRegion};
+}
+
+StorageReportFull Allocator::GetStorageReportFull() const
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
```
Infinite666: Same comment as above.

ucm/shared/CMakeLists.txt
hcygit1: Done. `pool/detail` is now added under `pool` and is built only for the `ascend` and `simu` environments. Pool tests are excluded from other runtime environments.

ucm/shared/pool/buffer_region.h
hcygit1: Done. `PoolDetail` has been removed. `BufferRegion` is now defined directly in the `UC` namespace, and all references have been updated.

ucm/shared/pool/variable_buffer_pool.cc
hcygit1: Done. The redundant `GetAllocationSize` validation has been removed from `VariableBufferPool::Free`. `Allocator::Free` now performs the final handle validation under its lock, so the free path only acquires the allocator lock once.

ucm/shared/pool/detail/offset_allocator.cc
hcygit1: The extra call from `VariableBufferPool::Free` has been removed. `GetAllocationSize` remains a public read-only API used by allocator tests, so it is protected by `std::shared_lock` to avoid races with allocation and release operations.

ucm/shared/pool/detail/offset_allocator.cc
hcygit1: Done. `GetStorageReport` now uses `std::shared_lock<std::shared_mutex>` because it only reads allocator state.

ucm/shared/pool/detail/offset_allocator.cc
hcygit1: Done. `GetStorageReportFull` now also uses `std::shared_lock<std::shared_mutex>` for concurrent read access.

Infinite666 (approved):

PR #1148, opened 2026-07-23
[Feat] Add asynchronous buffer delegator executor

pyxyzc:
## Purpose

Introduce asynchronous buffer delegator executor.

## Modification

image

Each task is divided into shards and placed into a global load or dump queue. Available buffer slots determine how many shards may run concurrently. 

Reserved shards are organized as follows:

- `TransferBatch` represents the work acquired in one scheduling round.
- `TransferGroup` contains shards from one task and acts as the backend submission unit.

Load and dump use dedicated workers. Loads are prioritized when reserving new work, while already-running dump operations are not interrupted. Specially for load worker, groups are checked independently so a completed group can scatter its data without waiting for another slower group.

Failures are recorded per group and propagated to the corresponding task as its first error. Once a task fails, its remaining queued shards are completed without additional copying or endpoint I/O. Other tasks in the same batch continue normally.

## Test

image

### DIFF
--- a/ucm/store/delegator/cc/delegator_executor.cc
+++ b/ucm/store/delegator/cc/delegator_executor.cc
@@ -73,32 +73,32 @@ Status BindDevice(std::int32_t deviceId)
 }  // namespace
 
 Expected<std::unique_ptr<Executor>> Executor::Create(std::shared_ptr<StoreV1> backend,
-                                                     std::vector<std::size_t> tensor_sizes,
-                                                     std::int32_t device_id, std::size_t slot_num,
-                                                     std::size_t stream_number)
+                                                     std::vector<std::size_t> tensorSizes,
+                                                     std::int32_t deviceId, std::size_t slotNum,
+                                                     std::size_t streamNumber)
 {
-    if (backend == nullptr || device_id < 0 || slot_num == 0 || stream_number == 0 ||
-        tensor_sizes.empty()) {
+    if (backend == nullptr || deviceId < 0 || slotNum == 0 || streamNumber == 0 ||
+        tensorSizes.empty()) {
         return Status::InvalidParam("invalid delegator executor config");
     }
 
     std::size_t payloadSize = 0;
-    for (const auto size : tensor_sizes) {
+    for (const auto size : tensorSizes) {
         if (size == 0 || size > std::numeric_limits<std::size_t>::max() - payloadSize) {
             return Status::InvalidParam("invalid delegator tensor sizes");
         }
         payloadSize += size;
     }
 
-    auto status = ValidateCurrentDevice(device_id);
+    auto status = ValidateCurrentDevice(deviceId);
     if (status.Failure()) { return status; }
 
     auto executor = std::unique_ptr<Executor>{new (std::nothrow) Executor(
-        std::move(backend), std::move(tensor_sizes), device_id, slot_num, stream_number)};
+        std::move(backend), std::move(tensorSizes), deviceId, slotNum, streamNumber)};
     if (!executor) { return Status::OutOfMemory(); }
 
     try {
-        status = executor->Start(payloadSize, slot_num);
+        status = executor->Start(payloadSize, slotNum);
     } catch (const std::bad_alloc&) {
         return Status::OutOfMemory();
     } catch (const std::system_error& error) {
@@ -109,37 +109,37 @@ Expected<std::unique_ptr<Executor>> Executor::Create(std::shared_ptr<StoreV1> ba
     return executor;
 }
 
-Executor::Executor(std::shared_ptr<StoreV1> backend, std::vector<std::size_t> tensor_sizes,
-                   std::int32_t device_id, std::size_t slot_num, std::size_t stream_number)
+Executor::Executor(std::shared_ptr<StoreV1> backend, std::vector<std::size_t> tensorSizes,
+                   std::int32_t deviceId, std::size_t slotNum, std::size_t streamNumber)
     : backend_{std::move(backend)},
-      tensor_sizes_{std::move(tensor_sizes)},
-      device_id_{device_id},
-      stream_number_{std::min(stream_number, slot_num)}
+      tensorSizes_{std::move(tensorSizes)},
+      deviceId_{deviceId},
+      streamNumber_{std::min(streamNumber, slotNum)}
 {
 }
 
-Status Executor::Start(std::size_t payload_size, std::size_t slot_num)
+Status Executor::Start(std::size_t payloadSize, std::size_t slotNum)
 {
-    auto status = buffer_pool_.Init("delegator_buffer_pool", BufferPool::MemoryType::ASCEND_DEVICE,
-                                    payload_size, slot_num, false, kBufferAlignment);
+    auto status = bufferPool_.Init("delegator_buffer_pool", BufferPool::MemoryType::ASCEND_DEVICE,
+                                   payloadSize, slotNum, false, kBufferAlignment);
     if (status.Failure()) { return status; }
-    slot_num_ = slot_num;
+    slotNum_ = slotNum;
     if (backend_->NeedRegisterKVCaches()) {
         const KVCacheRegistration registration{
-            reinterpret_cast<std::uintptr_t>(buffer_pool_.GetDeviceAddr()),
-            buffer_pool_.GetTotalSize()};
+            reinterpret_cast<std::uintptr_t>(bufferPool_.GetDeviceAddr()),
+            bufferPool_.GetTotalSize()};
         status = backend_->RegisterKVCaches(&registration, 1);
         if (status.Failure()) { return status; }
     }
-    available_slots_ = slot_num;
+    availableSlots_ = slotNum;
 
     std::promise<Status> dumpStarted;
     std::promise<Status> loadStarted;
     auto dumpStartedFuture = dumpStarted.get_future();
     auto loadStartedFuture = loadStarted.get_future();
     try {
-        dump_thread_ = std::thread(&Executor::DumpLoop, this, std::ref(dumpStarted));
-        load_thread_ = std::thread(&Executor::LoadLoop, this, std::ref(loadStarted));
+        dumpThread_ = std::thread(&Executor::DumpLoop, this, std::ref(dumpStarted));
+        loadThread_ = std::thread(&Executor::LoadLoop, this, std::ref(loadStarted));
     } catch (...) {
         Shutdown();
         return Status::Error();
@@ -164,15 +164,15 @@ Expected<Detail::TaskDesc> Executor::MakeBackendTask(const TransferGroup& group)
     try {
         task.reserve(group.shards.size());
         for (const auto& shard : group.shards) {
-            const auto& source = group.task->desc[shard.shard_index];
+            const auto& source = group.task->desc[shard.shardIndex];
             Detail::Shard backendShard;
             backendShard.owner = source.owner;
             backendShard.index = source.index;
-            backendShard.addrs.reserve(tensor_sizes_.size());
+            backendShard.addrs.reserve(tensorSizes_.size());
 
             auto* address = static_cast<std::byte*>(shard.slot.device_addr);
             std::size_t offset = 0;
-            for (const auto size : tensor_sizes_) {
+            for (const auto size : tensorSizes_) {
                 backendShard.addrs.push_back(address + offset);
                 offset += size;
             }
@@ -189,13 +189,13 @@ Status Executor::ValidateTask(const Detail::TaskDesc& task, Operation operation)
     if (operation != Operation::LOAD && operation != Operation::DUMP) {
         return Status::InvalidParam("invalid delegator operation");
     }
-    if (task.empty() || tensor_sizes_.empty()) {
+    if (task.empty() || tensorSizes_.empty()) {
         return Status::InvalidParam("empty delegator task");
     }
     // Verify that the submitted task matches the tensor layout configured for this executor.
     for (std::size_t shardIndex = 0; shardIndex < task.size(); ++shardIndex) {
         const auto& addrs = task[shardIndex].addrs;
-        if (addrs.size() != tensor_sizes_.size()) {
+        if (addrs.size() != tensorSizes_.size()) {
             return Status::InvalidParam("invalid delegator shard({}) address count", shardIndex);
         }
         for (const auto* addr : addrs) {
@@ -210,7 +210,7 @@ Status Executor::ValidateTask(const Detail::TaskDesc& task, Operation operation)
 Status Executor::GatherAsync(const TransferGroup& group, CopyStream& streams)
 {
     for (const auto& shard : group.shards) {
-        const auto& desc = group.task->desc[shard.shard_index];
+        const auto& desc = group.task->desc[shard.shardIndex];
         const auto stream = streams.NextStream();
         if (group.task->desc.prerequisiteHandle != 0) {
             const auto status = streams.WaitEvent(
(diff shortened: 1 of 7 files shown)


ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,561 @@
...
+                    shard.status = submitted.Error();
+                    break;
+                }
+                shard.backend_task = std::move(submitted).Value();
+                shard.backend_pending = true;
+            }
+            if (submittedCount < batch.shards.size()) {
+                for (std::size_t index = submittedCount + 1; index < batch.shards.size(); ++index) {
+                    batch.shards[index].status =
+                        Status::Error("delegator load submission cancelled");
```
pyxyzc: TODO

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,561 @@
...
+    }
+    taskContext->wait_claimed = true;
+
+    taskContext->completed.wait(lock, [&taskContext]() { return taskContext->remaining == 0; });
+    const auto status = taskContext->first_error;
+    tasks_.erase(task);
+    return status;
+}
+
+Expected<Executor::TransferBatch> Executor::AcquireBatch(const std::shared_ptr<TaskContext>& task)
```
Infinite666: Just use `TransferBatch` directly, no need to add `Executor::`.

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,561 @@
...
+            return shutdown_started_ ||
+                   (available_slots_ != 0 &&
+                    (task->operation == Operation::LOAD || unfinished_load_shards_ == 0));
+        });
+        if (shutdown_started_) { return Status::Error(); }
+
+        firstShard = task->next_shard;
+        shardCount = std::min(available_slots_, task->desc.size() - firstShard);
+        available_slots_ -= shardCount;
+        task->next_shard += shardCount;
```
Infinite666: It seems like the meaning of `next_shard` needs more comments to me.

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,561 @@
...
+                    } else {
+                        batch.status = submitted.Error();
+                    }
+                } catch (const std::bad_alloc&) {
+                    batch.status = Status::OutOfMemory();
+                }
+            }
+
+            if (batch.backend_submitted) {
+                const auto waitStatus = backend_->WaitIo(batch.backend_task);
```
Infinite666: Is `WaitIo` a syncronize interface? If so, why should we call `SubmitDumpAll` and then call `WaitIo` immediately instead of using a syncronize submit interface?

ucm/store/delegator/cc/delegator_executor.h
```diff
@@ -0,0 +1,144 @@
...
+    std::int32_t device_id_{-1};
+    std::size_t queue_capacity_{0};
+    std::size_t stream_number_{0};
+
+    std::mutex mutex_;
+    std::condition_variable dispatch_ready_;
+    std::condition_variable slots_ready_;
+    std::condition_variable shutdown_completed_;
+
+    std::deque<std::shared_ptr<TaskContext>> dump_queue_;
```
Infinite666: Take a look at `SpscRingQueue` to consider if that's a better choice.

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,561 @@
...
+Expected<Executor::TransferBatch> Executor::AcquireBatch(const std::shared_ptr<TaskContext>& task)
+{
+    std::size_t firstShard = 0;
+    std::size_t shardCount = 0;
+    {
+        std::unique_lock<std::mutex> lock(mutex_);
+        slots_ready_.wait(lock, [this, &task]() {
+            return shutdown_started_ ||
+                   (available_slots_ != 0 &&
+                    (task->operation == Operation::LOAD || unfinished_load_shards_ == 0));
```
Infinite666: Will this logic keep dump task waiting all the time?

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,561 @@
...
+                    }
+                    shard.backend_pending = false;
+                    --pendingCount;
+                }
+            }
+            const auto syncStatus = streams.SynchronizeAll();
+            const auto shardFailed =
+                std::any_of(batch.shards.begin(), batch.shards.end(),
+                            [](const auto& shard) { return shard.status.Failure(); });
+            if (!shardFailed && batch.status.Success()) { batch.status = syncStatus; }
```
Infinite666: The `batch.status` is always success since nowhere will modify it except for here.

ucm/store/delegator/cc/delegator_executor.h
```diff
@@ -0,0 +1,144 @@
...
+#include "pool/buffer_pool.h"
+
+namespace UC::Delegator {
+
+enum class Operation {
+    LOAD,
+    DUMP,
+};
+
+class Executor {
```
Infinite666, line 53: How does Executor interact with Store?

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,561 @@
...
+            dispatch_ready_.wait(lock, [this]() { return !dump_queue_.empty() || !accepting_; });
+            if (dump_queue_.empty()) {
+                if (!accepting_) { break; }
+                continue;
+            }
+            task = std::move(dump_queue_.front());
+            dump_queue_.pop_front();
+        }
+
+        while (task->next_shard < task->desc.size()) {
```
Infinite666: Maybe we should consider processing multiple tasks to fulfill the slots.

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,561 @@
...
+{
+    std::lock_guard<std::mutex> lock(mutex_);
+    const auto iter = tasks_.find(task);
+    if (iter == tasks_.end()) { return Status::NotFound(); }
+    return bool{iter->second->remaining == 0};
+}
+
+Status Executor::Wait(Detail::TaskHandle task)
+{
+    std::unique_lock<std::mutex> lock(mutex_);
```
Infinite666: The wait logic should not be holding by lock, get the lock and erase the task then you can release the lock immediately.

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,561 @@
...
+            if (result.Success() && shard.status.Failure()) { result = shard.status; }
+        }
+    }
+    for (const auto& shard : batch.shards) {
+        const auto status = buffer_pool_.Free(shard.slot.slot_index);
+        if (result.Success() && status.Failure()) { result = status; }
+    }
+
+    {
+        std::lock_guard<std::mutex> lock(mutex_);
```
Infinite666: The mutex_ lock has so many different responsibilities.

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,561 @@
...
+}
+
+Status Executor::Wait(Detail::TaskHandle task)
+{
+    std::unique_lock<std::mutex> lock(mutex_);
+    const auto iter = tasks_.find(task);
+    if (iter == tasks_.end()) { return Status::NotFound(); }
+
+    const auto taskContext = iter->second;
+    if (taskContext->wait_claimed) {
```
Infinite666: No need for wait_claimed any more if we erase the task before actual waiting.

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: TransferBatch is private :)

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: Done.

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: No longer exists after refactor :)

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: Split the lock into sched_mutex_ and tasks_mutex_ :)

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: Still keep this design for now. Logically it won't be an issue, but we plan to observe the circum under some actual load and then decide whether to change it :)

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: Done. Use a more explicit error logging code style :)

ucm/store/delegator/cc/delegator_executor.h
pyxyzc: Will be introduced in next PR :)

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: Done. Impl in current ver :)

ucm/store/delegator/cc/delegator_executor.h
pyxyzc: SpscRingQueue need fixed size. And most of the time, we use queue/ring under the `sched_mutex_`. It seems that we cannot leverage its advantage :(

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: Consider the condition var wait releases `tasks_mutex_` while the task is incomplete, I think cur impl is acceptable :)
Another consider is assume now we erase task from `tasks_` in one Wait, and then we have Check on this task, we will get a NotFound. Sounds really weird :(

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: Make sense, will be considered in the next PR :)

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,697 @@
...
+        auto shard = std::move(queue.front());
+        queue.pop_front();
+        DiscardShard(shard, status, ShardStage::QUEUED);
+    }
+}
+
+void Executor::DumpLoop(std::promise<Status>& started)
+{
+    // Init
+    CopyStream streams;
```
Fengli5355, line 612: Should we first wait the event of prerequisiteHandle before Dump?

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,697 @@
...
+            for (auto& group : batch.groups) {
+                if (!group.transfer_pending) { continue; }
+
+                auto completed = endpoint_->Check(group.transfer_task);
+                if (!completed) {
+                    group.error = completed.Error();
+                } else if (!completed.Value()) {
+                    continue;
+                } else {
+                    const auto scatterStatus = ScatterAsync(group, streams);
```
Fengli5355: if completed == true, i.e. task.remaining == 0, is there a case that some batch failed but *finished*?

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: We set a strict rule to let dump shard/dump task is admitted only when no load shard is queued or in flight :)

Fengli5355: I mean before kv cache in the infra is fully computed, will Dump actually dumps dirty data?

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: In the cur impl, we consider `completed` to mean the cur task is done by delegator :) Whether it is a normal completion depends on the recorded error.

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: > I mean before kv cache in the infra is fully computed, will Dump actually dumps dirty data?

Make sense. Will be introduced in next PR :)

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,697 @@
...
+        const auto rollbackPublishedShards = [&queue, previousQueueSize]() noexcept {
+            while (queue.size() != previousQueueSize) { queue.pop_back(); }
+        };
+        try {
+            for (std::size_t shardIndex = 0; shardIndex < count; ++shardIndex) {
+                queue.push_back(QueuedShardContext{taskContext, shardIndex});
+            }
+            {
+                std::lock_guard<std::mutex> tasksLock(tasks_mutex_);
+                if (!tasks_.emplace(handle, taskContext).second) {
```
Infinite666, line 331: You can try to insert for tasks first before adding to queue, so that this abnormal branch doesn't need to rollback.

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,697 @@
...
+        queued += count;
+        AssertSchedulerInvariantsLocked();
+    }
+    slots_ready_.notify_all();
+    return Detail::TaskHandle{handle};
+}
+
+Expected<bool> Executor::Check(Detail::TaskHandle task)
+{
+    std::lock_guard<std::mutex> lock(tasks_mutex_);
```
Infinite666: The operation of read only needs shared lock.

ucm/store/delegator/cc/delegator_executor.cc
Infinite666: I think it's unacceptable to hold the lock while waiting, during the waiting no other operation of tasks can do.

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,697 @@
...
+                // Wait only for the first shard; do not wait to fill a partial batch.
+                if (reservedShards.empty()) {
+                    slots_ready_.wait(
+                        lock, [this, &canReserve]() { return shutdown_started_ || canReserve(); });
+                }
+                if (shutdown_started_) {
+                    if (reservedShards.empty()) { return Status::Error(); }
+                    break;
+                }
+                if (!canReserve()) { break; }
```
Infinite666: Looks like this judgement is useless since you have already hold the mutex.

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,697 @@
...
+            ++available_slots_;
+        }
+        --outstanding_shards_;
+        AssertSchedulerInvariantsLocked();
+    }
+
+    slots_ready_.notify_all();
+}
+
+void Executor::DrainQueue(std::deque<QueuedShardContext>& queue)
```
Infinite666, line 599: Why don't we need to aquire mutex for this function?

ucm/store/delegator/cc/delegator_executor.cc
```diff
@@ -0,0 +1,697 @@
...
+    : endpoint_{std::move(endpoint)},
+      tensor_sizes_{std::move(tensor_sizes)},
+      device_id_{device_id},
+      stream_number_{std::min(stream_number, slot_num)}
+{
+}
+
+Status Executor::Start(std::size_t payload_size, std::size_t slot_num)
+{
+    auto status = buffer_pool_.Init("delegator_buffer_pool", BufferPool::MemoryType::ASCEND_DEVICE,
```
Infinite666: Take a look at all the naming cleancode, the format in ucm is like aaaBBB.

pyxyzc: Done.

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: DrainQueue is called after all worker threads have stopped. At this point, Shutdown is the only consumer of both queues :)
Note: DiscardShard used in DrainQueue  also acquires `schedMutex_` :)

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: Done.

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: Done.

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: Done.

ucm/store/delegator/cc/delegator_executor.cc
pyxyzc: Consider if we first insert and it is still possible that queue.push_back() throw std::bad_alloc, which means we have to tasks_.erase(handle);, an even more expensive op :(
By the way, rollback is just pop so I think it's ok :)

Infinite666 (approved):

PR #1150, opened 2026-07-23
[Feat] Add DramPool completion poller

yumingyue624:
## Purpose
Add asynchronous completion polling to DramPool so submitted DUMP and LOAD transfers are settled only after transport reaches a terminal state, and response buffers remain valid until the corresponding response transfer completes.

## Modifications
1. Add CompletionPoller to poll data and response transfers, settle StoreEnd, LoadEnd, and Delete paths, and manage response flag-buffer slot lifetime.
2. Disconnect timed-out peer transfers without fabricating terminal transport status, retry transient disconnect failures, and let shutdown proceed without releasing memory before TransportManager shutdown.
3. Integrate CompletionPoller into DramPool server startup and shutdown, share runtime constants and completion record types, and remove redundant state.
4. Build DramPool with the upstream P2P transport targets, keep DramStore client and DramPool server test links explicit, use GCC/Clang CMake settings, and collect DramPool production and test sources with the .cc suffix.
5. Add CompletionPoller coverage for lifecycle, metadata settlement, response buffers, timeout, disconnect, shutdown, and error paths, plus YAML validation coverage for poller depth exceeding available flag-buffer slots.

## Test
```
[==========] Running 136 tests from 13 test suites.
[  PASSED  ] 136 tests.
```


ucm/store/dram/cc/drampool/drampool_server.cc
```diff
@@ -403,24 +429,11 @@ void DramPoolServer::StopGCThread()
 void DramPoolServer::StopTransportService()
 {
     if (transportManager_) {
-        UnregisterBufferPools();
         const auto status = transportManager_->Shutdown();
         if (status.Failure()) { UC_ERROR_UNLIMITED("DramPool TransportManager shutdown failed"); }
     }
 }
 
-void DramPoolServer::UnregisterBufferPools()
```
Infinite666, line 412: Why do we delete the unregister logic?

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,314 @@
...
+    const auto submitStatus = runtime_.transport.ExecuteAsync(operation, handle);
+    if (submitStatus.Failure() || handle == transport::kInvalidTransferHandle) {
+        const auto freeStatus = runtime_.flagBufferPool.Free(record.resp_buffer.slot_index);
+        if (freeStatus.Failure()) {
+            UC_ERROR(
+                "CompletionPoller release flag buffer after submit failure failed, slot={}, "
+                "error={}",
+                record.resp_buffer.slot_index, freeStatus);
+        }
+        record.resp_buffer = {};
```
yuanzhg078: Consider extracting a helper to centralize the Free(slot) + resp_buffer = {} cleanup across these failure branches

yumingyue624: Done. Centralized flag-buffer release, descriptor reset, and failure logging in ReleaseResponseBuffer.

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+#include "core/transport_manager.h"
+#include "drampool_config.h"
+#include "logger/logger.h"
+#include "metadata.h"
+
+namespace UC::DramPool {
+namespace {
+
+void ReleaseResponseBuffer(BufferPool& flagBufferPool, CompletionRecord& record,
+                           const char* context)
```
Infinite666: Looks like there's no need to pass context here, the error branch will return error status and it will be printed at the outside function.

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+        }
+        if (newPendingCount == 0 && pending_.empty()) {
+            std::this_thread::sleep_for(kThreadIdleSleepDuration);
+        }
+    }
+}
+
+void CompletionPoller::SetDisconnectAllTransfers() noexcept
+{
+    disconnectAllTransfers_.store(true, std::memory_order_release);
```
Infinite666: No need to extract for a function if there will not be any more logics in the future.

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+                // Keep unfinished records and their buffers alive. DramPoolServer shuts down
+                // transport before destroying the poller and memory pools.
+                UC_ERROR_UNLIMITED(
+                    "CompletionPoller stopped waiting because an in-flight peer could not be "
+                    "disconnected");
+                break;
+            }
+            if (newPendingCount == 0 && pending_.empty()) { break; }
+        }
+        if (newPendingCount == 0 && pending_.empty()) {
```
Infinite666: If `newPendingCount > 0` and `pending_.empty()`, it means this loop get some task from `completionQueue` and all of them are finished? If so, maybe `newPendingCount` is useless, `pending_.empty()` is enough. But btw, should we also check the remaining size of `completionQueue`?

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+}
+
+bool CompletionPoller::OperationTimedOut(const CompletionRecord& record, std::uint64_t nowMs) const
+{
+    if (nowMs < record.submit_ms) { return false; }
+    return nowMs - record.submit_ms >= g_config.opTimeoutMs;
+}
+
+void CompletionPoller::DisconnectPeer(CompletionRecord& record, TransportHandle handle,
+                                      const char* transferType)
```
Infinite666: Same as `ReleaseResponseBuffer`, do not pass `transferType` for the inner error log, just print the error log in the branch calling `DisconnectPeer` will be more clear.

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+        // GetStatus removes failed handles, so an API failure is also terminal.
+        UC_ERROR("CompletionPoller data GetStatus failed, handle={}", record.data_handle);
+        SettleDataTransfer(record, transport::TransferStatus::Failed);
+        record.data_handle = transport::kInvalidTransferHandle;
+        record.stage = CompletionStage::SubmitResponse;
+        return true;
+    }
+
+    if (transportStatus == transport::TransferStatus::Waiting) {
+        if (!record.disconnect_attempted &&
```
Infinite666: Will a record be disconnected but still process by `PollDataTransfer` and get to here?

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+            (disconnectAllTransfers_.load(std::memory_order_acquire) ||
+             OperationTimedOut(record, SteadyNowMs()))) {
+            DisconnectPeer(record, record.data_handle, "data");
```
Infinite666: If the record is already disconnected, why don't we set the stage to `SubmitResponse` and return true?

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+            const auto releaseStatus = runtime_.metadata.LoadEnd(item.key);
+            if (releaseStatus.Failure()) {
+                UC_ERROR("CompletionPoller LoadEnd failed, handle={}, error={}", record.data_handle,
+                         releaseStatus);
+            } else if (terminalStatus == transport::TransferStatus::Completed) {
+                result = DumpLoadResult::Ok;
+            }
+        }
+
+        if (item.index_in_request >= record.results.size()) {
```
Infinite666: In which scenario, the size of item can be larger than results?

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+        ReleaseResponseBuffer(runtime_.flagBufferPool, record, "response packing");
+        return protocolStatus;
+    }
+
+    transport::Operation operation;
+    operation.opcode = transport::Opcode::Write;
+    operation.direct = transport::OperationDirect::RemoteDeviceHost;
+    operation.target_manager = record.peer_one_sided_id;
+    operation.ops.emplace_back(
+        transport::Segment{record.resp_buffer.local_addr, record.response_addr, len});
```
Infinite666: The naming of `resp_buffer` and `response_addr` is too similar.

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+    // A terminal GetStatus releases the data handle before business state is settled.
+    SettleDataTransfer(record, transportStatus);
+    record.data_handle = transport::kInvalidTransferHandle;
+    record.stage = CompletionStage::SubmitResponse;
+    return true;
+}
+
+Status CompletionPoller::SubmitResponse(CompletionRecord& record)
+{
+    if (record.peer_one_sided_id.empty()) {
```
Infinite666: What kind of record will come to this branch?

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+        return Status::InvalidParam("response peer_one_sided_id is empty");
+    }
+
+    const auto packedSize =
+        runtime_.protocol.GetPackedResponseSize(record.opcode, record.results.size());
+    if (packedSize == 0 || packedSize > std::numeric_limits<std::uint32_t>::max()) {
+        return Status::InvalidParam("invalid packed response size");
+    }
+    const auto len = static_cast<std::uint32_t>(packedSize);
+    if (record.resp_buffer.local_addr != nullptr) {
```
Infinite666: What kind of record will come to this branch?

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+
+Status CompletionPoller::SubmitResponse(CompletionRecord& record)
+{
+    if (record.peer_one_sided_id.empty()) {
+        return Status::InvalidParam("response peer_one_sided_id is empty");
+    }
+
+    const auto packedSize =
+        runtime_.protocol.GetPackedResponseSize(record.opcode, record.results.size());
+    if (packedSize == 0 || packedSize > std::numeric_limits<std::uint32_t>::max()) {
```
Infinite666: Do we need to write error code back to dram store in this abnormal branch?

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+        return Status::InvalidParam("invalid packed response size");
+    }
+    const auto len = static_cast<std::uint32_t>(packedSize);
+    if (record.resp_buffer.local_addr != nullptr) {
+        return Status::InvalidParam("response flag buffer slot is already allocated");
+    }
+    auto allocateStatus = runtime_.flagBufferPool.Allocate(record.resp_buffer);
+    if (allocateStatus.Failure()) { return allocateStatus; }
```
Infinite666: We should retry if the failure is because of no space, but now we will erase the record because of the  code here:
```
                const auto status = SubmitResponse(*iter);
                if (status.Failure()) {
                    UC_ERROR("CompletionPoller SubmitResponse failed, opcode={}, error={}",
                             static_cast<int>(iter->opcode), status);
                    iter = pending_.erase(iter);
                } else {
                    ++iter;
                }
```

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+    if (submitStatus.Failure() || handle == transport::kInvalidTransferHandle) {
+        ReleaseResponseBuffer(runtime_.flagBufferPool, record, "response submission");
+        if (submitStatus.Failure()) { return submitStatus; }
+        return Status::Error("ExecuteAsync response returned an invalid handle");
+    }
+
+    // ExecuteAsync now owns access to the bytes; retain the buffer until handle terminal.
+    record.response_handle = handle;
+    record.submit_ms = SteadyNowMs();
+    record.disconnect_attempted = false;
```
Infinite666: Why do we need to set `disconnect_attempted` to false here? In which scenario the `disconnect_attempted` is true and come to here?

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+            (disconnectAllTransfers_.load(std::memory_order_acquire) ||
+             OperationTimedOut(record, SteadyNowMs()))) {
+            DisconnectPeer(record, record.response_handle, "response");
+        }
+        return false;
+    } else if (transportStatus != transport::TransferStatus::Completed) {
+        UC_ERROR("CompletionPoller response transfer failed, handle={}", record.response_handle);
+    }
+
+    const auto releasedSlot = record.resp_buffer.slot_index;
```
Infinite666: Same as `ReleaseResponseBuffer`.

ucm/store/dram/cc/drampool/drampool_server.h
```diff
@@ -114,7 +114,7 @@ class DramPoolServer {
     std::unique_ptr<transport::TransportManager> transportManager_;
     std::unique_ptr<transport::TcpMessageChannel> tcpMessageChannel_;
     std::unique_ptr<BufferManager> bufferManager_;
-    std::vector<transport::MemoryHandle> bufferPoolMemoryHandles_;
+    std::unique_ptr<UC::BufferPool> flagBufferPool_;
     std::unique_ptr<UC::DramPool::MetadataManager> metadataManager_;
```
Infinite666: `MetadataManager` is under the same namespace of the file.

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -0,0 +1,310 @@
...
+    }
+}
+
+}  // namespace
+
+void CompletionPoller::Run(const std::atomic_bool& stop)
+{
+    while (true) {
+        const bool stopRequested = stop.load(std::memory_order_acquire);
+        if (stopRequested) { SetDisconnectAllTransfers(); }
```
Infinite666: If `stopRequested`, do we still need to call `FillPendingWindow` to fetch new tasks?

ucm/store/dram/cc/drampool/drampool_server.cc
yumingyue624: TransportManager::Shutdown() internally invokes each transport's Shutdown(), which already iterates over all registered memory regions and calls UnregisterMemory on them (see hixl_transport.cpp:173-179). Therefore, a separate UnregisterBufferPools call is redundant.

ucm/store/dram/cc/drampool/drampool_server.h
yumingyue624: Done. Changed UC::DramPool::MetadataManager to MetadataManager.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Done. Removed SetDisconnectAllTransfers() and inlined the assignment. Also demoted disconnectAllTransfers_ from std::atomic_bool to bool since it's only accessed within the poller thread.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Done. Removed context parameter from ReleaseResponseBuffer.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Done. Replaced the inline buffer release with ReleaseResponseBuffer helper to avoid duplication.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Done.  Removed transferType parameter from DisconnectPeer.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Removed. This was defensive programming, but item.index_in_request is always assigned when the record is created and guaranteed to be within record.results.size() bounds.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Removed. This was defensive programming, but peer_one_sided_id is already validated in TaskWorker before the record reaches SubmitResponse, so the check is unnecessary.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Removed. This was defensive programming. A record only enters SubmitResponse once — after buffer allocation, it either proceeds to PollResponseTransfer (where the buffer is released) or fails and gets erased from pending. There's no path where a record reaches this point with resp_buffer.local_addr already set.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Done. Changed SubmitResponse to return bool (same semantics as PollDataTransfer/PollResponseTransfer): true means the record is done and can be erased, false means it should remain pending. When Allocate returns NoSpace, it returns false so the record stays in pending and retries next round.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Done. Renamed resp_buffer → local_resp_slot and response_addr → remote_resp_addr to clearly distinguish the local buffer slot from the remote peer address.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Removed. This was defensive programming. packedSize is guaranteed to be valid by the DramStore process, and TaskWorker also validates the response size against flagBufferSlotSizeBytes before the record reaches here. The check has been removed.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Done. After calling DisconnectPeer, the record is now immediately settled as Failed and returns true, so it will no longer re-enter PollDataTransfer with Waiting status.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Done. After DisconnectPeer, we now directly call SettleDataTransfer(Failed) and return true to advance the record to SubmitResponse. The same approach applies to PollResponseTransfer — disconnect and immediately return true.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Removed. disconnect_attempted has been removed entirely — after DisconnectPeer, the record is now immediately settled as failed, so there's no need for this flag.

ucm/store/dram/cc/drampool/completion_poller.cc
yumingyue624: Done. Removed newPendingCount and now use pending_.empty() directly. FillPendingWindow() keeps popping until the window is full or TryPop() observes an empty queue, so a separate queue-size check would be redundant and racy; during shutdown, TaskWorker is joined before stopping the poller, so no new completions can be enqueued.

Infinite666 (approved):

Wwwzff (approved):

PR #1159, opened 2026-07-25
[Refactor] Adjust register and query

Fengli5355:
## Purpose
This PR simplifies ASU registered-memory management and improves multi-transport query execution. It removes duplicated ownership and result types, makes cleanup best-effort, and changes Client queries from serial execution to asynchronous fan-out followed by a unified wait phase.

## Modifications 
1. Commit e67f870 replaces ownedRegisteredRegionHandles_with a boolean ownsRegisteredRegionHandles_;
2. Commit 0c44080: Unregister now only logs error when fails;
3. Commit 5979b7b reuses RegisteredMemory in the Client cache;
4. Commit f759cb5 consolidates registration result types;
5. Commit 1bb259e fans out queries before waiting in Client.

## Test
To be examined.

### DIFF
--- a/ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
+++ b/ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
@@ -362,19 +362,9 @@ Status AsuTransportImpl::RegisterRegions(const std::vector<MemoryRegion>& region
     }
 
     std::vector<MRHandle> mrHandles;
-    auto trackReturnedHandles = [&]() {
-        for (std::size_t index = 0; index < mrHandles.size(); ++index) {
-            RegisteredMemory registeredMemory;
-            if (index < regions.size()) { registeredMemory.region = regions[index]; }
-            registeredMemory.handle = mrHandles[index];
-            registeredRegions_[mrHandles[index]] = registeredMemory;
-        }
-        if (!mrHandles.empty()) { ownsRegisteredRegionHandles_ = true; }
-    };
 
     auto status = transProvider_->RegisterMemory(registerDescs, mrHandles);
     if (!status.ok()) {
-        trackReturnedHandles();
         const auto cleanupStatus = UnregisterOwnedRegionHandles(mrHandles);
         return Status::Error(StatusCode::PARTIAL_FAILED,
                              cleanupStatus.ok()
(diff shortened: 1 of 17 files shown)


ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -355,16 +362,20 @@ Status AsuTransportImpl::RegisterRegions(const std::vector<MemoryRegion>& region
...
+            registeredMemory.handle = mrHandles[index];
+            registeredRegions_[mrHandles[index]] = registeredMemory;
+        }
+        if (!mrHandles.empty()) { ownsRegisteredRegionHandles_ = true; }
+    };
+
     auto status = transProvider_->RegisterMemory(registerDescs, mrHandles);
     if (!status.ok()) {
-        ownedRegisteredRegionHandles_.insert(mrHandles.begin(), mrHandles.end());
+        trackReturnedHandles();
```
Infinite666: Looks like we don't need to track handles in this abnormal branch.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -373,9 +384,8 @@ Status AsuTransportImpl::RegisterRegions(const std::vector<MemoryRegion>& region
     if (mrHandles.size() != regions.size()) {
         status = Status::Error(StatusCode::INTERNAL_ERROR,
                                "register result count does not match region count");
-        ownedRegisteredRegionHandles_.insert(mrHandles.begin(), mrHandles.end());
+        trackReturnedHandles();
```
Infinite666: Same like above.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -387,12 +397,8 @@ Status AsuTransportImpl::RegisterRegions(const std::vector<MemoryRegion>& region
         status = transProvider_->GetMemTokenId(mrHandles[index], tokenIds[index]);
         if (status.ok()) { continue; }
 
-        ownedRegisteredRegionHandles_.insert(mrHandles.begin(), mrHandles.end());
+        trackReturnedHandles();
```
Infinite666: Although the `trackReturnedHandles` is idempotent, calling `trackReturnedHandles` in this for loop will cause duplicate handle insertion. Maybe recording the handles after making sure every region are registered and got token id will be a better choice.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
Fengli5355: Agree. The lambda function trackReturnedHandles is removed.

Infinite666 (approved):

PR #1164, opened 2026-07-27
[Feat] Coordinate transport connection lifecycle

ayaka836:
### Update `TransportManager::Connect/Disconnect` semantics

`TransportManager` now synchronizes connection state with the peer through the TCP control plane.

Calling `A.Connect(B)` establishes the local connection on A and sends a control request to B to establish the corresponding connection. The disconnect flow follows the same mechanism. Both operations wait for the peer acknowledgment before returning.

### Extend the TCP control protocol

The control plane now supports:

- Transport metadata exchange
- Connection synchronization
- Disconnection synchronization

### Add HIXL role initialization

The following HIXL roles are supported:

- `Client`: Initiates connections without listening on a HIXL port.
- `Server`: Listens for and accepts connections, but cannot initiate data transfers.
- `Bidirectional`: Supports listening, initiating connections, and data transfers.

### Reorganize the P2P Transport directory structure

The `one_sided` and `two_sided` directory hierarchy has been replaced with responsibility-based modules:

- `core`
- `control`
- `common`
- `channels/tcp`
- `protocols/hixl`


ucm/transport/p2p/src/protocols/hixl/hixl_transport.cpp
```diff
@@ -362,9 +377,19 @@ Status HixlTransport::Connect(const ManagerID& manager_id)
     if (peer.connected) { return Status::OK(); }
     if (peer.local_index >= instances_.size()) { return Status::Error(); }
 
+    if ((role_ == HixlRole::Server && peer.role == HixlRole::Server) ||
```
Infinite666: Why can't we simplify the judgement like it in `DisconnectRoute`?

ucm/transport/p2p/include/control/control_protocol.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include <cstdint>
+#include <optional>
+#include "core/transport.h"
+
+namespace transport {
+
+enum class ControlOperation : uint32_t {
+    ExchangeMetadata = 0,
```
Infinite666, line 10: In our logic, is `ExchangeMetadata` more like a Connect, and `Connect` is more like a Reconnect? If so, we can rename them to make our code more clear.

ayaka836: ExchangeMetadata handles metadata exchange through the TCP connection. The connection created by Connect is the actual transport channel used for data transfer.

ucm/transport/p2p/src/protocols/hixl/hixl_transport.cpp
ayaka836: Fixed.

ucm/transport/p2p/include/control/control_protocol.h
Infinite666: Just to make sure if I get it, so we need to send a request of `ExchangeMetadata` and another request of `Connect` to create connection?

ayaka836: Yes.

Infinite666 (approved):

PR #1166, opened 2026-07-28
[FEAT] Add dramstore transport and reply poller

nrj868:
## Purpose
Add trans executor & reply poller for dramstore:
- TransportExecutor — a sharded, bounded-queue worker pool that drives the backend and publishes completion events, with separate command/fence admission budgets per node.
- ReplyService — the single owner of reply-slot allocation and polling; decodes DramPool responses and publishes ReplyObserved events, releasing a slot only at a safe terminal point.

## Test
tested on mocked ibverbs

### DIFF
--- a/ucm/shared/CMakeLists.txt
+++ b/ucm/shared/CMakeLists.txt
@@ -1,7 +1,7 @@
 add_subdirectory(vendor)
 add_subdirectory(infra)
 add_subdirectory(trans)
-if(RUNTIME_ENVIRONMENT STREQUAL "ascend")
+if(RUNTIME_ENVIRONMENT STREQUAL "ascend" OR RUNTIME_ENVIRONMENT STREQUAL "simu")
     add_subdirectory(pool)
 endif()
 add_subdirectory(metrics)
--- a/ucm/store/dram/cc/transport_manager_backend.cc
+++ b/ucm/store/dram/cc/transport_manager_backend.cc
@@ -1,0 +1,154 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2026 Huawei Technologies Co., Ltd. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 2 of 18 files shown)


ucm/shared/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 add_subdirectory(vendor)
 add_subdirectory(infra)
 add_subdirectory(trans)
-if(RUNTIME_ENVIRONMENT STREQUAL "ascend")
+if(RUNTIME_ENVIRONMENT STREQUAL "ascend" OR RUNTIME_ENVIRONMENT STREQUAL "simu")
```
Infinite666, line 4: Please also modify the CMake of buffer pool's tests.

ucm/store/dram/cc/transport_manager_backend.cc
```diff
@@ -0,0 +1,175 @@
...
+#include <utility>
+#include "core/transport_init_attrs.h"
+#include "core/transport_manager.h"
+#include "transport_executor.h"
+#include "two_sided/tcp/tcp_message_channel.h"
+
+namespace UC::Dram {
+namespace {
+
+class TransportManagerBackend final : public ITransportBackend {
```
Infinite666: Does this meet the standard of cleancode to put the definition of class here instead of in header file?

ucm/store/dram/cc/transport_manager_backend.cc
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    Status Fence(const ::UC::Dram::FenceEpoch& command) noexcept override
+    {
+        const auto found = nodes_.find(command.nodeId);
+        if (found == nodes_.end()) { return Status::InvalidParam("unknown DramPool node"); }
+        try {
+            // The transport Manager contract guarantees that successful Disconnect
+            // synchronously revokes old-connection access to local registered memory.
+            return manager_.Disconnect(transport::TransportProtocol::Hixl,
```
Infinite666: What will this disconnect process actually do? @ayaka836 said only dram pool service can do the disconnection. And, 1. do we need to mark this node as down? 2. when will we try to reconnect?

ucm/store/dram/cc/transport_manager_backend.cc
```diff
@@ -0,0 +1,175 @@
...
+                control_.Send(endpoint, command.payload.data(), command.payload.size());
+            if (status.Success()) { return TransmitCompleted{command.token, Status::OK()}; }
+            return TransmitCompleted{command.token, status};
+        } catch (...) {
+            return TransmitCompleted{command.token,
+                                     Status::Error("TCP transmit threw an exception")};
+        }
+    }
+
+    Status Connect(const ::UC::Dram::Connect& command) noexcept override
```
Infinite666: Looks like it's more reasonable to insert for node after connected?

ucm/store/dram/cc/transport_manager_backend.cc
nrj868: @ayaka836 will provide the `Connect` and `Disconnect` implementations for `dramstore` to invoke.
When a disconnection occurs, we will mark the affected lane as down and attempt to reconnect at a fixed interval.

ucm/shared/CMakeLists.txt
nrj868: Will fix in another pr

ucm/store/dram/cc/transport_manager_backend.cc
nrj868: Fixed.

ucm/store/dram/cc/transport_manager_backend.cc
nrj868: Node will be inserted via a `Command` by upper-layer component, which is responsible for manageing view/dht

Infinite666 (approved):

PR #1169, opened 2026-07-28
Fix model path issue in ucm_connector.py. 

luomoduchu:
## Modifications 
The original configuration caused different cache hashes for the same model when loaded from different paths. After the fix, the specific path is decoupled; as long as the final model name in the path is identical, the hash is considered consistent, and the cache is stored in the same location.

## Test
test on cuda Qwen2.5 succeed

### DIFF
--- a/ucm/integration/vllm/ucm_connector.py
+++ b/ucm/integration/vllm/ucm_connector.py
@@ -835,7 +835,7 @@ class RequestHasher:
         sparse_sfa_c8 = bool(additional_config.get("enable_sparse_sfa_c8", False))
         sparse_li_c8 = bool(additional_config.get("enable_sparse_li_c8", False))
         sparse_c8_info = f":sfa_c8={int(sparse_sfa_c8)}:li_c8={int(sparse_li_c8)}"
-        model_name=vllm_config.model_config.model.split('/')[-1]
+        model_name = vllm_config.model_config.model.split("/")[-1]
         meta = (
             f"{model_name}:"
             f"{vllm_config.parallel_config.tensor_parallel_size}:"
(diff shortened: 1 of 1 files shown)


qyh111 (dismissed):

ucm/integration/vllm/ucm_connector.py
```diff
@@ -876,8 +876,9 @@ def __init__(self, vllm_config, rank_id):
         sparse_sfa_c8 = bool(additional_config.get("enable_sparse_sfa_c8", False))
         sparse_li_c8 = bool(additional_config.get("enable_sparse_li_c8", False))
         sparse_c8_info = f":sfa_c8={int(sparse_sfa_c8)}:li_c8={int(sparse_li_c8)}"
+        model_name = vllm_config.model_config.model.split("/")[-1]
```
qyh111: 一个具体可改进点：vllm-ascend 用的是 model_config.model.rstrip("/").split("/")[-1]，而 PR #1169 只用了 model.split("/")[-1]。带尾斜杠的路径（如 Qwen/Qwen2.5/）会被 PR 的写法切成空串 ""，导致所有这类模型塌缩到同一个空 model_name；而 vllm-ascend 的 rstrip("/") 能正确得到 Qwen2.5。建议把 PR 改成对齐：
model_name = vllm_config.model_config.model.rstrip("/").split("/")[-1]
这样既与 Ascend 路径行为一致，也修复了尾斜杠的边界 bug。

luomoduchu: now use model_name = vllm_config.model_config.model.rstrip("/").split("/")[-1]

qyh111 (approved):

PR #1171, opened 2026-07-29
[Feat] Trigger connection recovery after maxErrorCount consecutive failures

Fengli5355:
## Purpose
Trigger connection recovery after maxErrorCount consecutive failures

## Modifications 
1. Add ReportSuccess to reset error count.
2. Add configuration of maxErrorCount in .yaml file.

### DIFF
--- a/ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
+++ b/ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
@@ -139,6 +139,9 @@ Status AsuTransportImpl::Init(const TransportConfig& config)
         return Status::OK();
     }
     config_ = config;
+    if (config_.maxErrorCount == 0) {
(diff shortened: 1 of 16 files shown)


ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -663,6 +667,9 @@ void AsuTransportImpl::PollTaskCompletions(const TransportTaskContextPtr& ctx)
 
                 std::fill(subBatchContext.entryStatus.begin(), subBatchContext.entryStatus.end(),
                           timeoutStatus);
+                if (subBatchContext.channel != nullptr) {
```
yuanzhg078: channel != nullptr is already contained in the ReportFailure.

Fengli5355: Agree. If-condition is removed.

Infinite666 (approved):

PR #1173, opened 2026-07-29
[Refactor] Improve ASU transport connection handling, task completion, and shutdown lifecycle

yuanzhg078:
## Purpose
Harden ASU transport task lifecycle and shutdown behavior so task cancellation, sub-batch completion, concurrent submission, and in-flight task draining are handled consistently and safely.
image

## Modifications

- Unify task terminal state as COMPLETED and represent cancellation through the final task status.
- Track remaining sub-batches with idempotent completion and reliable task finalization.
- Serialize task submission against shutdown and reject submissions after the acceptance gate closes.
- Stop the request worker first while allowing the completion worker to process in-flight tasks for a bounded timeout.
- Simplify connection handling according to the documented provider contract.
- Add lifecycle, shutdown callback, in-flight completion, and sub-batch completion unit-test coverage.
- Add an ASU transport porting guide documenting the required behavior and shutdown ordering.

## Test
Unit tests are all passed.
A UCM offline inference is passed. (MLA/GQA+TP2+Layerwise(ture/false))

### DIFF
--- a/ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
+++ b/ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
@@ -118,7 +118,7 @@ void AsuTransportImpl::CompleteSubBatch(TransportTaskContext& task,
     ReleaseSubBatchResources(subBatchContext);
     subBatchContext.state = TransportSubBatchState::COMPLETED;
     subBatchContext.status = status;
-    ++task.completedSubBatchCount;
+    --task.remainingSubBatchCount;
 }
 
 AsuTransportImpl::~AsuTransportImpl() { Shutdown(); }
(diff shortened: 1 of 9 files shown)


ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -241,6 +243,31 @@ Status AsuTransportImpl::Init(const TransportConfig& config)
 Status AsuTransportImpl::Shutdown()
 {
     Status finalStatus = Status::OK();
+    {
+        std::lock_guard<std::mutex> lock(producerMu_);
+        acceptingTasks_.store(false, std::memory_order_release);
```
Infinite666: Looks like the changes of `acceptingTasks_` and `stopWorker_` are always at the same time, do we still need two flag attributes?

Infinite666 (approved):

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
yuanzhg078: Done. Only use stopWorker_  now.

PR #1174, opened 2026-07-29
[Fix] Avoid GCC 12 -Werror=nonnull false positive in kv_protocol_test

yumingyue624:
## Purpose
GCC 12 optimizer incorrectly infers a null source pointer in the std::copy -> __builtin_memmove inlined path when vector<uint8_t> is assigned via initializer_list. This triggers -Werror=nonnull and breaks the build.

## Modifications
Add set_source_files_properties in ucm/store/test/CMakeLists.txt to suppress -Wno-nonnull for kv_protocol_test.cc, scoped to GCC only.

## Test
[==========] Running 40 tests from 1 test suite.
[  PASSED  ] 40 tests.

### DIFF
--- a/ucm/store/test/CMakeLists.txt
+++ b/ucm/store/test/CMakeLists.txt
@@ -13,6 +13,10 @@ if(BUILD_UNIT_TESTS)
         list(FILTER UCMSTORE_TEST_SOURCE_FILES EXCLUDE REGEX ".*[/\\\\]dram[/\\\\].*")
     endif()
     add_executable(ucmstore.test ${UCMSTORE_TEST_SOURCE_FILES})
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "12" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "13")
+        set_source_files_properties(case/dram/kv_protocol_test.cc
+            PROPERTIES COMPILE_OPTIONS "-Wno-nonnull")
+    endif()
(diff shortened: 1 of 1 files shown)


ucm/store/test/case/dram/kv_protocol_test.cc
```diff
@@ -271,7 +271,8 @@ TEST_F(KvProtocolTest, ServerRoundTripDumpLoad)
 
     // server packs response, client unpacks it
     KvResponse resp;
-    resp.results = {0x0};  // batch_size == 1 errcode
+    resp.results.clear();
```
nrj868: clear() is redundant in here

ucm/store/test/case/dram/kv_protocol_test.cc
```diff
@@ -617,7 +623,9 @@ TEST_F(KvProtocolTest, UnpackResponseReturnsRetryWithoutChangingResultsWhilePend
 {
     const std::uint8_t pending[] = {static_cast<std::uint8_t>(ResponseStatus::Pending), 0xFF};
     KvResponse response;
-    response.results = {7, 8};
+    response.results.clear();
+    response.results.push_back(7);
```
nrj868: Just using `response.results = {7, 8, xxx};` would be much more readable

ucm/store/test/case/dram/kv_protocol_test.cc
yumingyue624: Fixed by suppressing the -Wno-nonnull warning via CMake for kv_protocol_test.cc only, scoped to GCC 12.x where this false positive is a known regression (fixed in GCC 13+).

Infinite666 (approved):

PR #1182, opened 2026-08-03
[Feat] Refactor AsuClient task APIs

Fengli5355:
## Purpose
Refactor task management of AsuClient & AsuTransport.

## Modifications 
1. Commit 4624cd3b unifies client and transport task model;
2. Commit ac0a0cd7 extracts transport task executor;
3. Commit 75c3a640 normalizes load and store transport operations;
4. Commit 455db543 makes client query asynchronous.

## Test
Run offline inferences successfully with AIV TransProvider commit bb2e968.
Unit tests all passed.


ucm/store/asu/cc/asu_store.cc
Infinite666, line 361: Take a look at this function, if any error happens, this function should return zero represented for nothing found instead of status.

ucm/transport/kv/asu/client/src/asu_client_impl.cpp
```diff
@@ -274,7 +157,7 @@ Status AsuClientImpl::StoreAsync(const std::vector<KVBuffer>& entries, TaskId& t
 
 Status AsuClientImpl::DeleteAsync(const std::vector<CacheKey>& keys, TaskId& taskId)
 {
-    return SubmitAsync(ClientOpType::DELETE, keys, taskId);
+    return SubmitAsync(ClientOpType::DELETE, keys, 0, taskId);
```
Wwwzff: what's the behavior when timeout value is set to 0?

Fengli5355: In Transport Impl, the timeout would be replaced by timeout from configs:
```c++
const auto timeoutMs = task->timeoutMs == 0 ? config_.timeoutMs : task->timeoutMs;
```

If the timeout in config file is 0, then the Delete operation will not return error in finite time. I will add a config check on initializing, and use timeout from config here.

ucm/transport/kv/asu/trans/src/asu_transport_impl.cpp
```diff
@@ -51,74 +47,20 @@ namespace UC::ASU {
...
-    return Status::OK();
-}
-
 }  // namespace
 
-void AsuTransportImpl::ReleaseSubBatchResources(TransportSubBatchContext& subBatchContext)
+AsuTransportImpl::AsuTransportImpl()
+    : taskExecutor_(std::make_unique<TransportTaskExecutor>(
+          config_, ioScheduler_, transProvider_, sendBufferManager_, flagBufferManager_,
+          protocolManager_, connManager_, nextRequestCid_, registeredRegionsMu_,
```
yuanzhg078: TransportTaskExecutor is constructed before its dependencies are initialized, creating implicit lifecycle coupling; consider creating it after dependency initialization in Init() and destroying it before releasing those dependencies in Shutdown().

Fengli5355: Agree.

ucm/store/asu/cc/asu_store.cc
```diff
@@ -241,7 +241,19 @@ class ClientBackend final : public AsuBackend {
     AsuStatus Query(const std::vector<UC::ASU::CacheKey>& keys,
                     const UC::ASU::QueryOptions& options, UC::ASU::QueryResult& result) override
     {
-        return client_->Query(keys, options, result);
+        UC::ASU::TaskId taskId = UC::ASU::kInvalidTaskId;
+        auto status = client_->QueryAsync(keys, options, taskId);
+        if (!status.ok()) { return status; }
+
+        UC::ASU::TaskResult taskResult;
+        status = client_->Wait(taskId, options.timeoutMs, taskResult);
```
yuanzhg078, line 250: AsuStore::ClientBackend::Query discards the taskId when Wait() times out, while ClientTaskManager retains timed-out tasks. 
ref:
```
if (status.code != StatusCode::TIMEOUT) {
    (void)Remove(taskId);
}
```

Fengli5355: Done

ucm/store/asu/cc/asu_store.cc
Fengli5355: The current implementation agrees with the statement of LookupOnPrefix in ucmstore_v1.h:
```c++
    /**
     * @brief Check whether the given blocks exist in storage.
     *
     * @param blocks Array of block identifiers to test.
     * @param num Number of block identifiers to test.
     * @return Expected<ssize_t>
     *   - On success: an index representing the maximum index of blocks found in storage
     *     is present, returns -1 if none are found.
     *   - On failure: appropriate Status code.
     * */
    virtual Expected<ssize_t> LookupOnPrefix(const Detail::BlockId* blocks, size_t num) = 0;
```

Fengli5355: When it returns a status code here, a Status struct object is returned actually, which will not be implicitly converted into an integer.

Fengli5355: *Expected* has two prototypes.
```c++
template <class T>
class Expected {
    std::variant<Status, T> v_;

public:
    Expected(T&& val) : v_(std::move(val)) {}
    Expected(Status err) : v_(err) {}
    bool HasValue() const noexcept { return v_.index() == 1; }
    explicit operator bool() const noexcept { return HasValue(); }
    T& Value() & { return std::get<T>(v_); }
    T&& Value() && { return std::get<T>(std::move(v_)); }
    Status Error() const { return std::get<Status>(v_); }
};
```

Infinite666 (approved):

PR #1191, opened 2026-08-04
[Fix] Fix the logic of AsuClient Check.

Fengli5355:
## Purpose
Modify AsuClient Check.

## Modifications 
1. Check API removes *result*.
2. Check only returns whether a task is in-progress/pending (false) or elsewise (true).

## Test
Offline inference test is passed.
Unit tests are all passed.

### DIFF
--- a/ucm/store/asu/cc/asu_store.cc
+++ b/ucm/store/asu/cc/asu_store.cc
@@ -275,10 +275,7 @@ public:
         return client_->DeleteAsync(keys, taskId);
     }
 
-    AsuStatus Check(UC::ASU::TaskId taskId, UC::ASU::TaskResult& result) override
-    {
-        return client_->Check(taskId, result);
-    }
+    bool Check(UC::ASU::TaskId taskId) override { return client_->Check(taskId); }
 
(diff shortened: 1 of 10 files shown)


pyxyzc:
LGTM :)

ucm/store/asu/cc/asu_store.cc
```diff
@@ -275,10 +275,7 @@ class ClientBackend final : public AsuBackend {
         return client_->DeleteAsync(keys, taskId);
     }
 
-    AsuStatus Check(UC::ASU::TaskId taskId, UC::ASU::TaskResult& result) override
-    {
-        return client_->Check(taskId, result);
-    }
+    bool Check(UC::ASU::TaskId taskId) override { return client_->Check(taskId); }
```
Infinite666, line 278: Is `ClientBackend` this class still neccessary?

Infinite666 (approved):

ucm/store/asu/cc/asu_store.cc
Fengli5355: No, it's not. Further PR will remove this adaptor.

PR #1194, opened 2026-08-05
[Refactor] Remove useless ClientBackend

Fengli5355:
## Purpose
AsuStore no more exposes either Client or Transport as optional backends. This PR removes the class ClientBackend.

## Modifications 
Remove ClientBackend. AsuStore replaces backend_ with client_.

### DIFF
--- a/ucm/store/asu/cc/asu_store.cc
+++ b/ucm/store/asu/cc/asu_store.cc
@@ -213,99 +213,34 @@ UC::ASU::TransportConfig BuildTransportConfig(const Config& config, std::size_t 
     return transportConfig;
 }
 
-class ClientBackend final : public AsuBackend {
-public:
-    AsuStatus Init(const Config& config) override
-    {
-        client_ = UC::ASU::CreateAsuClient();
-        UC::ASU::AsuClientConfig asuConfig;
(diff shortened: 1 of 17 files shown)


ucm/store/asu/cc/asu_store.cc
```diff
@@ -701,6 +634,24 @@ class AsuStore final : public StoreV1 {
         throw std::logic_error("unhandled ASU tensor layout");
     }
 
+    AsuStatus Query(const std::vector<UC::ASU::CacheKey>& keys,
+                    const UC::ASU::QueryOptions& options, UC::ASU::QueryResult& result) const
+    {
+        UC::ASU::TaskId taskId = UC::ASU::kInvalidTaskId;
+        auto status = client_->QueryAsync(keys, options, taskId);
```
Infinite666: Do we still need `QueryOptions` for this function?

Infinite666 (approved):

ucm/store/asu/cc/asu_store.cc
Fengli5355: QueryOptions and QueryMode will be removed in this PR.

PR #1219, opened 2026-08-11
[Feat] Add LookupOnReverse interface for hybrid model mamba block lookup

sumingZero:
## Purpose
Add a reverse-order query interface (LookupOnReverse) to the UCM store layer, dual to the existing LookupOnPrefix. This enables hybrid (full-attention + Mamba) models to maximise KV cache reuse: FA blocks use prefix lookup (front-to-back, stop at first miss), while Mamba blocks use reverse lookup (back-to-front, stop at first hit). Unlike prefix, presence is not required to be contiguous — the function simply returns the index of the rightmost block found in storage.
## Modifications 
Store layer — new LookupOnReverse / lookup_on_reverse interface across the full stack:
- C++ StoreV1 interface (ucmstore_v1.h): pure virtual LookupOnReverse — scans from the last index down to 0, stops at the first present block, returns its index (or -1)
- C++ store implementations (8 stores):
- Posix SpaceManager: sequential back-to-front single-block file access scan
- Cache BufferManager: checks buffer hits for all blocks; if the last block is a buffer hit, returns immediately; otherwise only queries backend for miss blocks after the buffer hit index; returns max(bufferHit, backendHit)
- MooncakeStore: batch RPC existence check, delegates missed blocks to the right of the mooncake hit to backend reverse scan
- Ds3fsStore / Compressor / HealthBreakerStore / EmptyStore / FakeStore: straightforward implementations
- Python (ucmstore_v1.py): abstract lookup_on_reverse on UcmKVStoreBaseV1
- Python connectors: UcmPcStoreV1 and UcmPipelineStore concrete implementations
- pybind11 binding (pipeline_store.py.cc): LookupOnReverse method + module registration
- Test mock (mock_store.h): MOCK_METHOD for LookupOnReverse
Integration layer:
- RankConsistencyManager.lookup_on_reverse: masks known-missing blocks with re-query on hit, mirroring the lookup_on_prefix pattern
- hma_connector: replaces the per-key linear backward scan loop (with TODO) with a single lookup_on_reverse call
- hla_connector: replaces Stage 2 sequential backward scan (up to M lookup_all calls) with per-state-group reverse scan (N lookup_on_reverse calls); takes the min of rightmost hits across state groups; earlier state groups truncate the search window for subsequent ones
## Test
- C++ unit tests (4 test files): verified LookupOnReverse returns -1 when no blocks exist, returns num-1 when all exist, and returns the correct rightmost hit index when blocks are partially present (e.g., [hit, hit, miss, hit] → reverse returns 3, prefix returns 1)
- Cache buffer manager test: verified reverse delegates to backend only for blocks after the buffer hit index, and returns max(bufferHit, backendHit)
- Health breaker test: verified reverse returns -1 when unhealthy, passes through when healthy

### DIFF
--- a/ucm/store/cache/cc/buffer_manager.h
+++ b/ucm/store/cache/cc/buffer_manager.h
@@ -77,6 +77,13 @@ public:
         }
         return LookupOnPrefixFast(blocks, num);
     }
+    Expected<ssize_t> LookupOnReverse(const Detail::BlockId* blocks, size_t num)
+    {
+        if (!buffer_ || loadBackendOnly_) {
+            return LookupThrough<&StoreV1::LookupOnReverse>(blocks, num);
+        }
+        return LookupOnReverseFast(blocks, num);
+    }
     void Prefetch(const Detail::BlockId* blocks, size_t num)
     {
         if (backend_) { backend_->Prefetch(blocks, num); }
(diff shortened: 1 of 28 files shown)


ucm/store/ucmstore_v1.h
```diff
@@ -91,6 +91,28 @@ class StoreV1 {
...
+     * (back-to-front, stop at first hit) to maximise the reuse window.
+     *
+     * @param blocks Array of block identifiers to test.
+     * @param num Number of block identifiers to test.
+     * @return Expected<ssize_t>
+     *   - On success: the index of the rightmost block present in storage.
+     *     Returns -1 if none of the blocks are found.
+     *   - On failure: appropriate Status code.
+     * */
+    virtual Expected<ssize_t> LookupOnSuffix(const Detail::BlockId* blocks, size_t num) = 0;
```
mag1c-h: `LookupOnReverse` may be better.
Scans the block array from the last element toward the first and returns the index of the first block found in storage. **Presence is not required to be contiguous.**

ucm/store/cache/cc/buffer_manager.h
```diff
@@ -151,6 +158,33 @@ class BufferManager {
...
+        std::vector<size_t> missIdx;
+        Lookup(blocks, num, results, missBlk, missIdx);
+        ssize_t bufferHitIdx = -1;
+        for (ssize_t i = static_cast<ssize_t>(num) - 1; i >= 0; --i) {
+            if (results[i]) {
+                bufferHitIdx = i;
+                break;
+            }
+        }
+        if (missBlk.empty()) { return bufferHitIdx; }
```
mag1c-h: The function aims to find the maximum hit index.
1. If the cache has already hit the last block, there's no need to query the backend;
2. If a backend query is required, only query the blocks following the maximum hit index in the cache.

mag1c-h (changes requested):

ucm/store/ucmstore_v1.h
sumingZero: Renamed LookupOnSuffix → LookupOnReverse across the full stack (C++ interface, all store implementations, Python interface/connectors, pybind11 binding, tests, docs). Also updated the docstring to clarify that presence is not required to be contiguous — the function simply scans from the last element toward the first and returns the index of the first block found in storage.

ucm/store/cache/cc/buffer_manager.h
sumingZero: Both optimizations applied in LookupOnReverseFast:

1. If bufferHitIdx == num - 1 (the last block is a buffer hit), return immediately without querying the backend — it's already the maximum possible index.
2. When a backend query is needed, only blocks with missIdx > bufferHitIdx are forwarded to backend_->LookupOnReverse(). Blocks at or before the buffer hit index are skipped since they cannot produce a higher hit.

mag1c-h (approved):

PR #1226, opened 2026-08-13
[Feat] Add CPU-accessible Ascend device buffers

yumingyue624:
## Purpose
Enable BufferPool to allocate Ascend device memory that can also be polled by the CPU on supported platforms.

## Modifications
1. Add an Ascend VMM allocation flow using aclrtMallocPhysical, aclrtReserveMemAddress, aclrtMapMem, and host read-write access.
2. Add a dedicated CPU-accessible device memory type to BufferRegion and BufferPool without changing existing memory types.
3. Keep unsupported backends explicit through ReservedBuffer and add allocation and slot-clearing coverage.

## Tests
Run on Ascend A2 with CANN 8.5.1:
```text
./build-physical-flag-buffer-tests/ucm/shared/test/ucmshared.test --gtest_filter=BufferPoolTest.*
[==========] Running 13 tests from 1 test suite.
[==========] 13 tests from 1 test suite ran. (8377 ms total)
[  PASSED  ] 13 tests.
```

### DIFF
--- a/ucm/shared/pool/buffer_pool.cc
+++ b/ucm/shared/pool/buffer_pool.cc
@@ -139,7 +139,8 @@ bool BufferPool::IsValidPointer(const void* ptr) const
 
 Status BufferPool::ZeroMemory(void* ptr, std::size_t size) const
 {
-    if (memory_type_ == MemoryType::ASCEND_DEVICE) {
+    if (memory_type_ == MemoryType::ASCEND_DEVICE ||
+        memory_type_ == MemoryType::ASCEND_DEVICE_CPU_ACCESSIBLE) {
         const auto status = UC::Trans::ZeroDeviceMemory(ptr, size);
         if (status.Failure()) { return Status::Error(name_ + ": failed to zero device memory"); }
     } else {
--- a/ucm/shared/pool/buffer_pool.cc
+++ b/ucm/shared/pool/buffer_pool.cc
@@ -140,7 +140,7 @@ bool BufferPool::IsValidPointer(const void* ptr) const
 Status BufferPool::ZeroMemory(void* ptr, std::size_t size) const
 {
     if (memory_type_ == MemoryType::ASCEND_DEVICE ||
-        memory_type_ == MemoryType::ASCEND_DEVICE_CPU_ACCESSIBLE) {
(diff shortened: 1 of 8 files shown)


ucm/shared/trans/ascend/ascend_buffer.cc
```diff
@@ -141,6 +159,61 @@ std::shared_ptr<void> Trans::AscendBuffer::MakeDeviceBuffer(size_t size)
...
+        &prop, ACL_RT_MEM_ALLOC_GRANULARITY_MINIMUM, &granularity);
+    if (ret != ACL_SUCCESS || granularity == 0 ||
+        size > std::numeric_limits<size_t>::max() - (granularity - 1)) {
+        return nullptr;
+    }
+    const auto allocationSize = (size + granularity - 1) / granularity * granularity;
+
+    aclrtDrvMemHandle handle = nullptr;
+    ret = aclrtMallocPhysical(&handle, allocationSize, &prop, 0);
+    if (ret != ACL_SUCCESS) { return nullptr; }
```
Infinite666: Add some error logs inside the abnormal branches.

ucm/shared/pool/buffer_pool.cc
```diff
@@ -139,7 +139,8 @@ bool BufferPool::IsValidPointer(const void* ptr) const
 
 Status BufferPool::ZeroMemory(void* ptr, std::size_t size) const
```
Infinite666, line 140: @pyxyzc 
1. Here can just call `UC::Trans::ZeroDeviceMemory` since we have already defined it as a common interface in `Trans`, no need to determine the memory type.
2. Change `ZeroDeviceMemory` to `Memset` which is a more common interface for usage.

ucm/shared/pool/buffer_region.h
Infinite666, line 32: @pyxyzc 
If the naming of `BufferMemoryType` change to sth like `Host/HostMappedDevice/Device/DeviceMappedHost`, that will be more readable and common for other devices.

ucm/shared/trans/ascend/ascend_buffer.cc
yumingyue624: Done

Infinite666 (approved):

PR #1233, opened 2026-08-17
[Opt] Initialize DramStore

nrj868:
## Purpose
Initialize the DramStore P2P KV-cache client so it can be loaded as a leaf store through the UcmPipeline connector and run end-to-end on Ascend NPU (hixl/HCCS one-sided transport).

### DIFF
--- a/ucm/shared/trans/ascend/ascend_event.cc
+++ b/ucm/shared/trans/ascend/ascend_event.cc
@@ -1,7 +1,7 @@
 /**
  * MIT License
  *
- * Copyright (c) 2025 Huawei Technologies Co., Ltd. All rights reserved.
+ * Copyright (c) 2026 Huawei Technologies Co., Ltd. All rights reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -21,22 +21,16 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  * */
-#ifndef UNIFIEDCACHE_TRANS_SIMU_BUFFER_H
-#define UNIFIEDCACHE_TRANS_SIMU_BUFFER_H
-
-#include "trans/detail/reserved_buffer.h"
+#include <acl/acl.h>
+#include "trans/event.h"
 
 namespace UC::Trans {
 
-class SimuBuffer : public ReservedBuffer {
-public:
-    // Simu has no real device, so device-mapped-host memory is plain host memory.
-    bool SupportsDeviceMappedHostBuffer() const override { return true; }
-    std::shared_ptr<void> MakeDeviceMappedHostBuffer(size_t size) override;
-    std::shared_ptr<void> MakeDeviceBuffer(size_t size) override;
-    std::shared_ptr<void> MakeHostBuffer(size_t size) override;
-};
+Status Event::Synchronize() const
+{
+    if (!Valid()) { return Status::OK(); }
+    const auto ret = aclrtSynchronizeEvent(reinterpret_cast<aclrtEvent>(NativeHandle()));
+    return ret == ACL_SUCCESS ? Status::OK() : Status{ret, std::to_string(ret)};
+}
 
 }  // namespace UC::Trans
-
-#endif
--- a/ucm/store/asu/cc/asu_store.cc
+++ b/ucm/store/asu/cc/asu_store.cc
@@ -22,7 +22,6 @@
  * SOFTWARE.
  * */
 #include "asu_store.h"
-#include <acl/acl.h>
 #include <algorithm>
 #include <any>
 #include <cctype>
@@ -38,6 +37,7 @@
 #include <utility>
 #include "asu_client/asu_client.h"
 #include "logger/logger.h"
+#include "trans/event.h"
 #include "ucmstore_v1.h"
 
 namespace UC::AsuStore {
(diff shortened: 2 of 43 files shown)


ucm/store/dram/cc/task_manager.cc
```diff
@@ -207,6 +209,9 @@ std::vector<Request> TaskManager::BuildRequests(OpType op, std::vector<IoEntry>
 
     const auto routed = dependencies_.router->RouteKeys(keys);
     const auto batchLimit = std::min(config_.requestBatchEntries, kMaxProtocolBatchEntries);
+    for (const auto& [node, indexes] : routed) {
```
Infinite666: Delete this log or you can do it inside the loop below.

ucm/store/dram/cc/reply_service.cc
```diff
@@ -45,8 +47,20 @@ Status ReplyService::Init()
         !options_.publishEvent) {
         return Status::InvalidParam("invalid ReplyService options");
     }
-    auto status = buffers_.Init("dram_reply_slots", BufferPool::MemoryType::HOST, options_.slotSize,
-                                options_.slotCount, true);
+#ifdef UC_DRAM_ASCEND_BACKEND
```
Infinite666: Use `Device::Init` and `Device::Setup` instead.

ucm/store/dram/cc/node_scheduler.cc
```diff
@@ -137,6 +141,20 @@ void NodeScheduler::Publish(NodeId nodeId, NodeEvent event)
 void NodeScheduler::RunActors(Runner& runner) noexcept
 {
     try {
+        {
+#ifdef UC_DRAM_ASCEND_BACKEND
```
Infinite666: Same as `reply_service.cc`, use the common method in `Device`. And btw, do we really need to call `aclInit` when running with ucm and vllm?

ucm/store/dram/cc/node_actor.cc
```diff
@@ -39,20 +53,33 @@ void FillTransferEntries(const std::vector<IoEntry>& entries,
...
-        std::memcpy(target.key.data(), source.blockId.data(), source.blockId.size());
+        const auto key = StorageKey(source.blockId, source.shardId);
+        std::memcpy(target.key.data(), key.data(), key.size());
         target.addr = source.buffer.address;
         target.len = static_cast<std::uint32_t>(source.buffer.length);
         target.idx = source.shardId;
     }
 }
 
+const char* RequestStateName(RequestState state) noexcept
```
Infinite666: Maybe writting a `toString` method in its header file is more clean.

ucm/store/dram/cc/dram_store.cc
```diff
@@ -100,123 +101,182 @@ class DramStore final : public StoreV1 {
...
+            const auto initRet = aclInit(nullptr);
+            if (initRet != ACL_SUCCESS && initRet != ACL_ERROR_REPEAT_INITIALIZE) {
+                return Status::Error("aclInit failed: " + std::to_string(initRet));
+            }
+            const auto setRet = aclrtSetDevice(config_->deviceId);
+            if (setRet != ACL_SUCCESS) {
+                return Status::Error("aclrtSetDevice failed: " + std::to_string(setRet));
+            }
+        }
+        const auto transferTimeout =
```
Infinite666, line 106: What's the point to split the timeout config into load and dump if we only need the maximum one?

ucm/store/dram/cc/dram_store.cc
```diff
@@ -100,123 +101,182 @@ class DramStore final : public StoreV1 {
...
         return Status::Unsupported();
 #endif
     }
 
     void StopGraph()
     {
-        if (!config) { return; }
+        if (!config_) { return; }
+
+        const bool wasStarted = started_;
```
Infinite666: Looks like this variable is useless. Just print the log directly or add the `started_` in log is enough.

ucm/store/dram/cc/dram_store.cc
nrj868: load & dump are timed seperated in TaskManager, but the Transport only takes a single deadline.

ucm/store/dram/cc/dram_store.cc
nrj868: yep

ucm/shared/trans/ascend/ascend_event.cc
```diff
@@ -0,0 +1,36 @@
...
+ * */
+#include <acl/acl.h>
+#include "trans/event.h"
+
+namespace UC::Trans {
+
+Status Event::Synchronize() const
+{
+    if (!Valid()) { return Status::OK(); }
+    const auto ret = aclrtSynchronizeEvent(reinterpret_cast<aclrtEvent>(NativeHandle()));
```
Infinite666, line 32: Maybe we can add a function like `Get()` that returns `reinterpret_cast<cudaEvent_t>(event.NativeHandle())` in cuda or `reinterpret_cast<aclrtEvent>(NativeHandle())` in ascend.

ucm/shared/test/case/trans/trans_test.cc
```diff
@@ -23,9 +23,68 @@
...
+
+TEST_F(UCTransUnitTest, EventPreservesBorrowedNativeHandle)
+{
+    constexpr std::uintptr_t nativeHandle = 42;
+    const UC::Trans::Event event{nativeHandle};
+    EXPECT_TRUE(event.Valid());
+    EXPECT_EQ(event.NativeHandle(), nativeHandle);
+}
+
+#ifdef UCM_TEST_RUNTIME_ASCEND
```
Infinite666: This definition will be deleted by #1244.

ucm/store/asu/cc/asu_store.cc
```diff
@@ -103,10 +103,10 @@ void LogAsuStatus(const char* operation, const AsuStatus& status)
 
 Status WaitPrerequisiteEvent(std::uintptr_t eventHandle)
 {
-    if (eventHandle == 0) { return Status::OK(); }
-    auto ret = aclrtSynchronizeEvent(reinterpret_cast<aclrtEvent>(eventHandle));
-    if (ret == ACL_SUCCESS) { return Status::OK(); }
-    return Status::Error("aclrtSynchronizeEvent failed: " + std::to_string(ret));
+    const auto status = Trans::Event{eventHandle}.Synchronize();
+    if (status.Success()) { return AsuStatus::OK(); }
```
Infinite666: Just use the status in UCM instead of `AsuStatus`.

Infinite666 (approved):

Infinite666 (approved):

Wwwzff (approved):

PR #1234, opened 2026-08-17
[Feat] support yuanrongStore|Posix

iamk123:
## Purpose
Add yuanrongStore in pipelineStore，Solve #https://github.com/ModelEngine-Group/unified-cache-management/issues/1177

Solution：
Single-node Load/Dump
image

Dual-node Load/Dump
image

## Modifications 
add yuanrongStore


ucm/store/yuanrongstore/cc/yuanrong_store.cc
```diff
@@ -0,0 +1,476 @@
...
+#include "logger/logger.h"
+#include "task_manager.h"
+#include "time/now_time.h"
+#include "ucmstore_v1.h"
+#include "yuanrong_config.h"
+#include "yuanrong_helper.h"
+
+namespace UC::YuanRongStore {
+
+class YuanRongStore : public StoreV1 {
```
dante159753, line 45: The current develop base (46dc9c9) adds the pure virtual StoreV1::LookupOnReverse. This class does not override it, so after merging the current base YuanRongStore is abstract and MakeYuanRongStore() cannot instantiate it whenever the YuanRong target is enabled. Please implement the tiered reverse lookup (YuanRong plus the Posix misses) and cover it with a test before merging.

test/test_yuanrong_dump_queue_source.py
```diff
@@ -0,0 +1,240 @@
...
+                depth += 1
+            elif source[index] == "}":
+                depth -= 1
+            index += 1
+        self.assertEqual(depth, 0, f"{name} body is not balanced")
+        return source[start : index - 1]
+
+    def test_prerequisite_stream_is_created_in_worker_thread(self):
+        setup_body = self._function_body("DumpQueue::Setup")
+        worker_body = self._function_body("DumpQueue::WorkerStage")
```
dante159753: These source tests are out of sync with the implementation in this PR. Running python test/test_yuanrong_dump_queue_source.py executes 18 tests and 9 fail: DumpQueue::WorkerStage / DumpQueue::DumpOne no longer exist, and yuanrong_reaper_queue_depth is not present in the config. Separately, python test/test_yuanrong_pipeline_builder.py aborts in setUpClass because the fake native module does not define StoreNotFoundError / StoreUnhealthyError. Please update both test files and add the passing commands/results to the PR.

ucm/store/pipeline/connector.py
```diff
@@ -303,6 +303,54 @@ def _mooncake_posix_pipeline_builder(
...
+        if config.get("io_direct", False):
+            if object_size % 4096:
+                raise ValueError(
+                    "YuanRong object size must be aligned to 4096 bytes for "
+                    "io_direct"
+                )
+        shards_per_block = block_size // shard_size
+        posix_config["tensor_size"] = object_size
+        posix_config["shard_size"] = object_size
+        posix_config["block_size"] = object_size * shards_per_block
```
dante159753, line 349: This makes the worker-side Posix block size use the compact YuanRong payload, but the scheduler skips this branch (device_id == -1) and keeps the aligned block size that the worker published before this builder ran. When sum(tensor_size_list) < shard_size because of 4096-byte Store padding, workers write smaller files while scheduler-side Posix GC accounts each file using the larger block size (maxFileCount = capacity / blockSize), causing premature recycling and lower effective capacity. Please publish/derive the same effective Posix block size for the scheduler, or make GC use the actual persisted file size. This is conditional on padding and is a GC-capacity issue rather than an I/O data-corruption issue.

dante159753 (commented):
Reviewed the three requested blocking/integration points.

ucm/store/pipeline/connector.py
iamk123: Resolved

test/test_yuanrong_dump_queue_source.py
iamk123: Resolved

ucm/store/yuanrongstore/cc/yuanrong_store.cc
iamk123: Resolved

ucm/store/yuanrongstore/cc/yuanrong_store.cc
```diff
@@ -0,0 +1,528 @@
...
+        std::vector<bool> exists;
+        auto start = NowTime::Now();
+        auto status = heteroClient_->Exist(keys, exists);
+        auto end = NowTime::Now();
+        if (status.IsError() && exists.empty()) {
+            return Status::Error("YuanRong Exist failed: " + status.ToString());
+        }
+        if (exists.size() != num) {
+            return Status::Error("YuanRong Exist returned an unexpected result size");
+        }
```
mag1c-h, line 288: When `Exist` returns an error but exists is non-empty with the correct size, the error is not propagated — it's only logged at DEBUG, and the results are used as-is. This exists vector feeds directly into the hit/miss vector returned to the vLLM scheduler.
If the SDK ever returns a non-empty but unreliable result on partial failure, the scheduler could treat a missing block as cached (or vice versa), causing incorrect scheduling against stale or non-persisted KV cache data. Severity depends on the SDK contract: safe if non-empty results are always trustworthy regardless of status. Otherwise, consider propagating the error unconditionally on `status.IsError()` — preferring an explicit miss/error over a silent false hit when the SDK cannot guarantee a usable result.

ucm/store/yuanrongstore/cc/yuanrong_config.h
```diff
@@ -0,0 +1,77 @@
...
+
+namespace UC::YuanRongStore {
+
+inline constexpr size_t kPersistenceQueueDepth = 128;
+inline constexpr size_t kDefaultMemoryAlignment = 64;
+inline constexpr size_t kDirectIoMemoryAlignment = 4096;
+
+struct Config {
+    std::string host{"127.0.0.1"};
+    int32_t port{9088};
```
mag1c-h: `yuanrong_config.h` defaults `port` to `9088`, but the deployment guide parameter table, example config, and deployment scripts all use `18483`. Deployments that explicitly set `yuanrong_port` are unaffected, but anything relying on the code default connects to the wrong port and fails misleadingly. Consider aligning the default to `18483` or documenting the discrepancy at the config field.

ucm/store/yuanrongstore/docs/deployment_guide.md
```diff
@@ -0,0 +1,495 @@
...
+git am /workspace/0001-delete-yuanrong_backend-param-enable_exclusive_conne.patch
+```
+
+### 3.7 升级CANN (单机可跳过)
+
+升级 Hixl 
+
+```sh
+bash cann-hixl_9.1.0_linux-aarch64.run  --upgrade --quiet --pylocalann-hixl_9.1.0_linux-aarch64.run  --upgrade --quiet --pylocal
+```
```
mag1c-h, line 427: The command could not be executed, likely due to an incorrect paste operation.

ucm/store/yuanrongstore/cc/yuanrong_config.h
iamk123: Resolved .Modified to mandatory display of configuration

ucm/store/yuanrongstore/docs/deployment_guide.md
iamk123: Resolved.Correction command

ucm/store/yuanrongstore/cc/yuanrong_store.cc
iamk123: Resolved

mag1c-h (approved):

PR #1237, opened 2026-08-17
[Bugfix] Add timeout mechanism to GC threads & resolve mutex issue among multi…

UESTC-AHao:
…ple GC instances

## Purpose
Two problems in Posix Store's shard GC.

**1. The GC thread pool has no timeout**

`ShardGarbageCollector::Setup` never called `SetWorkerTimeoutFn`, and `ThreadPool `only starts its monitor thread under if `(timeoutMs_ > 0)` — so no watchdog existed at all. When the NFS mount stops responding, workers block in `opendir/stat` and GC stalls silently: no log, no metric, no alert.

GC does recover once the mount comes back (the blocked syscall returns and `Done()` follows), but nothing is observable in the meantime, and effective concurrency decays to zero as workers block one by one.

**2. Multiple instances sharing a backend collide**

`_gc_owner `evaluates `role != WORKER && dp_rank == 0`, which is a process-local decision. N independently launched instances each have a `dp_rank=0` scheduler, and each concludes it should run GC. Sharing one NFS backend leads to over-eviction.

Not data corruption; it surfaces as freshly written blocks being over-evicted and a lower hit rate.

## Modifications 
**GC task timeout**
Wires the existing `SetWorkerTimeoutFn `into the GC pool, with a new `posix_gc_task_timeout_ms `(default 300000). It bounds a single shard task (one `ProcessTask`); queue wait and total round duration are excluded.

**Cross-instance GC exclusion**
Adds `GcLease`, built from `mkdir `and `rename` only. The choice of primitives follows from the target deployment's mount options:

`nolock: flock/fcntl` degrade to node-local locking and return success while excluding nothing across nodes. Single-host tests pass; multi-node exclusion silently disappears.

`nocto`: file contents read across nodes may be stale indefinitely, so owner identity lives in a file name, read via `readdir`.

On-disk layout (all dot-prefixed; `space_layout.cc` skips hidden entries during scanning, otherwise GC would delete its own lock):

<backend>/.ucm_gc.lock/ held by whoever's mkdir succeeds
<backend>/.ucm_gc.lock/hb.<host>.<pid>.<n> identity in the name, mtime is the heartbeat
<backend>/.ucm_gc.checktime touched to sample server time

Liveness subtracts two server-stamped mtimes `(utime(p, nullptr)` asks the server for its own clock), so cross-node clock skew is irrelevant. Detection is two-strike: a single `stat `may return a copy cached for up to `acregmax `(60s by default), so only reading the same mtime one interval later counts as evidence. Correctness therefore does not depend on the mount's actual attribute timeout.

Takeover renames the whole lock directory away in one atomic step rather than unlinking the heartbeat and rmdir-ing. The latter races: B and C both judge the lock stale, B wins and writes a fresh heartbeat, then C's unlink deletes B's fresh heartbeat, C's rmdir succeeds on the now-empty directory, and C's mkdir succeeds too — two holders, both heartbeating normally, each seeing the other as alive, never converging. With `rename `the loser gets ENOENT and takeover is serialised.

The lock is held across sampling and every drain round, then released. It is acquired before sampling: a sample taken before acquiring may be a stale snapshot from before another instance finished draining, and reclaiming on that basis is over-eviction. Releasing on a timer is equally wrong — the drain loop `(while (trigger))` has no bounded duration and can run for minutes on a backlog, so a timed release hands the lock away mid-deletion.

New configuration (both derived from `posix_gc_check_interval_sec`; explicit values win):
Option                                          | Default | Derivation
posix_gc_cross_instance_lock       | true      | master switch
posix_gc_heartbeat_interval_sec  | 0           | max(1, interval / 4)
posix_gc_stale_threshold_sec       | 0           | max(interval * 2, 60)


ucm/store/posix/cc/shard_gc.cc
```diff
@@ -116,6 +124,33 @@ void ShardGarbageCollector::GCCheckLoop()
...
+    int rounds = 0;
+    const bool gcRunning = trigger;
+    if (gcRunning) { UC::Metrics::UpdateStats(NAME_TO_METRIC_ID("posix_gc_running"), 1.0); }
+    while (!stop_.load() && trigger) {
+        if (leaseEnable_ && !lease_.HoldsLock()) {
+            UC_WARN(
+                "Lost the GC lock mid-drain after {} round(s); a peer may be reclaiming "
+                "concurrently.",
+                rounds);
+        }
```
mag1c-h, line 141: `HoldsLock()==false` means a peer has already run `TakeOverStale` and renamed our lock dir away — i.e., concurrent reclamation is in progress. The code logs and then calls `Execute()` anyway, continuing to delete files for more rounds. This is exactly the over-eviction the lease exists to prevent.

The PR argues against releasing mid-drain on a *timer*, but losing the lock is different: the takeover has already happened, so continuing is actively harmful. `break` is safe — the next cycle re-acquires the lease and resumes.

ucm/store/posix/cc/shard_gc.cc
```diff
@@ -80,6 +85,7 @@ void ShardGarbageCollector::StopBackgroundCheck()
         stop_ = true;
     }
     gcCheckCv_.notify_all();
+    if (leaseEnable_) { lease_.RequestStop(); }
     if (gcCheckWorker_.joinable()) { gcCheckWorker_.join(); }
```
mag1c-h, line 88: `RequestStop` runs before `gcCheckWorker_.join()`. If `Execute()` is mid-drain and slow, the heartbeat file goes stale, a peer takes over after `posixGcStaleThresholdSec` (60s), and concurrent reclamation resumes during shutdown.

`lease_.Release()` (called at end of `RunGcCycle`) already does `StopHeartbeat()` (`RequestStop`+`join`), so the early `RequestStop` is redundant for the between-cycles case and harmful for the mid-cycle case. Move it after the join:

```cpp
if (gcCheckWorker_.joinable()) { gcCheckWorker_.join(); }
if (leaseEnable_) { lease_.RequestStop(); }
```

ucm/store/posix/cc/posix_store.cc
```diff
@@ -260,6 +277,24 @@ class PosixStore : public StoreV1 {
...
             }
+            constexpr size_t kMinGcTaskTimeoutMs = 1000;
+            if (config.posixGcTaskTimeoutMs != 0 &&
+                config.posixGcTaskTimeoutMs < kMinGcTaskTimeoutMs) {
+                return Status::InvalidParam(
+                    "invalid gc task timeout({}ms), use 0 to disable or at least {}ms",
+                    config.posixGcTaskTimeoutMs, kMinGcTaskTimeoutMs);
+            }
+            if (config.posixGcCrossInstanceLock) {
+                if (config.posixGcHeartbeatIntervalSec == 0) {
```
mag1c-h: `DeriveGcLeaseTimings` (called earlier in `ParseConfig`) already ensures `posixGcHeartbeatIntervalSec >= 1`, so the `== 0` branch below is dead code; only the `staleThreshold <= heartbeat` check is reachable. Drop the dead branch, or move derivation after validation if explicit 0 should be a user error.

ucm/store/posix/cc/shard_gc.cc
```diff
@@ -90,22 +96,24 @@ void ShardGarbageCollector::GCCheckLoop()
...
+                    lease_.Release();
+                    break;
+                case GcLease::Acquisition::HeldByPeer:
+                    UC_INFO("Another instance holds the GC lock; skipping this cycle.");
+                    break;
+                case GcLease::Acquisition::Unavailable:
+                    UC_ERROR(
+                        "GC lock unavailable; running GC without cross-instance "
+                        "exclusion, which may over-evict if peers do the same.");
+                    RunGcCycle();
```
mag1c-h: Non-`DuplicateKey` `Claim` failures (e.g. `EACCES`/`EIO`) → `Unavailable` → `RunGcCycle` runs anyway with a single ERROR. Deliberate fallback, but a persistent backend permission/IO problem silently degrades to pre-PR over-eviction, hidden behind one ERROR per cycle.

Consider skipping the cycle (like `HeldByPeer`) when the lock mechanism itself is broken, or surface it more loudly (metric/counter) so the silent degradation is noticeable.

mag1c-h (changes requested):

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: Agreed, the branch is unreachable. Removed the branch.

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: Moved it after the join as suggested.

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: Resolved,changed to skip the cycle.

ucm/store/posix/cc/posix_store.cc
UESTC-AHao: Agreed, the branch is unreachable. Removed the branch.

ucm/store/posix/cc/shard_gc.cc
UESTC-AHao: Resolved， added break.

mag1c-h (approved):

PR #1238, opened 2026-08-17
[Refactor] Refactor Buffer Pool and Format

pyxyzc:
## Purpose

Unify buffer memory interface across different backends, while improving the readability and consistency of Buffer Pool naming conventions.

## Modifications 

  - Unified ZeroDeviceMemory as the backend-independent Trans::Memset interface.
  - Renamed BufferMemoryType values:
      - Host
      - HostMappedDevice
      - Device
      - DeviceMappedHost
  - Updated Buffer Pool, Variable Buffer Pool, Offset Allocator, and related call sites.
  - Converted pool var and member names to lower camel case for consistency with ucm.

## Test

All pool-related tests passed.


ucm/shared/pool/variable_buffer_pool.cc
```diff
@@ -157,16 +155,16 @@ void VariableBufferPool::Reset()
...
 
 Status VariableBufferPool::ZeroMemory(void* address, std::size_t size) const
 {
-    if (memory_type_ == MemoryType::ASCEND_DEVICE) {
-        const auto status = Trans::ZeroDeviceMemory(address, size);
+    if (memoryType_ == MemoryType::Device) {
+        const auto status = Trans::Memset(address, size);
         if (status.Failure()) { return Status::Error(name_ + ": failed to zero device memory"); }
     } else {
         std::memset(address, 0, size);
```
Infinite666: Delete the `std::memset` logics.

ucm/shared/trans/ascend/ascend_buffer.cc
```diff
@@ -299,7 +299,7 @@ Status Buffer::RegisterHostBuffer(void* host, size_t size, void** pDevice)
 
 void Buffer::UnregisterHostBuffer(void* host) { aclrtHostUnregister(host); }
 
-Status ZeroDeviceMemory(void* ptr, std::size_t size)
+Status Memset(void* ptr, std::size_t size)
```
Infinite666: Memset should be a function that can pass the value instead of only setting to zero.

ucm/shared/pool/variable_buffer_pool.cc
pyxyzc: Done.

ucm/shared/trans/ascend/ascend_buffer.cc
pyxyzc: Done.

Infinite666 (approved):

PR #1250, opened 2026-08-18
[Feat] Support MLA + linear attention hybrid models in HLA connector

sumingZero:
## Purpose
Support MLA + linear attention (KDA) hybrid models (e.g., KimiK3) in the HLA connector. The existing connector only supports GQA + GDN hybrids (e.g., Qwen3Next) and fails on MLA hybrids due to five issues:
1. Layout calculation (Ascend): MLA's head_size (= kv_lora_rank + qk_rope_head_dim) is a combined latent, not per-K or per-V. The original code treats it as both K and V sizes, doubling the attention cache and causing tail_size=0 → ValueError.
2. TP dump/load: is_mla=True triggers rank-0-only dump for all blocks. KDA's ssm_state/conv_state are TP-sharded (per-head, different per rank), so non-rank-0 KDA data is lost.
3. Heat update: Scheduler looks up blocks with rank-0 hash. If rank-0's KDA blocks were per-rank hashed, the lookup misses and heat is never updated.
4. Mamba copy order (vllm-ascend >= 0.21.0): vllm-ascend defers do_mamba_copy_block to after start_load_kv, causing the state copy to overwrite UCM-loaded KV cache with zeros. 
5. Missing KV transfer hooks (vLLM >= 0.27.0, CUDA): Kimi-K3's custom MultiHeadLatentAttention bypasses the @maybe_transfer_kv_layer decorator, so wait_for_layer_load and save_kv_layer are never called, breaking the layerwise connector on CUDA.
## Modifications 
hla_connector.py — issues 1-3:
- _attention_component_sizes: Changed from @staticmethod to instance method. Added MLAAttentionSpec branch: K=kv_lora_rank, V=qk_rope_head_dim (from hf_text_config), instead of using head_size for both.
- HLARequestDispatchMeta(RequestDispatchMeta): New subclass with load_full_attn_count / dump_full_attn_count fields. Does not modify the base class.
- _generate_hla_dispatch_meta: Two-pass iteration (full-attn blocks first, then mamba). Returns HLARequestDispatchMeta with boundary counts set when is_mla=True.
- _mla_split_scope / _scope_blocks: New helpers. Split blocks into MLA (shared hash, rank-0-only dump) and KDA (rank-0 shared hash, non-rank-0 per-rank hash, all-ranks dump). Filter nulls per subset to preserve boundary correctness.
- wait_for_save: Removed if is_mla and tp_rank != 0: return. Uses _scope_blocks(is_dump=True).
- save_kv_layer: Removed if is_mla and tp_rank != 0: return.
- _build_dump_transfer_data: Uses _scope_blocks(is_dump=True).
- start_load_kv (new bulk override + layerwise): Uses _scope_blocks(is_dump=False). MLA blocks shared hash, KDA blocks per-rank hash for all ranks. Added device.synchronize() to ensure mamba state copy (compute stream) completes before load DMA (store stream).
- get_num_new_matched_tokens: _prefetch_other_rank_hashes only called with KDA hashes when is_mla=True (MLA is TP-replicated, no per-rank entries).
- _append_mamba_align_state_block: Extracted from nested closure in _generate_hla_dispatch_meta to a standalone method.

v0210/vllm_ascend/mamba_copy_order_patch.py (new) — issue 4:
- Monkey-patches vllm-ascend's preprocess_mamba to execute do_mamba_copy_block immediately (restoring upstream vLLM behavior), then resets copy_bufs.offset=0 so the deferred call in model_runner becomes a no-op. @when_imported is self-guarding — only fires when vllm_ascend.patch.worker.patch_mamba_utils is imported.

v0270/vllm/models/kimi_k3/nvidia/kimi_k3_mla_kv_hook_patch.py (new) — issue 5:
- Monkey-patches MultiHeadLatentAttention.forward to call wait_for_layer_load before and save_kv_layer after, matching the @maybe_transfer_kv_layer decorator contract. No-op when no KV transfer group is active. @when_imported only fires when vllm.models.kimi_k3.nvidia.mla is imported.
apply_patch.py:
- Register both new patches (unconditional import; @when_imported only fires when the target module exists).

### DIFF
--- a/ucm/integration/vllm/patch/apply_patch.py
+++ b/ucm/integration/vllm/patch/apply_patch.py
@@ -126,6 +126,8 @@ def get_supported_versions() -> list[str]:
         "0.24.0",
         "0.25.1",
         "0.26.0",
+        "0.27.0",
+        "0.28.0",
     ]
 
 
@@ -260,18 +262,15 @@ def apply_all_patches() -> None:
             case _:
                 pass
 
-        # Mamba copy order fix: vllm-ascend >= 0.21.0 defers
-        # do_mamba_copy_block to after start_load_kv, causing it to
-        # overwrite UCM-loaded KV cache data. This patch restores
-        # the copy to before start_load_kv (matching upstream vLLM).
-        #
-        # Always import — the @when_imported hook only fires when
-        # vllm_ascend.patch.worker.patch_mamba_utils exists, and the
-        # wrapper is self-guarding (idempotency flag + offset check).
-        # This handles nightly/dev builds whose version string (e.g.
-        # 0.1.dev4128) doesn't match the release version scheme.
-        logger.info("UCM patching mamba copy order...")
+        # Fix: vllm-ascend >= 0.21.0 defers do_mamba_copy_block to after
+        # start_load_kv, overwriting UCM-loaded data. @when_imported is
+        # self-guarding (only fires when the module exists).
         import ucm.integration.vllm.patch.v0210.vllm_ascend.mamba_copy_order_patch
+
+        # Fix: vLLM >= 0.27.0 Kimi-K3's MLA bypasses @maybe_transfer_kv_layer,
+        # so wait_for_layer_load/save_kv_layer are never called. @when_imported
+        # only fires when vllm.models.kimi_k3.nvidia.mla is imported.
+        import ucm.integration.vllm.patch.v0270.vllm.models.kimi_k3.nvidia.kimi_k3_mla_kv_hook_patch
 
         logger.info("UCM patch initialization completed!")
 
--- a/ucm/integration/vllm/patch/apply_patch.py
+++ b/ucm/integration/vllm/patch/apply_patch.py
@@ -260,6 +260,19 @@ def apply_all_patches() -> None:
             case _:
                 pass
 
+        # Mamba copy order fix: vllm-ascend >= 0.21.0 defers
+        # do_mamba_copy_block to after start_load_kv, causing it to
+        # overwrite UCM-loaded KV cache data. This patch restores
+        # the copy to before start_load_kv (matching upstream vLLM).
+        #
+        # Always import — the @when_imported hook only fires when
+        # vllm_ascend.patch.worker.patch_mamba_utils exists, and the
+        # wrapper is self-guarding (idempotency flag + offset check).
+        # This handles nightly/dev builds whose version string (e.g.
+        # 0.1.dev4128) doesn't match the release version scheme.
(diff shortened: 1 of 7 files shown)


qyh111 (commented):
device.synchronize() 在 start_load_kv 无条件调用 注释称"开销可忽略"。但对非 mamba / 非 MLA 的 HLA 模型（如 Qwen3Next 这类 GQA+GDN）是无谓同步。建议在无 mamba-align 组或非目标场景跳过。

sumingZero:
> device.synchronize() 在 start_load_kv 无条件调用 注释称"开销可忽略"。但对非 mamba / 非 MLA 的 HLA 模型（如 Qwen3Next 这类 GQA+GDN）是无谓同步。建议在无 mamba-align 组或非目标场景跳过。

Qwen系列的混合模型也需要这个同步，因为mamba块拷贝是共性问题

ucm/integration/vllm/patch/apply_patch.py
```diff
@@ -260,6 +262,16 @@ def apply_all_patches() -> None:
...
 
+        # Fix: vllm-ascend >= 0.21.0 defers do_mamba_copy_block to after
+        # start_load_kv, overwriting UCM-loaded data. @when_imported is
+        # self-guarding (only fires when the module exists).
+        import ucm.integration.vllm.patch.v0210.vllm_ascend.mamba_copy_order_patch
+
+        # Fix: vLLM >= 0.27.0 Kimi-K3's MLA bypasses @maybe_transfer_kv_layer,
+        # so wait_for_layer_load/save_kv_layer are never called. @when_imported
+        # only fires when vllm.models.kimi_k3.nvidia.mla is imported.
+        import ucm.integration.vllm.patch.v0270.vllm.models.kimi_k3.nvidia.kimi_k3_mla_kv_hook_patch
```
qyh111, line 273: 如果vllm/vllm-ascend后续已经修改了相关函数，这个patch强制打上去会有影响吗

sumingZero: 如果vllm/vllm-ascend后续已经修改相关函数，那这个vllm_ascend.patch.worker.patch_mamba_utils模块就不会存在，这个patch就不会应用

sumingZero: 会有影响。如果后续版本vllm增加了layerwsie的调用，再打这个patch就会重复dump和load，此时这个patch应该禁用，因此需要关注vllm后续的演进，当前已有的0.27.0、0.27.1、0.28.0均需要此patch

qyh111 (approved):

PR #1279, opened 2026-08-25
[fix]validate shared-buffer /dev/shm capacity early and surface minimum cache_buffer_capacity_gb

qyh111:
## Purpose
- Avoid late, hard-to-diagnose failures when the shared-buffer cache (`share_buffer_enable=true`) cannot fit in `/dev/shm`. The buffer is backed by `shm_open` in `/dev/shm` (a tmpfs with a fixed size limit), so an oversized `cache_buffer_capacity_gb` currently fails deep inside the C++ store. Fail early in Python with an actionable message instead.
- Make HMA's `cache_buffer_capacity_gb` handling consistent: HMA creates two shared-buffer stores (FA/WA), so a user-set capacity should be split evenly between them, matching the existing `posix_capacity_gb` halving and the 128GB default split.
- Surface the minimum required `cache_buffer_capacity_gb` in the C++ "too small buffer" error so users know what value to set.
- Align the `load_tokens_threshold` default with the recommended value (2048) so the example config matches the default.

## Modifications
- **`ucm/integration/vllm/ucm_connector.py`**
  - Add module-level helper `_check_shm_capacity(gb)`: compares the configured buffer size against `/dev/shm`'s tmpfs size limit and raises a `RuntimeError` with remediation guidance (increase `/dev/shm` size or decrease `cache_buffer_capacity_gb`) before store creation; no-ops on non-Linux hosts.
  - `_set_default_shm_buffer_capacity`: after resolving `cache_buffer_capacity_gb`, call `_check_shm_capacity` for early validation.
  - Add `import shutil`.
- **`ucm/integration/vllm/hma_connector.py`**
  - `_set_default_shm_buffer_capacity`: now halves `cache_buffer_capacity_gb` whether user-set or default (previously only the 128GB default was split), so FA/WA each get half — consistent with the existing `posix_capacity_gb` halving. Guard with `max(cap // 2, 1)` to avoid producing 0, which the C++ store treats as "unset" (silent fallback to the 256GB default).
  - Import and call `_check_shm_capacity` after splitting.
  - `load_tokens_threshold` default: `0` → `2048`.
- **`ucm/store/cache/cc/cache_store.cc`**
  - Replace `bufferNumber < 1024 || bufferNumber < loadExclusiveBufferNumber * 2` with the equivalent `bufferNumber < max(1024, loadExclusiveBufferNumber * 2)`, and append the minimum required GB to the error: `please set cache_buffer_capacity_gb >= XGB`, where `X = ceil(max(1024, 2 * loadExclusiveBufferNumber) * shardSize / 1GB)`.
  - Add `#include <algorithm>` for `std::max`.
- **`examples/ucm_config_example.yaml`**: unchanged (kept at `load_tokens_threshold: 2048`, now consistent with the code default).


mag1c-h (approved):

ucm/integration/vllm/ucm_connector.py
```diff
@@ -143,6 +144,35 @@ def _get_store_gc_block_size(
...
+    in ``/dev/shm`` (a tmpfs with a fixed size limit). If the configured buffer
+    capacity exceeds what ``/dev/shm`` can hold, allocation fails deep inside
+    the C++ store; raise here instead with an actionable message.
+    """
+    if cache_buffer_capacity_gb <= 0:
+        return
+    try:
+        shm_total = shutil.disk_usage(_SHM_DIR).total
+    except OSError:
+        # /dev/shm unavailable (e.g. non-Linux dev host); defer to the store.
```
harrisonyhq, line 163: HMA 的 /dev/shm 校验只检查了单 store 容量，未考虑 FA+WA 的总占用

PR #1311, opened 2026-08-31
[Fix] fix Asu configs and implementation

Fengli5355:
## Purpose
This PR fixes a bug of fake backend and unifies some configurations.

## Modifications 
1. Seek-control(sc) now uses the value of config, rather than always uses default value.
2. Unify config `timeout_ms` and `default_wait_timeout_ms` with `wait_timeout_ms`.
3. Fake provider processes Exist with respect to `sc` config.

## Test
1. Unit tests are all passed.
4. Offline inference is passed.
5. kv-test exist now behaves correctly with different `sc` configs.

### DIFF
--- a/ucm/store/asu/cc/asu_store.h
+++ b/ucm/store/asu/cc/asu_store.h
@@ -38,6 +38,7 @@ struct Config {
     std::string fakeBackendPath;
     std::uint64_t fakeBackendLatencyMs{1};
     ssize_t sharedProviderMode{0};
+    bool sc{false};
     std::unordered_map<std::string, std::string> clientAttrs;
 };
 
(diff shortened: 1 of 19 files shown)


ucm/store/asu/cc/asu_store.h
```diff
@@ -38,6 +37,7 @@ struct Config {
     std::string fakeBackendPath;
     std::uint64_t fakeBackendLatencyMs{1};
     ssize_t sharedProviderMode{0};
+    bool sc{false};
```
Infinite666: The default value of sc should be true.

Fengli5355: Done

Infinite666 (approved):

PR #1317, opened 2026-09-01
[Opt] Asu fakebackend adds concurrent task processing in provider

Fengli5355:
## Purpose
This PR optimizes FakeTransProvider to accelerate task consuming.

## Modifications 
1. FakeTransProvider processes tasks with a worker pool.
2. Add a configuration for workerThreads number in both asu_kv_test.conf and ucm_config_*.yaml.

## Test
Offline inference with fake provider is passed.
kv-test bench proves that multi-concurrency with worker threads > 1 can reduce time cost.

### DIFF
--- a/ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
+++ b/ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
@@ -23,15 +23,20 @@
  * */
 #include "fake_trans_provider.h"
 #include <algorithm>
+#include <array>
 #include <chrono>
+#include <condition_variable>
 #include <cstddef>
 #include <cstdint>
 #include <cstring>
+#include <deque>
 #include <filesystem>
 #include <fstream>
 #include <iomanip>
 #include <limits>
+#include <shared_mutex>
 #include <sstream>
+#include <stdexcept>
 #include <thread>
 #include <utility>
 #include <vector>
@@ -42,6 +47,7 @@ namespace UC::ASU {
 namespace {
 
 constexpr std::uint16_t kCqeSuccess = 0x000;
+constexpr std::uint16_t kCqeInternalError = 0x006;
 constexpr std::uint16_t kCqeCheckResultBuffer = 0x732;
 constexpr std::uint8_t kBatchEntryOk = 0x0;
 constexpr std::uint8_t kBatchEntryKeyNotFound = 0x3;
@@ -50,6 +56,9 @@ constexpr std::uint8_t kDeleteEntryFailed = 0x1;
 constexpr std::uint8_t kExistEntryNotExist = 0x0;
 constexpr std::uint8_t kExistEntryExist = 0x1;
 constexpr std::uint32_t kExistSeekControlMask = 1U << 16;
+constexpr std::size_t kKeyLockCount = 256;
+
+std::array<std::shared_mutex, kKeyLockCount> g_keyMutexes;
 
 std::uint64_t ReadU64(std::uint32_t low, std::uint32_t high)
 {
@@ -72,7 +81,7 @@ CacheKey ReadKey(const std::uint32_t* data)
     return key;
 }
 
-std::string KeyFileName(const CacheKey& key)
+std::uint64_t KeyHash(const CacheKey& key)
 {
     std::uint64_t hash = 1469598103934665603ULL;
     for (auto byte : key) {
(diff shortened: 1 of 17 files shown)


ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
```diff
@@ -530,26 +668,66 @@ std::vector<Status> FakeTransProvider::Send(const std::vector<SendIoBatch>& ioBa
...
         }
-        std::memcpy(localFlagBuffer, completion.data(), completionSize);
+
+        IoTask task;
+        task.request.resize(ioBatch.len / sizeof(std::uint32_t) +
+                            (ioBatch.len % sizeof(std::uint32_t) != 0));
+        std::memcpy(task.request.data(), localSendBuffer, static_cast<std::size_t>(ioBatch.len));
+        task.requestLength = ioBatch.len;
+        task.flagBuffer = static_cast<std::uint32_t*>(localFlagBuffer);
+        if (!workerPool_->Push(std::move(task))) {
```
yuanzhg078, line 693: `IoTask` retains a raw pointer to a flag-buffer slot that may be released by the timeout/cancel paths before this queued task reaches `PublishCompletion()`. A late worker completion can therefore write into a slot that has already been reused by another request; please keep the slot alive until the task is retired, or guard publication with a cancellation/generation check.

yuanzhg078: Image

ucm/transport/kv/asu/trans/src/fake_trans_provider.h
```diff
@@ -98,8 +116,13 @@ class FakeTransProvider : public TransProvider {
     std::atomic<std::uintptr_t> nextMrHandle_{1};
     std::mutex registeredMemoryMu_;
     std::unordered_map<MRHandle, RegisteredMemory> registeredMemories_;
+    std::unique_ptr<WorkerPool> workerPool_;
 };
 
 FakeTransProviderConfig MakeFakeTransProviderConfig(const TransportConfig& config);
 
+#ifdef ASU_BUILD_TESTS
```
Infinite666: With
```
#define private public
#include "fake_trans_provider.h"
#undef private
```
in `fake_trans_provider_test.cpp`, we can call the private function directly.

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
Fengli5355: Done. Generation machanism is added.

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
```diff
@@ -171,8 +189,118 @@ std::size_t CompletionDwordCount(const std::uint32_t* request)
...
+    const auto entryDwordCount = opcode == KvOpcode::BatchStore || opcode == KvOpcode::BatchRetrieve
+                                     ? kBatchEntryDwordCount
+                                     : kKeyEntryDwordCount;
+    return kSqeDwordCount + static_cast<std::size_t>(batchNumber) * entryDwordCount;
+}
+
+void PublishCompletion(std::uint32_t* flagBuffer, const std::vector<std::uint32_t>& completion)
+{
+    std::memcpy(flagBuffer, completion.data(), 3 * sizeof(std::uint32_t));
+    if (completion.size() > kCqeDwordCount) {
```
Infinite666: We can calculate the actual copy size then call `std::memcpy` once.

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
```diff
@@ -530,26 +668,92 @@ std::vector<Status> FakeTransProvider::Send(const std::vector<SendIoBatch>& ioBa
...
+void FakeTransProvider::PrepareCompletionSlot(IoTask& task)
+{
+    {
+        std::lock_guard<std::mutex> lock(completionSlotsMu_);
+        auto& state = completionSlots_[task.flagBuffer];
+        if (state == nullptr) { state = std::make_shared<CompletionSlotState>(); }
+        task.completionSlotState = state;
+    }
+
+    std::lock_guard<std::mutex> lock(task.completionSlotState->mutex);
```
Infinite666: Do we really need the mutex and generation to protect the task and buffer? Isn't the buffer address different for every task?

Fengli5355: Flag buffer address could be the same for later tasks.

ucm/transport/kv/asu/trans/src/fake_trans_provider.h
Fengli5355: Agree.

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
Fengli5355: Generation machanism is removed. We avoid slot racing by a client-server contract which should be considered more later.

ucm/transport/kv/asu/trans/src/fake_trans_provider.cpp
Fengli5355: Generation will be removed.

Infinite666 (approved):

PR #1326, opened 2026-09-04
[Feat] add extra info into request hash for multimodal/lora/salt support

harrisonyhq:
## Purpose
Make UCM external KV-cache keys request-semantic-aware so that requests sharing identical placeholder tokens (e.g., two different images expanded to the same multimodal placeholder IDs) can no longer false-hit each other's KV, while the shared text prefix stays reusable. Keys remain reproducible across processes without PYTHONHASHSEED.

## Modifications
Add ucm/integration/vllm/request_hasher.py: RequestHasher + a reusable make_request_block_hasher closure that folds vLLM's generate_block_hash_extra_keys (multimodal identifier+position, LoRA identity, cache salt, prompt-embedding content) into the chained block hash; unsupported semantics fail closed via RequestHashError (no token-only fallback).
Switch Direct/LayerWise/Mock/Lite/CP/HLA/FAWA connectors to the common closure; CacheBlend stays plain-text-only (validates and rejects extra semantics).
Keep the existing model/dtype/TP/rank/spec/sparse namespace and physical layout unchanged; the new 3-tuple hash input naturally supersedes old token-only keys (cold-cache rollout).
Add design docs (vllm_request_aware_prefix_hashing_design*.md) and unit tests (test_ucm_request_hasher.py, test_ucm_blend_hash.py, test_ucm_hla_hash.py).
## Test
Unit: 28 passed / 1 skipped on Ascend (vllm 0.23.0) — covers shared-prefix hit semantics, multi-image cursor advance, pure-image divergence, edge cases, connector-level fail-closed, cross-instance reproducibility; blend tests skipped where triton is unavailable.
E2E: Qwen3.8-27B-w8a8 (hybrid) + UCM Cache|Posix, TP=2 — anti-false-hit confirmed in both HBM prefix-cache off and on: reqA(imgX) dumps, then reqB(same tokens, different img) hits only the text-prefix block (~1536 tokens) and correctly misses all image blocks (request-aware hash diverges), then dumps its own image blocks. Text/image reuse (E2/E3) also verified.
Lint: format.sh (black/codespell/isort/actionlint) all pass.


ucm/integration/vllm/ucm_connector.py
```diff
@@ -1205,7 +1172,7 @@ def __init__(
         if role == KVConnectorRole.SCHEDULER:
             self.request_hasher = RequestHasher(vllm_config, 0)
             self._other_rank_hashers = self._make_other_rank_hashers(vllm_config)
-            self._seed = self.request_hasher("UCM_HASH_SEED")
+            self._seed = self.request_hasher.seed
```
flesher0813, line 1175: Why we need save seed here? Can we just use it in request_hasher?

ucm/integration/vllm/ucm_connector.py
```diff
@@ -2646,9 +2610,14 @@ def get_num_new_matched_tokens(self, request, num_computed_tokens):
             return 0, False
         if request.request_id not in self.requests_meta:
             hash_start = time.perf_counter()
-            ucm_block_ids = self.generate_hash(
-                self.hash_block_size, request.all_token_ids, self._seed
-            )
+            try:
+                ucm_block_ids = self.request_block_hasher(request)
+            except RequestHashError as e:
```
flesher0813, line 2615: RequestHashError only inherits from RuntimeError. Could this cause any other exception? Would it be better to catch Exception directly instead?

docs/source/developer-guide/vllm_request_aware_prefix_hashing_design_zh.md
```diff
@@ -0,0 +1,513 @@
+# vLLM 集成的请求语义感知 Prefix Hash 设计
+
+## 文档状态
+
+- 状态：已实施（Implemented）
```
flesher0813, line 5: Can we delete this?

ucm/integration/vllm/ucm_connector.py
harrisonyhq: we need to use this to bind request hasher below.

docs/source/developer-guide/vllm_request_aware_prefix_hashing_design_zh.md
harrisonyhq: fixed

ucm/integration/vllm/ucm_connector.py
harrisonyhq: fixed

harrisonyhq:
https://github.com/ModelEngine-Group/unified-cache-management/issues/1370

qyh111 (approved):

flesher0813 (approved):

SuperMarioYL (approved):

PR #1341, opened 2026-09-08
[Feat] AsuStore modifies dump waiting mechanism.

Fengli5355:
## Purpose
This PR aims to optimize kvClient & AsuStore workflow.

## Modifications 
1. Add `complete-immediately` option for FakeTransProvider, where the send kernel is replaced with a simple filling of flagBuffers.
2. Modify SPSC ring queue. In ConsumerLoop, `sleep` is replaced with a `wait_for` which will be waked up by notifying.
3. Move prerequisite event waits to kvClient worker thread to prevent synchronous Dump calls from blocking model execution.
4. kv-test bench submit tasks asynchronously with configuration term `io_interval_us`.

## Test
1. kv-test bench succeeded (fake backend).
2. offline inference passed (fake backend).


kv_semantics/src/CMakeLists.txt
```diff
@@ -77,11 +77,11 @@ add_library(kv_client SHARED ${KV_CLIENT_SOURCES})
 target_include_directories(kv_client
     PUBLIC
         ${CMAKE_CURRENT_SOURCE_DIR}/../include
+        ${CMAKE_CURRENT_SOURCE_DIR}/common/ds
```
Infinite666: Why do we move ds to public link?

kv_semantics/kv_test/include/asu_client_runner.h
```diff
@@ -24,6 +24,9 @@ class KvClientRunner {
                  CommandResult& result);
     Status Retrieve(const BufferSet& buffers, SubmitMode submitMode, std::uint64_t timeoutMs,
                     CommandResult& result);
+    Status SubmitStore(const BufferSet& buffers, SubmitMode submitMode, kv::TaskId& taskId);
```
Infinite666: Please help to rename the file in kv_test from `asu` prefix to `kv_client` prefix, forgot to change them in my last pr. Thank you!

kv_semantics/kv_test/include/asu_client_runner.h
```diff
@@ -24,6 +24,9 @@ class KvClientRunner {
...
+    Status SubmitRetrieve(const BufferSet& buffers, SubmitMode submitMode, kv::TaskId& taskId);
```
Infinite666: Looks like the name of the method is not compatible with its functionality, it's more like a `StoreAsync / RetrieveAsync`?

kv_semantics/src/CMakeLists.txt
Fengli5355: AsuStore also uses kv_semantics/src/common/ds in this PR.

ucm/store/asu/cc/asu_store.cc
```diff
@@ -24,19 +24,27 @@
...
 #include <mutex>
 #include <numeric>
 #include <stdexcept>
 #include <string>
+#include <thread>
+#include <unordered_map>
 #include <utility>
 #include "kv_client.h"
 #include "logger/logger.h"
+#include "spsc_ring_queue.h"
```
Infinite666: 1. ucm cannot use the spsc_ring_queue implemented in kv client, it should a inner structure.
2. Can we test for using a private stream for the event wait? It's too complicated to bring in a new dump queue here.

kv_semantics/kv_test/include/asu_client_runner.h
Fengli5355: Agree

kv_semantics/kv_test/include/asu_client_runner.h
Fengli5355: :D

ucm/store/asu/cc/asu_store.cc
Fengli5355: Agree

Infinite666 (approved):

Wwwzff (approved):

PR #1359, opened 2026-09-11
【fix】Optimize scheduler‑side lookup performance

Mio310:
## Purpose
Optimize waiting-queue repeated lookups in get_num_new_matched_tokens: avoid regenerating block hashes and re-scanning already confirmed external prefix hits on every scheduler pass,  with the goal of preventing TPOT degradation for long sequences.
## Modifications
Reuse ucm_block_ids from requests_meta when the request already has metadata; clear metadata on RequestStatus.PREEMPTED.
Incremental external prefix lookup when all of the following hold:
request exists in requests_meta
hbm_hit_block_num is not decreased vs recorded value
total_hit_block_num > hbm_hit_block_num
Resume from the last previously hit block (total_hit_block_num - 1) instead of rescanning from hbm_hit_block_num.
If that last block misses, fall back to a full prefix lookup from hbm_hit_block_num (same as the original path).
First-time requests, or requests whose hbm_hit_block_num changed, still do a full lookup from hbm_hit_block_num.

## Test
128k TPOT：11.5 ms


ucm/integration/vllm/ucm_connector.py
```diff
@@ -1547,7 +1554,26 @@ def get_num_new_matched_tokens(
...
+        # lookup, or HBM hit boundary moved) do a full prefix lookup from
+        # hbm_hit_block_num. If the last previously-hit block is gone, fall
+        # back to that same full lookup.
+        lookup_start_block_num = hbm_hit_block_num
+        resume_from_last_hit = False
+        if request.request_id in self.requests_meta:
+            req_meta = self.requests_meta[request.request_id]
+            if (
+                req_meta.hbm_hit_block_num == hbm_hit_block_num
+                and req_meta.total_hit_block_num > hbm_hit_block_num
```
mag1c-h, line 1569: Why is it necessary here to constrain the number of HBM hits to remain constant?

Mio310: modified to： hbm_hit_block_num is not decreased vs recorded value

ucm/integration/vllm/ucm_connector.py
```diff
@@ -1521,10 +1522,16 @@ def get_num_new_matched_tokens(
...
 
-        ucm_block_ids = self.generate_hash(
-            self.hash_block_size, request.all_token_ids, self._seed
-        )
+        if request.request_id not in self.requests_meta:
+            ucm_block_ids = self.generate_hash(
+                self.hash_block_size, request.all_token_ids, self._seed
+            )
+        else:
+            request_mate = self.requests_meta[request.request_id]
```
ygwpz: maybe request_meta?

Mio310: done

ucm/integration/vllm/ucm_connector.py
```diff
@@ -1797,15 +1798,21 @@ def get_num_new_matched_tokens(
...
+            assert self.request_block_hasher is not None
+            try:
+                ucm_block_ids = self.request_block_hasher(request)
+            except Exception as e:
+                logger.error(
+                    f"request {request.request_id} hash error. {type(e).__name__}: {e}"
+                )
+                return 0, False
+        else:
+            request_mata = self.requests_meta[request.request_id]
```
qyh111: 拼写错误，request_meta

qyh111 (commented):
需要进行热度更新

ucm/integration/vllm/ucm_connector.py
```diff
@@ -1743,10 +1744,10 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
         if self.device is None:
             raise RuntimeError(f"Unsupported device platform for UCMDirectConnector.")
 
-    def _prefetch_other_rank_hashes(self, rank0_block_ids: list[bytes]) -> None:
+    def _prefetch_all_rank_hashes(self, rank0_block_ids: list[bytes]) -> None:
         if not self._other_rank_hashers or not rank0_block_ids:
             return
-
+        self.store.prefetch(rank0_block_ids)
```
qyh111, line 1750: self.store.prefetch(rank0_block_ids) 应该在if not self._other_rank_hashers 之前，应该是先更新本rank的hash，然后判断如果有_other_rank_hasher的时候在更新其他rank

ucm/integration/vllm/ucm_connector.py
Mio310: 已修改_prefetch_direct_hit_key_hotness，已修改request_meta

ucm/integration/vllm/ucm_connector.py
Mio310: 已修改

qyh111 (approved):

PR #1361, opened 2026-09-11
[Refactor]: centralize connection and memory lifecycle management

ayaka836:
## Purpose

Refactor the P2P transport stack so peer channels, protocol connections, and registered memory regions have clear ownership and consistent lifecycles.

### `refactor(p2p): streamline transport manager error handling`

- Unify status propagation, error logging, and rollback handling across the P2P stack.

### `refactor(p2p): centralize transport memory region management`

- Introduce `MemoryRegionManager` as the single owner of registered memory regions.
- Share one logical memory handle across transports while maintaining protocol-specific registrations.
- Validate memory ranges centrally and roll back partial multi-transport registrations.
image

### `feat(p2p): add task thread pool`

- Add a reusable task thread pool for subsequent asynchronous control-plane processing.

### `refactor(p2p): integrate manager channel lifecycle`

- Replace the separate control and TCP message channels with `ChannelManager`.
- Unify TCP channels, protocol connections, and their lifecycles under one connection-management path.
- Exchange metadata for the selected protocol during connect and release it during disconnect.
- Simplify DramStore and DramPool transport endpoint configuration.
image

### `refactor(p2p): remove HIXL instance worker threads`

- Remove the dedicated worker thread owned by each HIXL instance.
- Run HIXL operations directly in the current execution context with the required device context.


ucm/transport/p2p/src/core/channel_manager.cpp
```diff
@@ -0,0 +1,591 @@
...
+    decoded.payload.assign(content + protocol->source_size, content + content_size);
+    return Status::OK();
+}
+
+Status ChannelManager::Init(const ManagerID& local, const Endpoint& endpoint, Handler handler)
+{
+    if (local.empty() || !handler) { return Status::InvalidParam(); }
+    Shutdown();
+    local_ = local;
+    thread_pool_ = std::make_unique<ThreadPool>();
```
Infinite666: Maybe we should add a config for max thread.

ucm/transport/p2p/src/core/channel_manager.cpp
```diff
@@ -0,0 +1,591 @@
...
+    const auto* content = frame.data() + sizeof(*protocol);
+    decoded.source.assign(reinterpret_cast<const char*>(content), protocol->source_size);
+    decoded.payload.assign(content + protocol->source_size, content + content_size);
+    return Status::OK();
+}
+
+Status ChannelManager::Init(const ManagerID& local, const Endpoint& endpoint, Handler handler)
+{
+    if (local.empty() || !handler) { return Status::InvalidParam(); }
+    Shutdown();
```
Infinite666: Why do we need to call Shutdown first, will this object be initialized for multiple times?

ucm/transport/p2p/src/core/channel_manager.cpp
```diff
@@ -0,0 +1,591 @@
...
+            }
+            request->ready.notify_one();
+        }
+    }
+    if (cleanup_transports && !peer.empty() && !transports.empty()) {
+        (void)SubmitTask(peer, [this, peer, transports = std::move(transports)]() {
+            for (const auto protocol : transports) {
+                Metadata response;
+                const auto status =
+                    handler_(ManagerMessageType::DisconnectRequest, protocol, peer, {}, response);
```
Infinite666: The IsConnected called by ApplyConnectionLocally is always false since we erase the peer before, so it will return early.

ucm/transport/p2p/src/core/transport_manager.cpp
```diff
@@ -98,59 +53,55 @@ bool TransportForDirect(OperationDirect direct, TransportProtocol& protocol)
...
-        return status;
-    }
+    channel_manager_ = std::make_unique<ChannelManager>();
+    auto rollback = MakeScopeGuard([this]() { channel_manager_.reset(); });
+    P2P_RETURN_IF_ERROR(
+        channel_manager_->Init(
+            manager_id_, LocalEndpoint(),
+            [this](ManagerMessageType type, TransportProtocol protocol, const ManagerID& peer,
+                   const Metadata& request, Metadata& response) {
+                return HandleControlRequest(type, protocol, peer, request, response);
```
Infinite666, line 81: The logics between transport manager and channel manager is a little bit messy.

ucm/transport/p2p/src/core/memory_region_manager.cpp
```diff
@@ -0,0 +1,205 @@
...
+        for (std::size_t i = 0; i < sizeof(value); ++i) {
+            hash ^= value & 0xffU;
+            hash *= kPrime;
+            value >>= 8U;
+        }
+    };
+    mix(detail::PtrToU64(memory.addr));
+    mix(memory.length);
+    mix(static_cast<std::uint32_t>(memory.device_id));
+    return hash == kInvalidMemoryHandle ? 1 : hash;
```
Infinite666: Should we handle the situation that the memory handle is invalid?

ucm/transport/p2p/src/core/transport_manager.cpp
```diff
@@ -172,457 +123,277 @@ Status TransportManager::Shutdown()
...
-        UC_ERROR("transport manager local {} invalid peer={} protocol={}",
-                 ControlOperationName(operation), manager_id, static_cast<uint32_t>(protocol));
-        return Status::InvalidParam();
+    std::lock_guard<std::mutex> lock(ConnectionMutex(manager_id));
+    const bool connected = channel_manager_->IsConnected(endpoint, protocol);
+    if ((type == ManagerMessageType::ConnectRequest && connected) ||
+        (type == ManagerMessageType::DisconnectRequest && !connected)) {
+        return Status::OK();
     }
     const auto it = protocol_map_.find(protocol);
```
Infinite666: Maybe transports or transport_map will be a more easier name to understand.

ucm/transport/p2p/src/core/channel_manager.cpp
ayaka836: Added manager_max_threads to the configuration.

ucm/transport/p2p/src/core/channel_manager.cpp
ayaka836: It will not be initialized multiple times. The redundant Shutdown() call has been removed.

ucm/transport/p2p/src/core/channel_manager.cpp
ayaka836: Good catch. The previous cleanup path erased the peer before invoking the disconnect handler, so the disconnect returned early. Automatic protocol cleanup on TCP failure has now been removed.

ucm/transport/p2p/src/core/transport_manager.cpp
ayaka836: protocol_map_ and transports_ represented the same transport instances. They have been consolidated into a single transports_ map from protocol to the owned transport instance.

ucm/transport/p2p/src/core/transport_manager.cpp
ayaka836: ChannelManager is responsible for all link management. It owns the TCP channels, tracks the connected transport protocols for each peer, serializes peer control operations, and determines when a peer channel can be removed. TransportManager does not retain peer or link state; it only dispatches protocol-specific operations to the selected Transport.

ucm/transport/p2p/src/core/memory_region_manager.cpp
ayaka836: Agreed.kInvalidMemoryHandle is unnecessary.It now been removed.

Infinite666 (approved):

Wwwzff (approved):

PR #1383, opened 2026-09-16
[Feat] Add KV standalone metrics framework

yuanzhg078:
## Purpose

Provide a reusable KV metrics facade and standalone Prometheus exporter for KV-Test without depending on UCM metrics.

## Modifications

- Add a backend-neutral KV metrics facade with cached metric handles and single/batch update interfaces.
- Add a standalone metrics backend supporting counters, gauges, and histograms.
- Implement thread-local double-buffered collection and periodic aggregation for concurrent metric updates.
- Support runtime YAML metric registration and compiled default descriptors.
- Add a generator and CMake targets for generating and validating default metric descriptors.
- Add an HTTP `/metrics` endpoint with configurable labels, address, port, path, and aggregation interval.
- Integrate standalone metrics startup, configuration validation, flushing, and shutdown into KV-Test.
- Add core LOAD/STORE and exporter metrics with the `kv:` prefix.
- Add a basic Grafana dashboard for exporter status, request counts, rates, errors, and submit latency.
- Add focused facade, exporter, YAML registration, concurrency, and aggregation tests.

## Test

- Default metric descriptor generation and consistency check passed.
- Metrics facade compilation check passed.
- Standalone metrics concurrency and exporter smoke tests passed.


kv_semantics/kv_test/src/kv_test_app.cc
```diff
@@ -408,10 +456,23 @@ int KvTestApp::Run(int argc, char** argv)
...
+        if (config.metrics.enabled) {
+            std::cout << " metrics_endpoint=http://" << config.metrics.listenAddress << ':'
+                      << config.metrics.port << config.metrics.path;
+        }
+        std::cout << '\n';
         return kExitSuccess;
     }
 
+    MetricsRuntime metricsRuntime;
+    status = metricsRuntime.Start(config.metrics);
```
Infinite666, line 470: What if the config.metrics.enabled is false?

kv_semantics/kv_test/include/kv_test_types.h
```diff
@@ -167,6 +167,19 @@ struct AsuRuntimeLibraryConfig {
...
 
+struct MetricsServerConfig {
+    bool enabled{false};
+    std::string definitionPath;
+    std::string listenAddress{"127.0.0.1"};
+    std::uint16_t port{9108};
+    std::string path{"/metrics"};
+    std::string source{"kv-test"};
+    std::string modelName{"standalone"};
+    std::string workerId{"asu-0"};
```
Infinite666: Try to avoid using asu semantic, use word like node or endpoint instead.

kv_semantics/src/client/kv_client_impl.cc
```diff
@@ -23,18 +23,69 @@
...
+        &KV_METRIC("kv_client_batch_store_entries_total"),
+        &KV_METRIC("kv_client_batch_store_errors_total"),
+        &KV_METRIC("kv_client_batch_store_submit_duration_seconds")};
+    return metrics;
+}
+
+void RecordSubmission(SubmissionKind kind, std::size_t entryCount, const Status& status,
+                      const metrics::MetricTimer& timer)
+{
+    if (!timer.enabled) { return; }
```
Infinite666: Why don't we use timer.begin or IsEnabled to determine if we enabled metrics so that we can erase the enabled attribute inside timer which looks strange?

kv_semantics/src/client/kv_client_impl.cc
```diff
@@ -23,18 +23,69 @@
...
+namespace {
+
+struct SubmissionMetrics {
+    metrics::CachedMetric* requests;
+    metrics::CachedMetric* entries;
+    metrics::CachedMetric* errors;
+    metrics::CachedMetric* duration;
+};
+
+enum class SubmissionKind { BATCH_LOAD, BATCH_STORE };
```
Infinite666: I think we can just use the op type, there're too many similar enum classes.

kv_semantics/metrics/src/standalone_metrics_backend.cc
```diff
@@ -0,0 +1,931 @@
...
+                }
+            }
+        }
+
+        int BeginWrite() noexcept
+        {
+            while (true) {
+                const int index = writeIndex.load(std::memory_order_seq_cst);
+                activeWriteIndex.store(index, std::memory_order_seq_cst);
+                if (writeIndex.load(std::memory_order_seq_cst) == index) { return index; }
```
Infinite666, line 466: Is here safe for concurrency? Will two competitors get the same index?

kv_semantics/kv_test/src/kv_test_app.cc
yuanzhg078: When config.metrics.enabled is false, MetricsRuntime::Start() returns Status::Success() immediately and does not initialize or start the metrics server. The MetricsRuntime instance remains inactive, and its destructor is therefore a no-op. I’ll add/keep this behavior covered by tests if needed.

kv_semantics/kv_test/include/kv_test_types.h
yuanzhg078: Updated the default worker ID and sample configuration to use the neutral endpoint-0 identifier.

kv_semantics/src/client/kv_client_impl.cc
yuanzhg078: Thanks for the suggestion. I replaced begin + enabled with std::optional<time_point> and added ElapsedSeconds(), so timer validity is explicit without using a sentinel value or checking the global metrics state again.

kv_semantics/src/client/kv_client_impl.cc
yuanzhg078: Agreed. I removed the redundant SubmissionKind enum and now use AsuOpType directly, recording metrics only for batch load and batch store operations.

kv_semantics/metrics/src/standalone_metrics_backend.cc
yuanzhg078: Each thread has its own thread-local ThreadBuffer, so two writer threads may get the same index, but they write to different buffers and do not conflict. The index only selects one of the two slots within the current thread’s buffer.

Infinite666 (approved):

PR #1394, opened 2026-09-18
[Feat]: Add metrics observability for DramPool

zwj20000328-dotcom:
## Purpose
This PR adds a metrics module on the UC::Metrics framework and instruments the dump/load/lookup pipeline: request volume, per-stage durations, NoSpace/failure counts, queue pressure, and pool usage gauges.

## Modifications
- Add the DramPool metrics design document and metrics summary.
- Implement the DramPool metrics module based on UC::Metrics.
- Add histogram-based metric visualization for latency metrics.

## Test
- `drampool_metrics.test` : All unit tests passed.


ucm/store/dram/cc/drampool/drampool_metrics.h
```diff
@@ -0,0 +1,193 @@
...
+
+// Dynamic per-slot-size gauge: drampool_buffer_pool_usage_ratio_<slot_size>.
+inline std::string BufferPoolUsageRatioName(std::uint64_t slotSize)
+{
+    return std::string("drampool_buffer_pool_usage_ratio_") + std::to_string(slotSize);
+}
+
+// ---- update helpers (thin wrappers over UC::Metrics::UpdateStats) ----
+// Accumulates a count (COUNTER semantics: value += count); zero is skipped.
+inline void MetricsCount(const std::string& name, std::uint64_t count)
```
yumingyue624: Could we remove this wrapper and call `UC::Metrics::UpdateStats` directly? The call sites already know whether and by how much to increment the counter; aggregated-count sites can guard `count != 0` locally.

ucm/store/dram/cc/drampool/drampool_metrics.h
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// Records one duration sample (HISTOGRAM semantics: push_back, in ms).
+inline void MetricsObserve(const std::string& name, double valueMs)
+{
+    UC::Metrics::UpdateStats(name, valueMs);
+}
+
+// Overwrites the latest value (GAUGE semantics: value = v).
+inline void MetricsSet(const std::string& name, double value)
```
yumingyue624: These wrappers are also redundant because they only forward their arguments. Please call `UC::Metrics::UpdateStats` directly at the call sites.

zwj20000328-dotcom: ok

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -159,23 +172,57 @@ bool CompletionPoller::PollDataTransfer(CompletionRecord& record)
...
             return false;
         }
 
+        UC::Metrics::UpdateStats(kResponseFailuresTotal, 1);
         UC_ERROR(
             "CompletionPoller flag buffer allocation failed, request_id={}, opcode={}, error={}",
             record.request_id, static_cast<int>(record.opcode), allocateStatus);
         return true;
     }
+    ++flagUsed_;
```
yumingyue624: Could `BufferPool` maintain and expose a thread-safe used-slot count through `Allocate`/`Free` instead? That would avoid duplicating allocation state here and manually updating `flagUsed_` across every release path.

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -159,23 +172,57 @@ bool CompletionPoller::PollDataTransfer(CompletionRecord& record)
 
 bool CompletionPoller::SubmitResponse(CompletionRecord& record)
 {
+    // Batch statistics point (metrics_design.md §4.7): every record passes here exactly
```
yumingyue624: Could we extract this block into a helper? It contains a complete once-only metrics operation and would keep `SubmitResponse()` focused on response submission.

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -263,6 +328,20 @@ bool CompletionPoller::PollResponseTransfer(CompletionRecord& record)
 void CompletionPoller::SettleDataTransfer(CompletionRecord& record,
                                           transport::TransferStatus terminalStatus)
 {
+    // Data-transfer terminal observation (metrics_design.md E group): this single exit
```
yumingyue624: Would it be cleaner to encapsulate this metrics block in a helper so `SettleDataTransfer()` stays focused on settling the transfer?

ucm/store/dram/cc/drampool/drampool_types.h
```diff
@@ -52,6 +52,13 @@ inline std::uint64_t SteadyNowMs()
         std::chrono::duration_cast<std::chrono::milliseconds>(now).count());
 }
 
+inline std::uint64_t SteadyNowUs()
```
yumingyue624, line 57: Could we use the shared `NowTime::Now()` here instead of maintaining separate steady-clock helpers? This would keep time measurement consistent across UCM.

ucm/store/dram/cc/drampool/drampool_metrics.h
```diff
@@ -0,0 +1,173 @@
...
+inline std::atomic<std::uint64_t> g_requestQueueLen{0};
+inline std::atomic<std::uint64_t> g_completionQueueLen{0};
+
+// Histogram observation capacity per thread (UC::Metrics C3 cap). Sized to
+// hold at least two 10s reporter windows of batch-level observations.
+inline constexpr std::size_t kMetricsMaxVectorLen = 10000;
+
+// ---- metric names (grouped as in metrics_design.md §3.2) ----
+// A. Request volume
+inline constexpr char kDumpRequestsTotal[] = "drampool_dump_requests_total";
```
yumingyue624, line 46: Could the metric update call sites use `NAME_TO_METRIC_ID("...")`, consistent with UCM? It caches the resolved metric ID after the first lookup, avoiding repeated string-based resolution on hot paths.

ucm/store/dram/CMakeLists.txt
```diff
@@ -45,53 +49,78 @@ target_link_libraries(dramstore
...
-if(UCM_RUNTIME_ASCEND_FAMILY)
-    # CANN DSOs depend on driver libraries supplied by the runtime host, not the build image.
-    target_link_options(drampool PRIVATE "LINKER:--allow-shlib-undefined")
-endif()
-set_target_properties(drampool PROPERTIES
-    INSTALL_RPATH "$ORIGIN/../ucm/transport/p2p;$ORIGIN/../ucm/shared/metrics"
-)
-install(TARGETS drampool RUNTIME DESTINATION bin COMPONENT ucm)
+
+if(BUILD_UCM_DRAMPOOL)
```
yumingyue624: DramPool should remain enabled by default. `BUILD_UCM_DRAMPOOL` is not defined anywhere, so this condition is false in a normal build and silently drops `drampool_core`, the `drampool` binary, and its install target.

ucm/store/dram/cc/drampool/metrics_reporter.cc
```diff
@@ -0,0 +1,203 @@
...
+#include "drampool_config.h"
+#include "logger/logger.h"
+
+namespace UC::DramPool {
+namespace {
+
+constexpr char kMetricsFileName[] = "drampool_metrics.prom";
+// Wake-up granularity while waiting for the next interval, so Stop() does not
+// block for a full interval (same pattern as HealthServer's poll interval).
+constexpr auto kStopPollInterval = std::chrono::milliseconds(100);
```
yumingyue624, line 45: Write the resource snapshot file in JSON format instead of Prometheus text. This allows YuanRong and DramPool to share the same UCM-side file reporting and parsing flow, and matches the DramPool snapshot contract already merged in #1396.

ucm/store/dram/CMakeLists.txt
```diff
@@ -45,53 +49,78 @@ target_link_libraries(dramstore
...
+        PUBLIC
+            infra_status
+            infra_logger
+            infra_template
+            infra_thread
+            buffer_pool
+            metrics
+            ucm_p2p_transport
+            trans
+            dramstore
```
yumingyue624: `drampool_core` does not use any symbols from `dramstore`. Please remove this redundant dependency to avoid unnecessary build and runtime coupling.

ucm/store/dram/cc/drampool/drampool_server.cc
```diff
@@ -518,10 +519,15 @@ void DramPoolServer::RequestReceiveLoop()
                  task->request->request_id, static_cast<int>(task->request->opcode), controlPeerId,
                  task->peer_one_sided_id);
         // This bounded handoff keeps transport I/O separate from potentially slow request handling.
+        ScopedTimer enqueueWaitTimer(NAME_TO_METRIC_ID(kQueueRequestEnqueueWaitMs));
         bool queueFullLogged = false;
         while (!requestReceiverStop_.load(std::memory_order_acquire)) {
-            if (requestQueue_.TryPush(std::move(task))) { break; }
+            if (requestQueue_.TryPush(std::move(task))) {
+                g_requestQueueLen.fetch_add(1, std::memory_order_relaxed);
```
yumingyue624, line 527: Suggestion: `drampool_queue_request_size` is updated only after a pop, so it may remain stale while the backlog grows. Consider reporting it periodically from the sole consumer to preserve the single-writer behavior while reflecting the current queue depth.

ucm/store/dram/cc/drampool/task_worker.cc
```diff
@@ -258,29 +288,42 @@ Status TaskWorker::ProcessLoad(const KvLoadRequest& request,
...
     if (runtime_.protocol.GetPackedResponseSize(OpType::LOOKUP, request.batch_size) >
         g_config.flagBufferSlotSizeBytes) {
         return Status::InvalidParam("LOOKUP response exceeds configured flag buffer slot size");
     }
     std::vector<std::uint8_t> results(request.batch_size,
                                       static_cast<std::uint8_t>(LookupResult::NotFound));
-    for (std::uint16_t index = 0; index < request.batch_size; ++index) {
-        if (runtime_.metadata.Exist(request.entries[index].key)) {
-            results[index] = static_cast<std::uint8_t>(LookupResult::Exists);
+    std::uint64_t missEntries = 0;
```
yumingyue624: Suggestion: Move lookup miss counting to `ReportBatchMetrics()` and derive it from `record.results`, consistent with dump failures. This keeps metrics bookkeeping out of the lookup scan and prevents `kLookupScanDurationMs` from including the counter update.

ucm/store/dram/cc/drampool/task_worker.cc
```diff
@@ -73,30 +79,36 @@ Status TaskWorker::ProcessOneRequest(RequestTaskPtr task)
...
 
 Status TaskWorker::ProcessDump(const KvDumpRequest& request,
-                               const transport::ManagerID& peerOneSidedId)
+                               const transport::ManagerID& peerOneSidedId,
+                               std::uint64_t beginUs)
 {
+    ScopedTimer prepareTimer(NAME_TO_METRIC_ID(kDumpPrepareDurationMs));
     if (runtime_.protocol.GetPackedResponseSize(OpType::DUMP, request.batch_size) >
         g_config.flagBufferSlotSizeBytes) {
+        prepareTimer.Disarm();
```
Infinite666: Maybe we can set the default value of arm as false, and only set it to true when we need to count the metrics at the normal branch, so that we don't need to call Disarm at every abnormal branch.

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -45,6 +47,65 @@ void ReleaseResponseBuffer(BufferPool& flagBufferPool, CompletionRecord& record)
...
+        case OpType::LOAD:
+            UC::Metrics::UpdateStats(NAME_TO_METRIC_ID(kLoadBatchTotalDurationMs), elapsedMs);
+            break;
+        case OpType::LOOKUP:
+            UC::Metrics::UpdateStats(NAME_TO_METRIC_ID(kLookupBatchTotalDurationMs), elapsedMs);
+            break;
+        default:
+            break;
+    }
+    if (record.opcode == OpType::DUMP) {
```
Infinite666: We can move this logic inside the switch case branch.

ucm/store/dram/cc/drampool/completion_poller.cc
```diff
@@ -45,6 +47,65 @@ void ReleaseResponseBuffer(BufferPool& flagBufferPool, CompletionRecord& record)
...
+}
+
+// Response-buffer release and RTT terminal exit shared by every terminal path:
+// observed from the response-transfer submission
+// (submit_ms, set in SubmitResponse) to the terminal state.
+void SettleResponseTransfer(BufferPool& flagBufferPool, CompletionRecord& record, bool failed)
+{
+    ReleaseResponseBuffer(flagBufferPool, record);
+    if (failed) { UC::Metrics::UpdateStats(NAME_TO_METRIC_ID(kResponseFailuresTotal), 1); }
+    UC::Metrics::UpdateStats(NAME_TO_METRIC_ID(kResponseRttMs),
```
Infinite666, line 120: Do we need to record theis stats when the response is abnormal?

ucm/transport/p2p/src/protocols/hixl/hixl_instance.h
```diff
@@ -60,6 +60,9 @@ class HixlInstance final {
     Status Disconnect(const std::string& remote_engine, int32_t timeout_ms);
     Status TransferSync(const std::string& remote_engine, Opcode opcode,
                         const std::vector<Segment>& segments, int32_t timeout_ms);
+    Status QueueTransferSync(const std::string& remote_engine, Opcode opcode,
```
Infinite666: Why are we modifing hixl related files in this PR? If it's another feature or bugfix, we better seperate the code.

examples/metrics/metrics_configs.yaml
```diff
@@ -354,6 +382,26 @@ gauge:
     documentation: "Aggregate reserved Fence admission capacity"
     multiprocess_mode: livemostrecent
 
+  # -- DramPool server observability ------------------------------------------
+  - name: "drampool_metadata_entry_count"
+    documentation: "Current number of cached entries (blocks) in the pool, summed over all shards"
+    multiprocess_mode: livemostrecent
+  - name: "drampool_buffer_pool_usage_ratio_<slot_size>"
```
yumingyue624: The UCM consumer registers only this literal placeholder name, while the DramPool snapshot contains concrete names such as `drampool_buffer_pool_usage_ratio_4096`. The concrete names fail metric ID resolution and are dropped, and no matching Prometheus mapping exists, so these gauges never produce samples in `/metrics`.

yumingyue624 (commented):
There are still several latency observability gaps:

1. `drampool_queue_request_enqueue_wait_ms` measures only the time spent waiting for `TryPush()` to succeed; it does not cover the time a request remains in the queue before `TaskWorker` dequeues it.
2. The dump/load/lookup preparation metrics cover multiple potentially expensive operations, including metadata processing and asynchronous HiXL submission, so they cannot identify which component caused the latency.
3. `drampool_load_transfer_duration_ms` measures from `ExecuteAsync()` returning until `CompletionPoller` observes a terminal status, combining actual HiXL transfer time, polling delay, and `GetStatus()` execution time.
4. `drampool_response_rtt_ms` similarly combines response transfer time, polling delay, and `GetStatus()` execution time.
5. The response-side `ExecuteAsync()` call itself is not included in `drampool_response_rtt_ms`, and no current metric records its submission latency.

yumingyue624 (commented):
6. `drampool_*_batch_total_duration_ms` is not a complete DramPool end-to-end latency. It starts when `TaskWorker` dequeues the request and ends before `SubmitResponse`, excluding request queue residence time, response-buffer waiting, response packing/submission, and response transfer completion. No current metric covers the full server-side request lifecycle.

yumingyue624 (commented):
6. `drampool_*_batch_total_duration_ms` is not a complete DramPool end-to-end latency. It starts when `TaskWorker` dequeues the request and ends before `SubmitResponse`, excluding request queue residence time, response-buffer waiting, response packing/submission, and response transfer completion. No current metric covers the full server-side request lifecycle.

PR #1396, opened 2026-09-18
[Feat] Import DramPool resource metrics from snapshot logs

yumingyue624:
## Purpose

Import DramPool metric snapshots from its resource log into the UCM metrics pipeline.

## Modifications

- Read and validate the latest complete DramPool snapshot, preserve counter and histogram deltas across restarts, and elect one reporter per host.
- Share file polling, leader election, and state persistence with the existing YuanRong resource reporter while keeping store-specific parsing and publication in each store.
- Merge pre-aggregated histogram buckets into the native metrics buffer through the shared `HistogramStat` representation, using the DramPool metric names directly without redundant unit or suffix rejection.
- Report resource-log collection failures through the common `${reporter_key}_resource_log_read_errors_total` counter.
- Start the reporter with the Dram pipeline and cover parsing, validation, reset handling, locking, lifecycle, and metric publication.

## Dependencies

The DramPool snapshot producer must register the source metric names and histogram buckets in the UCM metrics configuration.

## Test

All related tests passed.


ucm/store/dram/resource_reporter.py
```diff
@@ -0,0 +1,350 @@
...
+        finally:
+            self._release_leadership()
+
+
+def start_drampool_resource_reporter(config: dict) -> DramPoolResourceReporter | None:
+    global _REPORTER
+
+    path = str(config.get("drampool_resource_log_path", ""))
+    enabled = bool(config.get("drampool_resource_metrics_enable", bool(path)))
+    if not enabled or not path or int(config.get("device_id", -1)) >= 0:
```
Infinite666, line 340: Is there a mistake here? return None when device >= 0?

yumingyue624: `device_id == -1` identifies the scheduler process, while workers use `device_id >= 0`. The reporter runs only in the scheduler process to avoid importing the shared host-level log multiple times.

ucm/store/dram/resource_reporter.py
```diff
@@ -0,0 +1,350 @@
...
+    )
+
+
+def snapshot_deltas(
+    current: DramPoolResourceSnapshot, previous: DramPoolResourceSnapshot | None
+):
+    """Compute metric deltas, treating decreases as source resets."""
+    counters, histograms = {}, {}
+    for name, value in current.counters.items():
+        old = previous.counters.get(name, 0) if previous is not None else None
```
nrj868, line 135: Nit: Let `previous` be `current` when it is None may avoid these None checks.

nrj868 (dismissed):

ucm/store/dram/resource_reporter.py
yumingyue624: Done. `previous` now defaults to `current`, removing the redundant `None` checks.

nrj868 (approved):

Infinite666 (approved):

Wwwzff (approved):

PR #1413, opened 2026-09-20
[Feat][Cache2] Implement partitioned cache control

NaganooMei:
## What this PR does

- implements the shared Cache2 control-region lifecycle with memfd creation, fd passing, mmap, and join support
- materializes `CtrlLayout` metadata, rank descriptors, slots, buckets, lock stripes, and per-rank CLOCK hands
- implements striped bucket locking, optimistic hit pinning, and fixed-size CLOCK eviction confined to the current rank partition
- wires `Buffer` setup, allocation, lookup, release, and move-only handle semantics to the worker-local `deviceId` rank
- adds focused unit and multi-process tests for the shared control domain, rank-partitioned CLOCK, optimistic lookup, lock striping, owner election, and rollback

## Review conclusions (2026-09-22)

- **Scope:** connector changes have been removed. This PR is limited to the assigned `Buffer`, `CtrlStrategy`, and `CtrlLayout` control-plane implementation, plus their IPC helpers and focused tests.
- **`CtrlLayout` interface:** the public surface retains the layout operations required by its consumers. Layout construction helpers, raw header access, offsets, size calculations, CLOCK-hand access, and internal getters are private; `Buffer` and `CtrlStrategy` access them as friends. Test-only accessors avoid widening the production API.
- **Handle behavior:** invalid handles are guarded by `Data()`, `DeviceData()`, `HostAccessible()`, and state accessors. `Buffer::Get()` retries until it obtains a valid handle. Destroying an owner while its slot is still `Loading` publishes `Failed`, allowing a later requester to become the new owner.
- **Join lifecycle:** the creator serves exactly `localRankSize` joiners and then closes the listening socket. A domain containing one creator and `localRankSize` joiners therefore completes without leaving an unbounded accept loop.
- **CLOCK initialization:** all per-rank CLOCK hands are initialized once by `InitHeader()` before the control region is published. `InitSlotRange(rank)` initializes only that rank's slots and no longer resets the same CLOCK hand a second time.
- **Upstream compatibility:** the branch is rebased on the current `feature_a5` implementation and preserves the upstream `Handle::HostAccessible()` API with invalid-handle protection.

## Control-region sizing

Sizing example requested during review:

```text
node cache capacity = 128 GiB
localRankSize       = 8
shardSize           = 176 KiB
```

The current layout calculation gives:

```text
slotCount     = floor(128 * 2^30 / (176 * 2^10)) = 762,600
slotsPerRank  = 762,600 / 8                       = 95,325
bucket target = ceil(762,600 / 2)                 = 381,300
bucketCount   = next power of two                 = 524,288
lockCount     = min(bucketCount, 65,536)          = 65,536
buckets/lock  = 524,288 / 65,536                  = 8
```

Using the current 64-bit layout (`sizeof(Header) = 2,112`, `sizeof(SlotMeta) = 192`) and a 48-byte process-shared `SharedMutex` on the target AArch64 libc:

| Region | Calculation | Size |
| --- | ---: | ---: |
| Header | fixed, including `kMaxRanks = 128` descriptors and CLOCK hands | 2,112 B (2.06 KiB) |
| Buckets | 524,288 x 8 B | 4 MiB |
| Lock stripes | 65,536 x 48 B | 3 MiB |
| Slot metadata | 762,600 x 192 B | 139.64 MiB |
| **Total control region** | aligned fixed layout | **about 146.64 MiB** |

The 65,536-lock array is therefore not large relative to this configuration: it occupies about 3 MiB, roughly 2% of the control region and about 0.0023% of the 128 GiB data capacity. Each lock protects eight buckets, which is a reasonable stripe width for reducing false contention among concurrent miss, preallocation, and CLOCK-relocation paths. These process-shared mutexes are initialized once by the creator for each control domain. The current `kMaxLockStripes = 1 << 16` is retained.

## Validation

- `git diff --check`: passed
- C++ clang-format check: fixed after the latest rebase
- focused Cache2 tests cover DP2/TP8 control-region isolation and sharing, cross-process striped locking, rank-local CLOCK eviction with second chance and pin protection, optimistic ready hits, preallocation owner election, and cross-stripe rollback
- latest full-suite GTest run: passed

## Scope

This PR is rebased on `feature_a5` commit `4e18fca3`. `DataStrategy`, connector configuration, and the remaining data-plane import/export path are outside this PR.


ucm/store/cache2/cc/cache_buffer.h
```diff
@@ -52,6 +68,7 @@ class Buffer {
         }
 
     public:
+        Handle() = default;
```
mag1c-h: A Handle should not have an "invalid" state; external code must not be able to create an invalid Handle via a default constructor. Handles can only be created using the `Buffer::Get` factory method, and any handle obtained from `Buffer::Get` is guaranteed to be valid.

NaganooMei: Addressed in [95b6e53](https://github.com/ModelEngine-Group/unified-cache-management/commit/95b6e53e86655928ff2eac2c8a624d1d62d3616c).

`Handle` now has a deleted default constructor, no boolean conversion, and no invalid-handle fallback values in its accessors. The private `TryAcquireSlot()` returns a slot index or an internal retry sentinel; `Buffer::Get()` keeps retrying and constructs a `Handle` only after successfully acquiring a slot reference. Successful worker-side setup is an explicit precondition, checked with debug assertions; observer buffers must not call `Get()`.

Move operations still clear the source's ownership so its destructor cannot release the reference twice. This is only moved-from bookkeeping, not a failed acquisition result. The slot's `Loading` / `Ready` / `Failed` data states and abandoned-owner failure publication are preserved.

Tests now use scope-based handle destruction and slot-level assertions for acquisition retries. Added compile-time construction/interface checks and regression cases for move ownership and `Get()` waiting while all slots are pinned, then returning an acquired handle after release.

Validation: `git diff --check` and clang-format checks passed. C++ compilation and GTests have not been completed locally: the Windows environment lacks the required build dependencies and a usable Linux/WSL environment. CI update: rebased all 11 PR commits onto the latest feature_a5 (4294d54), replacing the previous merge-based synchronization. The PR history contains no merge commits, and the final file tree is identical to the pre-rebase version. Diff and clang-format checks passed. The rebased head is 95b6e53; [CI results](https://github.com/ModelEngine-Group/unified-cache-management/pull/1413/checks) are pending.

ucm/store/cache2/cc/cache_buffer.h
```diff
@@ -68,33 +86,27 @@ class Buffer {
...
+        size_t SlotIndex() const { return slotIdx_; }
         void* Data() { return buf_->data_.DataAt(slotIdx_); }
         void* DeviceData() { return buf_->data_.DeviceDataAt(slotIdx_); }
-        CtrlLayout::SlotMeta::State GetState() const
-        {
-            return buf_->ctrl_.Layout().SlotMetaArr()[slotIdx_].state.load(
-                std::memory_order_acquire);
-        }
+        CtrlLayout::SlotMeta::State GetState() const { return buf_->GetState(slotIdx_); }
+        bool Ready() const { return GetState() == State::Ready; }
```
mag1c-h: The Ready interface is redundant given that the GetState interface is already exposed.

ucm/store/cache2/cc/ctrl_strategy.h
```diff
@@ -23,17 +23,176 @@
...
+                cfg.bufferCapacity, cfg.shardSize, cfg.localRankSize);
+        }
+        dims.rankCount = cfg.localRankSize;
+        dims.slotsPerRank = slotCount / dims.rankCount;
+        dims.slotSize = cfg.shardSize;
+        slotCount = dims.rankCount * dims.slotsPerRank;
+        dims.bucketCount = CtrlLayout::RecommendBucketCount(slotCount);
+        if (dims.bucketCount == 0 || dims.bucketCount > kMaxBuckets ||
+            (dims.bucketCount & (dims.bucketCount - 1)) != 0) {
+            return Status::InvalidParam("invalid cache2 bucket count({})", dims.bucketCount);
```
mag1c-h: Under what circumstances does RecommendBucketCount return an invalid number of buckets?

ucm/store/cache2/cc/ctrl_strategy.h
```diff
@@ -23,17 +23,176 @@
...
     CtrlLayout& Layout() { return layout_; }
+    const CtrlLayout& Layout() const { return layout_; }
+
+private:
+    static Status ResolveDimensions(const Config& cfg, Dimensions& dims)
+    {
+        /* A5 has one data partition per local worker. The public configuration
+         * describes node-wide capacity; the control layout divides it evenly. */
+        if (cfg.localRankSize == 0 || cfg.localRankSize > kMaxRanks) {
+            return Status::InvalidParam("invalid cache2 local rank size({})", cfg.localRankSize);
```
mag1c-h: Parameter validity checks were performed multiple times.

ucm/store/cache2/cc/ctrl_strategy.h
```diff
@@ -23,17 +23,176 @@
...
+        return Status::OK();
+    }
+
+    Status SetupCreator(const Config& cfg)
+    {
+        Dimensions dims;
+        auto s = ResolveDimensions(cfg, dims);
+        if (s.Failure()) { return s; }
+        if (dims.slotsPerRank > std::numeric_limits<size_t>::max() / dims.rankCount) {
+            return Status::InvalidParam("cache2 slot count overflow");
```
mag1c-h: This condition will never be true.

ucm/store/cache2/cc/ctrl_strategy.h
```diff
@@ -23,17 +23,176 @@
...
+        }
+        auto slotCount = dims.rankCount * dims.slotsPerRank;
+        auto lockCount = CtrlLayout::LockStripeCount(dims.bucketCount);
+        auto prefixSize = CtrlLayout::SlotMetaOffset(dims.bucketCount, lockCount);
+        if (slotCount >
+            (std::numeric_limits<size_t>::max() - prefixSize) / sizeof(CtrlLayout::SlotMeta)) {
```
mag1c-h: This condition will never be true.

ucm/store/cache2/cc/cache_buffer.h
NaganooMei: Addressed in [32c1af2](https://github.com/ModelEngine-Group/unified-cache-management/commit/32c1af251dc8b5a7dd039d6588e23d28f93a05f7). Removed Handle::Ready() and updated all four test call sites to compare GetState() with State::Ready. No Ready() calls remain in Cache2 or its tests. Validation: clang-format and git diff --check passed; Linux C++/GTests were not run locally on Windows.

ucm/store/cache2/cc/ctrl_strategy.h
NaganooMei: Addressed in [32c1af2](https://github.com/ModelEngine-Group/unified-cache-management/commit/32c1af251dc8b5a7dd039d6588e23d28f93a05f7). RecommendBucketCount() clamps its target between the power-of-two bounds kMinBuckets and kMaxBuckets, then rounds up to a power of two. Removed the redundant validation of its return value.

ucm/store/cache2/cc/ctrl_strategy.h
NaganooMei: Addressed in [32c1af2](https://github.com/ModelEngine-Group/unified-cache-management/commit/32c1af251dc8b5a7dd039d6588e23d28f93a05f7). Removed the repeated localRankSize check from ResolveDimensions(); Buffer::Setup() retains the range check before its modulo operation and the call to CtrlStrategy::Setup(). Documented this precondition on CtrlStrategy::Setup(). Also consolidated joiner header validation into the comparison with the expected layout, including lockStripeCount. Config::Validate() is unchanged.

ucm/store/cache2/cc/ctrl_strategy.h
NaganooMei: Addressed in [32c1af2](https://github.com/ModelEngine-Group/unified-cache-management/commit/32c1af251dc8b5a7dd039d6588e23d28f93a05f7). Removed this slot-count overflow check: slotsPerRank is floor(slotCount / rankCount), so multiplying it back by rankCount cannot exceed the original slotCount.

ucm/store/cache2/cc/ctrl_strategy.h
NaganooMei: Updated in [32c1af2](https://github.com/ModelEngine-Group/unified-cache-management/commit/32c1af251dc8b5a7dd039d6588e23d28f93a05f7). Removed the explicit control-region byte-size overflow checks from both creator and joiner setup, along with the now-unused <limits> include. Validation: clang-format and git diff --check passed; Linux C++/GTests were not run locally on Windows.

mag1c-h (approved):

PR #1432, opened 2026-09-21
[feat] Pass world_size from the connector down to the store

sunyuecheng01:
## Purpose

Tell the UCM store how many ranks it serves. The connector passes `world_size`
down to the store now — it was missing before.

## Modifications

- Pass `world_size` to every store the connector creates (direct, FAWA, HLA).
- The number counts **one DP engine only** (TP x PP x PCP). DP is left out,
  because each DP engine owns its own store.
- Scheduler & Worker: get the number from the config only.


ucm/integration/vllm/hma_connector.py
```diff
@@ -706,6 +706,7 @@ def _create_store(
         """Instantiate one UCM store with worker tensor layout metadata."""
 
         name, module_path, config = self._base_store_config(store_suffix)
+        config["world_size"] = self._get_store_world_size()
```
mag1c-h: The configuration item name is `local_rank_size`.

ucm/integration/vllm/ucm_connector.py
```diff
@@ -1586,6 +1586,16 @@ def _set_default_shm_buffer_capacity(self, config: dict[str, Any]) -> None:
         # (before store creation) if the tmpfs cannot hold it.
         _check_shm_capacity(int(config["cache_buffer_capacity_gb"]))
 
+    def _get_store_world_size(self) -> int:
```
sunyuecheng01: 函数名不合理

ucm/integration/vllm/hma_connector.py
sunyuecheng01: change to “`config["local_rank_size"] = self._get_world_size()`”

mag1c-h (approved):

PR #1433, opened 2026-09-21
[Docs] Update model-tour

KINN123:
Summary
- Add bilingual UCM Docker deployment tutorials for Qwen, DeepSeek, MiniMax, and GLM.
- Organize each tutorial into A2 and H100 tabs, with A2 selected by default.
- Cover container setup, UCM configuration, service startup, and API requests.


docs/docs-site/tools/runtime_images.py
```diff
@@ -7,7 +7,8 @@
 
 def on_config(config):
     policy_path = Path(__file__).resolve().parents[3] / ".github/release/release.yaml"
-    policy = yaml.safe_load(policy_path.read_text())
+    # policy = yaml.safe_load(policy_path.read_text())
```
flesher0813, line 10: Maybe delete it would be better, same to every commented code

PR #1434, opened 2026-09-21
[feat] distribute MLA dumps across TP ranks

sunyuecheng01:
## Purpose

MLA KV cache is replicated across TP ranks and stored under rank-neutral keys, so in a TP group every rank previously dumped the same blocks — `TP ×` redundant writes to the UCM store, and rank-0-only participation in the direct/layerwise paths. This PR assigns each MLA block to a single writer rank so dump work is evenly distributed without changing what is stored.

## Modifications

- Add `_mla_dump_parallelism()` / `_select_mla_dump_blocks()` on `UCMDirectConnector`; MLA blocks are striped by `rank::tp_size` while UCM keys and vLLM block addresses stay paired. Non-MLA is unchanged.
- `UCMDirectConnector.wait_for_save`: remove the rank-0-only early return for MLA, select the local stripe per request, and keep UCM keys rank-neutral for MLA (`not self.is_mla` guards the per-rank re-hash).
- `UCMLayerWiseConnector.save_kv_layer` / `_submit_layerwise_dump_task` / `start_load_kv`: same stripe selection, so only locally submitted blocks are tracked and dumped per layer.
- `UCMHybridLinearAttentionConnector._mla_split_scope`: stop dropping MLA blocks on non-rank-0 workers — every rank dumps its own MLA stripe while KDA blocks keep their per-rank scoping.
- `UCMCPConnector`: override `_mla_dump_parallelism()` to stripe only within DCP replicas (PCP/DCP sharding is already applied in `bind_connector_metadata`).
- Load paths are intentionally untouched: each rank still loads the full block list. HMA behavior is unchanged.

## Test

TODO


ucm/integration/vllm/ucm_connector.py
```diff
@@ -2572,7 +2585,13 @@ def _submit_layerwise_dump_task(
 
         metadata = self._get_connector_metadata()
         block_ids_by_request = {
-            request_id: set(metadata.request_meta[request_id].dump_block_ids[0])
+            request_id: set(
+                self._select_mla_dump_blocks(
+                    *metadata.request_meta[request_id].dump_block_ids
```
sunyuecheng01: check是否影响其他

ucm/integration/vllm/ucm_connector.py
```diff
@@ -2825,6 +2847,11 @@ def wait_for_save(self) -> None:
 class UCMCPConnector(UCMLayerWiseConnector):
     _defer_scheduler_store = True
```
sunyuecheng01: DCP先不做

PR #1246, opened 2026-08-18
Support request-level asynchronous KV cache loading

lc1314555:
## Purpose
 Add request-level asynchronous KV cache loading for vLLM UCMDirectConnector.
 While one request is waiting for external KV data, vLLM can process other runnable requests and resume the waiting request after loading completes, reducing head-of-line blocking.
## Modifications 
 - Add the use_request_async configuration.
  - Record the UCM-to-vLLM block mapping after GPU block allocation.
  - Submit KV load tasks without blocking model execution.
  - Poll load completion through get_finished().
  - Prevent duplicate loads after rescheduling or preemption.
  - Report load failures and invalidate affected GPU blocks.
  - Add unit tests for dispatch, completion, duplicate prevention, and failure handling.
  - Currently support only UCMDirectConnector with a single KV cache group, PP=1, and without CP or layerwise loading.

PR #1442, opened 2026-09-22
[Feat] Step-wise asynchronous KV cache loading

CoooolHai:
## Purpose
Extend the request-level asynchronous KV cache load (PR #1246) to the HLA
    and HMA/FAWA connectors, and add the observability stack around it.
    
## Modifications 
  - Step-wise async load state protocol for HLA/HMA: submit/poll/wait task
    handling, multi-group mapping, and scheduler-side request resume
  - Connector metrics with default metrics config
  - metrics configs for connector observability


ucm/integration/vllm/patch/v0110/vllm/pc/v1/core/sched/scheduler.py
```diff
@@ -628,7 +632,8 @@ def _update_waiting_for_remote_kv(self, request) -> bool:
             failed_recving_kv_req_ids.discard(request.request_id)
         else:
             # Now that the blocks are ready, actually cache them.
-            (block_ids,) = self.kv_cache_manager.get_block_ids(request.request_id)
+            block_groups = self.kv_cache_manager.get_block_ids(request.request_id)
```
flesher0813, line 635: Do we need to modify 0.11.0's patch? Seems no difference

